PR #6, opened 2018-06-02
Avoid false positives when returning interface values

### LINKED ISSUE #5: false positive in unexported-return
```
revive -formatter friendly github.com/cockroachdb/cockroach/pkg/...
  ⚠  unexported-return  exported func NewZeroNode returns unexported type sql.planNode, which can be annoying to use
  pkg/sql/zero.go:36:49
```
The code in question: https://github.com/cockroachdb/cockroach/blob/97c3622c2a6f4ec579ed93655b9d8a83732ac654/pkg/sql/zero.go#L36

It's true that the returned object is of an unexported type, but it is being returned as an interface, which is exported.

tamird:
Fixes #5.

Please consider being a bit more patient before closing issues in the future.

### DIFF
--- a/fixtures/golint/unexported-return.go
+++ b/fixtures/golint/unexported-return.go
@@ -12,14 +12,20 @@ func Exported() hidden { // MATCH /exported func Exported returns unexported typ
 
 // ExpErr returns a builtin type.
 func ExpErr() error { // ok
+  return nil
 }
 
-func (hidden) ExpOnHidden() hidden { // ok
+func (h hidden) ExpOnHidden() hidden { // ok
+  return h
+}
+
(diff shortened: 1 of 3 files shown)


mgechev:
Thanks for the PR! I'll look at it the first thing in the morning tomorrow :-)

mgechev:
The tests are passing without the fix being introduced https://github.com/mgechev/revive/pull/8

tamird:
Thanks! The false positive I observed in CockroachDB requires this fix, so
there's something missing in the test I added.

I'll take a look at this tonight.

tamird:
Well, I haven't been able to write a test to demonstrate the problem, but this change does fix the false positive in #5.

PR #13, opened 2018-06-07
Adds rule superfluous-else (an extension of indent-error-flow)

chavacava:
The new rule checks for `if-then-else` where the `then` block ends with branching statement like `continue`,`break`, or `goto`.


mgechev:
@chavacava LGTM! Would you also add a row in the [Available Rules](https://github.com/mgechev/revive#available-rules) table?

chavacava:
@mgechev done.

mgechev:
🚢

PR #16, opened 2018-06-10
New rule: Confusing naming

chavacava:
Adds `confusing-naming`, a rule that warns on methods with names that differ only by capitalization (inspired by [FindBugs' _Nm: Confusing method names_](http://findbugs.sourceforge.net/bugDescriptions.html#NM_CONFUSING))

### DIFF
--- a/fixtures/confusing-naming1.go
+++ b/fixtures/confusing-naming1.go
@@ -45,3 +45,5 @@ func (t foo) ABar() { // Should not warn
 func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to other method of 'bar'/
 	return
 }
+
+func x() {}
--- a/fixtures/confusing-naming1.go
+++ b/fixtures/confusing-naming1.go
@@ -3,49 +3,45 @@
 // Package pkg ...
 package pkg
 
-type foo struct {}
+type foo struct{}
 
-func (t foo) aFoo() { 
-	return
-} 
-
-
-
-func (t *foo) AFoo() { // MATCH /Method 'AFoo' differs only by capitalization to other method of 'pkg/foo'/
-	return
-}
- 
-type bar struct {}
-
-func (t *bar) aBar() { 
-	return
-} 
-
-func (t *bar) aFoo() { // Should not warn 
+func (t foo) aFoo() {
 	return
 }
 
-
-
-func aGlobal(){
-
-}
-
-func AGlobal(){ // MATCH /Function 'AGlobal' differs only by capitalization to other function/
-}
-
-func ABar() { // Should not warn 
-
-}
-
-func aFoo() { // Should not warn 
-
-}
-
-func (t foo) ABar() { // Should not warn 
+func (t *foo) AFoo() { // MATCH /Method 'AFoo' differs only by capitalization to method 'aFoo'/
 	return
-} 
+}
 
-func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to other method of 'pkg/bar'/
+type bar struct{}
+
+func (t *bar) aBar() {
+	return
+}
+
+func (t *bar) aFoo() { // Should not warn
+	return
+}
+
+func aGlobal() {
+
+}
+
+func AGlobal() { // MATCH /Method 'AGlobal' differs only by capitalization to method 'aGlobal'/
+}
+
+func ABar() { // Should not warn
+
+}
+
+func aFoo() { // Should not warn
+
+}
+
+func (t foo) ABar() { // Should not warn
+	return
+}
+
+func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to method 'aBar'/
 	return
 }
--- a/fixtures/confusing-naming1.go
+++ b/fixtures/confusing-naming1.go
@@ -11,7 +11,7 @@ func (t foo) aFoo() {
 
 
 
-func (t *foo) AFoo() { // MATCH /Method 'AFoo' differs only by capitalization to method 'aFoo'/
+func (t *foo) AFoo() { // MATCH /Method 'AFoo' differs only by capitalization to other method of 'pkg/foo'/
 	return
 }
  
@@ -31,7 +31,7 @@ func aGlobal(){
 
 }
 
-func AGlobal(){ // MATCH /Method 'AGlobal' differs only by capitalization to method 'aGlobal'/
+func AGlobal(){ // MATCH /Function 'AGlobal' differs only by capitalization to other function/
 }
 
 func ABar() { // Should not warn 
@@ -46,6 +46,6 @@ func (t foo) ABar() { // Should not warn
 	return
 } 
 
-func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to method 'aBar'/
+func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to other method of 'pkg/bar'/
 	return
 }
--- a/fixtures/confusing-naming1.go
+++ b/fixtures/confusing-naming1.go
@@ -47,3 +47,16 @@ func (t bar) ABar() { // MATCH /Method 'ABar' differs only by capitalization to 
 }
 
 func x() {}
+
+type tFoo struct {
+	asd      string
+	aSd      int  // MATCH /Field 'aSd' differs only by capitalization to other field in the struct type tFoo/
+	qwe, asD bool // MATCH /Field 'asD' differs only by capitalization to other field in the struct type tFoo/
+	zxc      float32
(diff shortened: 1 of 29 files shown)


rule/confusing-naming.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+type lintConfusingNames struct {
+	methodNames map[string][]*ast.Ident // a map from struct names to method id nodes
+	onFailure   func(lint.Failure)
+}
+
+const defaultStructName = "_" // used to map functions
+
+//getStructName of a function receiver. Defaults to defaultStructName
```
mgechev: Would you add a space after `//`

rule/confusing-naming.go
```diff
@@ -0,0 +1,102 @@
...
+// ConfusingNamingRule lints method names that differ only by capitalization
+type ConfusingNamingRule struct{}
+
+// Apply applies the rule to given file.
+func (r *ConfusingNamingRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	fileAst := file.AST
+	walker := lintConfusingNames{
+		methodNames: make(map[string][]*ast.Ident),
```
mgechev: If we change the type of `methodNames` to `map[string]map[string]bool` where the key in the nested map is lower/uppercase identifier, we can make the `checkMethodName` run in `O(1)`.

rule/confusing-naming.go
```diff
@@ -0,0 +1,102 @@
...
+
+	return failures
+}
+
+// Name returns the rule name.
+func (r *ConfusingNamingRule) Name() string {
+	return "confusing-naming"
+}
+
+//checkMethodName checks if a given method/function name is similar (just case differences) to other method/function of the same struct/file.
```
mgechev: Maybe struct/package will be better? Are we going to get a warning if:

```go
// a.go
package foo
func bar() {}
```
```go
// b.go
package foo
func baR() {}
```

mgechev:
Love this rule! Left few comments.

rule/confusing-naming.go
chavacava: done

rule/confusing-naming.go
chavacava: done

rule/confusing-naming.go
chavacava: The answer, in short, is: no
The check is _file-wide_. To do it _package-wide_ I can make initialization of `methodNames` struct outside `Apply` (currently, the struct is initialized inside `Apply`, thus for each file)

I have a problem with the _global_ `methodNames` struct: how to differentiate methods/funcs of (actually) different packages? I can prefix package name to methods/funcs but that is not enough. For example many of the projects in the `vendor` directory have a function `main` in a package `main`

mgechev: What do you think about changing:

```diff
// Package represents a package in the project.
type Package struct {
	fset  *token.FileSet
-	files map[string]*File
+	Files map[string]*File

	TypesPkg  *types.Package
	TypesInfo *types.Info

	// sortable is the set of types in the package that implement sort.Interface.
	Sortable map[string]bool
	// main is whether this is a "main" package.
	main int
	mu   sync.Mutex
}
```

This way, from the current file you'll have access to all other files in the package. In `Apply` you can create an index of all functions and the associated files.

The problem is that this will need to be recomputed across all files in the same package. Maybe it's worth exposing a hash called `Symbols` or `Methods` in the package itself:

```
// map[method_name]file_name
type Methods = map[string]string
```

This map can be build before the linter starts analyzing given package. Also, this could be on demand thing, just like type checking.

Let me know what you think about this idea.

chavacava: As you suggest we need a global index of methods of each package. 
I'm drafting a version of the rule that builds the index on-the-fly: in `Apply` we retrieve the list of method names (in fact a `map[string]map[string]bool` as you proposed) of the package corresponding to the file (using `file.Pkg` as key), then we create a `walker` with this list.

You confirm that, in `Apply`, two `lint.File` of the same package have the same `file.Pkg` ?

fixtures/confusing-naming1.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (t *bar) aFoo() { // Should not warn
+	return
+}
+
+func aGlobal() {
+
+}
+
+func AGlobal() { // MATCH /Method 'AGlobal' differs only by capitalization to method 'aGlobal'/
```
mgechev: Maybe we can show the file where `aGlobal` resides? For example, if `AGlobal` is in:

```
/foo/bar/baz/a.go
```

And `aglobal` is in:

```
/foo/bar/baz/qux/b.go
```
We can show:

```
Method 'AGlobal' differs only by capitalization to method 'aGlobal' in `qux/b.go`
```

rule/confusing-naming.go
mgechev: Right [here](https://github.com/mgechev/revive/blob/master/lint/package.go#L151-L162), the package loops over all of its files and lints them one by one.

chavacava: Hi, I've pushed the draft version (it needs refactoring) of the rule. This version detects confusing names of methods/functions of the same package even if they are in different source files. 

:warning: Test does not pass because rule is applied on each file independently. I didn't find a way to test a rule on a set of files.

mgechev:
@chavacava I'll update the testing logic and merge over the weekend.

chavacava:
Hi @mgechev, perfect! 
Next week (I'll be off until next Monday) I'll do one more iteration on this rule implementation. Thanks.

rule/confusing-naming.go
chavacava: I've pushed a version that, when applied to `revive` source, outputs:

```
vendor/golang.org/x/tools/godoc/vfs/namespace.go:263:21: Method 'Stat' differs only by capitalization to method 'stat' in the same source file
vendor/golang.org/x/tools/godoc/vfs/zipfs/zipfs.go:186:18: Method 'Stat' differs only by capitalization to method 'stat' in the same source file
vendor/golang.org/x/tools/container/intsets/sparse.go:1057:18: Method 'check' differs only by capitalization to method 'Check' in vendor/golang.org/x/tools/container/intsets/util_test.go
vendor/golang.org/x/tools/go/buildutil/allpackages.go:84:6: Method 'allPackages' differs only by capitalization to function 'AllPackages' in the same source file
vendor/golang.org/x/tools/refactor/rename/rename_test.go:1332:6: Method 'main' differs only by capitalization to function 'Main' in vendor/golang.org/x/tools/refactor/rename/rename.go
vendor/github.com/BurntSushi/toml/encode.go:108:21: Method 'encode' differs only by capitalization to method 'Encode' in the same source file
vendor/golang.org/x/tools/go/buildutil/util.go:112:6: Method 'hasSubdir' differs only by capitalization to function 'HasSubdir' in the same source file
vendor/github.com/mattn/go-runewidth/runewidth_posix.go:62:6: Method 'IsEastAsian' differs only by capitalization to function 'isEastAsian' in the same source file
vendor/golang.org/x/tools/go/ast/astutil/imports.go:443:6: Method 'Imports' differs only by capitalization to function 'imports' in the same source file
vendor/golang.org/x/tools/godoc/analysis/analysis.go:267:20: Method 'FileInfo' differs only by capitalization to method 'fileInfo' in the same source file
vendor/github.com/olekukonko/tablewriter/table.go:364:6: Method 'pad' differs only by capitalization to function 'Pad' in vendor/github.com/olekukonko/tablewriter/util.go
vendor/golang.org/x/sys/unix/syscall_linux_amd64.go:78:6: Method 'Gettimeofday' differs only by capitalization to function 'gettimeofday' in vendor/golang.org/x/sys/unix/syscall_linux_amd64_gc.go
vendor/golang.org/x/tools/godoc/index_test.go:137:6: Method 'testIndex' differs only by capitalization to function 'TestIndex' in the same source file
vendor/golang.org/x/tools/godoc/snippet.go:107:24: Method 'NewSnippet' differs only by capitalization to method 'newSnippet' in the same source file
vendor/golang.org/x/tools/godoc/corpus.go:116:6: Method 'NewCorpus' differs only by capitalization to function 'newCorpus' in vendor/golang.org/x/tools/godoc/index_test.go
vendor/golang.org/x/tools/godoc/server.go:741:24: Method 'serveFile' differs only by capitalization to method 'ServeFile' in the same source file
vendor/golang.org/x/tools/godoc/godoc.go:912:24: Method 'WriteNode' differs only by capitalization to method 'writeNode' in the same source file
vendor/golang.org/x/tools/go/ssa/builder.go:2262:19: Method 'build' differs only by capitalization to method 'Build' in the same source file
```

PR #17, opened 2018-06-10
empty-block: ignore checking blocks of funcs and func literals

chavacava:
I've modified `empty-block` to ignore empty blocks of: functions and function literals.  They are a very common idiom.


mgechev:
@chavacava makes sense. Would you add a test case with a function with a receiver for completeness?

PR #22, opened 2018-06-14
Set error code to 1 by default

mfuentesg:
The status code is used for the majority of CIs.
Setting this value by default remove the need to have a configuration file.


mgechev:
Hey @mfuentesg, this is just a sample `toml` file with the default settings. By updating the error code you'll still not get the desired behavior. Also, to enforce errors for the failures, you'll have to set severity error for your rules.

PR #23, opened 2018-06-21
New rule: get-return

chavacava:
Warns on getters  that do not yield any result


rule/get-return.go
```diff
@@ -0,0 +1,71 @@
...
+
+type lintReturnRule struct {
+	onFailure func(lint.Failure)
+}
+
+func isGetter(name string) bool {
+	if strings.HasPrefix(strings.ToUpper(name), "GET") {
+		if len(name) > 3 {
+			c := name[3]
+			return (c >= 'A' && c <= 'Z')
```
mgechev: Why ignoring special chars?

mgechev:
Love it! Left one comment only.

rule/get-return.go
chavacava: `isGetter` is not perfect in determining if a function is a getter or not. Currently it just checks for patterns like `getSomething` or `GetSomething`. Do you think it needs to also check for special characters after `get`?   We can simply check that the fourth character is not lowercase?

```go
if strings.HasPrefix(strings.ToUpper(name), "GET") {
	if len(name) > 3 {
		c := name[3]
		return (c >= 'a' && c <= 'z')
```

mgechev: Oh what I meant was if it'll work with:

```
get_stuff
```

chavacava: Ok, I see what you mean. I will modify the function to cope with such cases

chavacava: I've searched on Docker, Kubernetes, and Go code bases and I did not found getters using _ or other character after the `get` prefix
I've modified to `isGetter` to check that the fourth character is not in [a-z].

PR #26, opened 2018-06-24
New rule deep-exit

chavacava:
`deep-exit` looks for program exits in functions other than `main()` or `init()`

Opinionated guidelines suggest that programs should exit only from `main()` or `init()` functions.


fixtures/deep-exit.go
```diff
@@ -0,0 +1,31 @@
...
+    log.Fatalln("v ...interface{}")
+}
+
+func bar() {
+    log.Fatal(1) // MATCH /calls to log.Fatal function should be made only in main() or init() functions/
+}
+
+func bar2() {
+    syscall.Exit(1) // MATCH /calls to syscall.Exit function should be made only in main() or init() functions/
+}
```
mgechev: newline

rule/deep-exit.go
```diff
@@ -0,0 +1,85 @@
...
+	if stmt, ok := node.(*ast.FuncDecl); ok {
+		w.updateIgnore(stmt)
+		return w
+	}
+
+	if w.ignore {
+		return w
+	}
+
+	if se, ok := node.(*ast.ExprStmt); ok {
```
mgechev: It'll be easier to follow if it's:

```go
```
if se, ok := node.(*ast.ExprStmt); !ok {
  return nil
}
if ce, ok := se.X.(*ast.CallExpr); !ok {
  return nil
}
...

rule/deep-exit.go
```diff
@@ -0,0 +1,85 @@
...
+		if ce, ok := se.X.(*ast.CallExpr); ok { // it's a function call
+			if fc, ok := ce.Fun.(*ast.SelectorExpr); ok {
+				if id, ok := fc.X.(*ast.Ident); ok {
+					fn := fc.Sel.Name
+					pkg := id.Name
+					if w.exitFunctions[pkg][fn] { // it's a call to an exit function
```
mgechev: `w.exitFunctions[pkg]` could be `nil`

chavacava: :+1:

rule/deep-exit.go
chavacava: :+1:

fixtures/deep-exit.go
chavacava: :+1:

rule/deep-exit.go
```diff
@@ -0,0 +1,95 @@
...
+
+	fn := fc.Sel.Name
+	pkg := id.Name
+	if w.exitFunctions[pkg] != nil && w.exitFunctions[pkg][fn] { // it's a call to an exit function
+		w.onFailure(lint.Failure{
+			Confidence: 1,
+			Node:       ce,
+			Category:   "bad practice",
+			URL:        "#deep-exit",
+			Failure:    fmt.Sprintf("calls to %s.%s function should be made only in main() or init() functions", pkg, fn),
```
mgechev: Can we try to remove the passive:

`call %s.%s only in main() and init() functions`

PR #28, opened 2018-07-03
New rule: unused-parameter

chavacava:
Suggests to rename or remove unused function parameters


rule/unused-param.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	}
+	return result
+}
+
+func checkUnusedParams(w lintUnusedParamRule, params map[string]bool, n *ast.FuncDecl) {
+	for k, v := range params {
+		if v {
+			w.onFailure(lint.Failure{
+				Confidence: 0.8, // confidence is not 1.0 because of shadow variables
```
mgechev: How hard do you think it'll be to take into consideration variable shadowing?
- We already have the base case (no local variables with same name & parameter not used)
- It's easy if we find that given parameter is used before there's a declaration of a variable with the same name
- While visiting blocks, we should keep track of the lexical scope as a stack of sets (or `map[string]bool`):
    - If there's no local variable with the same name as a parameter used in any nested block, we don't mark it (we traverse the stack and check if a variable in any block with the same name as the param exists)
    - Otherwise, we do

A complication I see are short forms, such as:

```go
if param := fn(); predicate(param) {
  // do stuff
}
```

Am I missing something?

chavacava: Following your suggestions, I'm working in a version that tracks scopes. The handling of short forms seems to be not so hard.

chavacava: Please check the last committed version. It handles cases like
```go
func f0(param int) {
	param := param
}

func f1(param int) { // MATCH /parameter 'param' seems to be unused, consider removing or renaming it as _/
	if param := fn(); predicate(param) {
		// do stuff
	}
}

func f2(param int) { // MATCH /parameter 'param' seems to be unused, consider removing or renaming it as _/
	switch param := fn(); param {
	default:
		
	}
}

func f3(param myStruct) { 
	a:= param.field
}

func f4(param myStruct, c int) { // MATCH /parameter 'c' seems to be unused, consider removing or renaming it as _/ 
	param.field = "aString"
	param.c = "sss" 
}
```

fixtures/unused-param.go
```diff
@@ -0,0 +1,98 @@
...
+
+func f4(param myStruct, c int) { // MATCH /parameter 'c' seems to be unused, consider removing or renaming it as _/ 
+	param.field = "aString"
+	param.c = "sss" 
+}
+
+func f5(a int, _ float) { // MATCH /parameter 'a' seems to be unused, consider removing or renaming it as _/
+	fmt.Printf("Hello, Golang\n")
+	{
+		if true {
```
mgechev: My guess is that this works with while, if, select, etc. blocks?

rule/unused-param.go
```diff
@@ -0,0 +1,207 @@
...
+
+type scope struct {
+	vars map[string]bool
+}
+
+func newScope() scope {
+	return scope{make(map[string]bool, 0)}
+}
+
+func (s *scope) addVar(exps []ast.Expr) {
```
mgechev: Maybe `addVars` since we add all the variablea from an expression?

mgechev:
@chavacava tested it over `vendor` directory and at first, it looks fine. I'll take another look tomorrow. Looking forward to your test too!

mgechev:
@chavacava oh, found something:

```go
func (c *chanList) remove(id uint32) {
  id -= c.offset
  c.Lock()
  if id < uint32(len(c.chans)) {
     c.chans[id] = nil
  }
  c.Unlock()
}
```

mgechev:
I saw we have `+=` and others handled.  How about the expression in the if statement? Why the rule didn't detect that the parameter was used there?

chavacava:
The previous version of the rule reported `id` as unused in the following code
```go
func (c *chanList) remove(id uint32) {
  id -= c.offset
  c.Lock()
  if id < uint32(len(c.chans)) {
     c.chans[id] = nil
  }
  c.Unlock()
}
```
because in the assignment 
```go
id -= c.offset
```
the rule did not take into account the fact that the new value of `id` depends on its old value (it did not see `id -= c.offset` as an equivalent of `id = id - c.offset`. The assignment is considered as the definition of a variable `id` that shadows parameter `id`, then the reference to `id` in the `if-then` is considered as reference to the shadowing variable and not to the parameter.

The current version works in the same way, for example, it will report `id` as unused in the following case:
```go
func f(id int) {
	id = 1 // standard assignment
	// ...
	if id > 0 {
		// ...
	}
}
```
Formally, there is no new variable shadowing parameter `id` but the original parameter's _value_ is actually ignored (not used). What do you think?

PR #29, opened 2018-07-07
Fix rule for documentation of exported types

mgechev:
The rule was failing in the following case:

```go
// A is an awesome type
type A = int
```

...because revive was skipping the leading `A`, considering it as an article.

@chavacava would you check this out?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ language: go
 go: "1.10.x"
 
 install: make install
-script: make build && make test.all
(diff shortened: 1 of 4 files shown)


chavacava:
@mgechev, it seems Ok. I've checked against Grafana and Prometheus code base.

Note: in Prometheus I've found 2 comments of the form _**This** TypeName ..._

mgechev:
Thanks for the note! I'll add this article as well.

PR #35, opened 2018-07-14
New rule: add-constant

chavacava:
Suggests creating constant for magic numbers and string literals

Rule configuration is of the form

```toml
[rule.add-constant]
  arguments = [{maxLitCount = "2",allowStrs ="\"\"",allowInts="0,1",allowFloats="0.0,1.0"}]
```


mgechev:
I'm curious, why did you decide to allow a configurable number of acceptable string literal repetitions but chose to have one as a strong limit for other types?

PR #38, opened 2018-07-15
New rule: unreachable-code

chavacava:
Warns on unreachable code like (from `github.com/BurntSushi/toml/cmd/toml-test-encoder/main.go`)

```go
func untag(typed map[string]interface{}) interface{} {
	t := typed["type"].(string)
	v := typed["value"]
	switch t {
	case "string":
		...
	case "integer":
		...
	}
	log.Fatalf("Unrecognized tag type '%s'.", t)
	panic("unreachable")
}
```


mgechev:
Looks good! What do you think about this case:

```go
function foo() int {
  fmt.Fatalf("%s", "About to fail")
  return 0
}
```

The code won't compile if we don't allow the return, although it's unreachable.

chavacava:
Yes, I saw cases like that on Prometheus and Kubernetes. I'll modify the rule to skip this cases.
(IMHO it is a _not so nice_ constraint of the GO compiler)

PR #40, opened 2018-07-17
simpler (new rule)

chavacava:
This rule proposes code simplifications:
- unnecessary `return` and `break` statements
- unnecessary comparisons with boolean constants

Examples (from k8s vendor):

```go
func (p *UpdateLoadBalancerParams) SetFordisplay(v bool) {
	if p.p == nil {
		p.p = make(map[string]interface{})
	}
	p.p["fordisplay"] = v
	return   // omit unnecessary return statement
}
``` 

```go
func yaml_emitter_need_more_events(emitter *yaml_emitter_t) bool {
	if emitter.events_head == len(emitter.events) {
		return true
	}
	var accumulate int
	switch emitter.events[emitter.events_head].typ {
	case yaml_DOCUMENT_START_EVENT:
		accumulate = 1
		break // omit unnecessary break at the end of case clause
	case yaml_SEQUENCE_START_EVENT:
		accumulate = 2
		break // omit unnecessary break at the end of case clause
	case yaml_MAPPING_START_EVENT:
		accumulate = 3
		break // omit unnecessary break at the end of case clause
	default:
		return false
	}

	if s.eof == true { // omit comparison with boolean constants
		s.position.Line++
		s.position.Column = 0
		s.eof = false
	}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,6 +221,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | `deep-exit`           |  n/a   | Looks for program exits in funcs other than `main()` or `init()` |    no    |  no   |
 | `unused-parameter`    |  n/a   | Suggests to rename or remove unused function parameters          |    no    |  no   |
 | `unreachable-code`    |  n/a   | Warns on unreachable code                                        |    no    |  no   |
+| `simpler`             |  n/a   | Proposes code simplifications                                    |    no    |  no   |
 
 ## Available Formatters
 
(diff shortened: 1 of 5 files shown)


mgechev:
@chavacava maybe we can rename the rule to make it more obvious what it does? I'm a bit unhappy with the naming of the rules inherited from golint, but I didn't come up with anything better. So far we did a good job naming the new rules accurately.

chavacava:
Ok. I'll close the PR and rework the rule.

PR #44, opened 2018-07-21
flag-parameter (new rule)

chavacava:
The rule `flag-parameter` warns on boolean parameters that create a [control coupling ](https://en.wikipedia.org/wiki/Coupling_(computer_programming)#Procedural_programming)

### DIFF
--- a/rule/flag-param.go
+++ b/rule/flag-param.go
@@ -1,0 +1,104 @@
+package rule
+
+import (
+	"fmt"
+	"github.com/mgechev/revive/lint"
+	"go/ast"
+)
+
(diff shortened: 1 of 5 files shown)


rule/flag-param.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+
+	uses := pick(ifStmt.Cond, fselect, nil)
+
+	if len(uses) < 1 {
+		return w
+	}
+
+	w.linter.onFailure(lint.Failure{
+		Confidence: 0.8,
```
mgechev: Why's it .8?

PR #45, opened 2018-07-26
unnecessary-stmt (new rule)

chavacava:
Suggests removing or simplifying unnecessary statements like `break`s at the end of `case`s or `return` at the end of bodies of functions returning nothing.


fixtures/unnecessary-stmt.go
```diff
@@ -0,0 +1,49 @@
...
+		a++
+	}
+
+loop:
+	for {
+		switch a {
+		case 1:
+			a++
+			println("one")
+			break // MATCH /omit unnecessary break at the end of case clause/
```
mgechev: I am not sure it's safe removing this break statement; if we do remove it we can enter other case clause in other scenarios.

chavacava: Hi, what do you mean with _other scenarios_? 

Both versions of the code, with `break` (https://play.golang.org/p/JJArilQ1jbZ) and without (https://play.golang.org/p/OrmF3W3PNGK) yield the same output.

mgechev: 👍 never mind

PR #47, opened 2018-07-28
struct-tag (new rule)

chavacava:
Checks common struct tags like `json`,`xml`,`yaml` and warns when using non-standard options or when expected types and values mismatch


chavacava:
I'm glad you like the rule.

PR #53, opened 2018-08-17
bool-comp (new rule)

chavacava:
Warns on comparisons with boolean constants like `bar == true`, `result != true`, `flag == false`


chavacava:
I'm closing the PR to do a little rework to the rule.

PR #54, opened 2018-08-18
bool-literal-in-expr (new rule)

chavacava:
Suggests removing boolean literals from logic expressions like `bar == true`, `arg == false`, `r != true`

It warns when logical expression always evaluates to the same value. For example it will detect that `false && boolExpr` and `boolExpr || true` always evaluate to `false` and `true` respectively.


chavacava:
Test case added.

PR #61, opened 2018-09-15
Add white & black lists for var-naming

### LINKED ISSUE #41: Feature Request: Customize initialisms or remove "ID" from the common list
In short, I'm looking for alternatives to `golint`, particularly because of this issue:
https://github.com/golang/lint/issues/89

There are some options to implement this feature.
1. We could provide a "whilelist" of allowed initialisms, such as `Id`, in order to name variables such as `projectId` via configuration.  This is the most flexible option, but may lead to other initialisms than `Id` to be whitelisted, spreading inconsistency across Go projects.
2. Remove `ID` completely from the [initialism list](https://github.com/mgechev/revive/blob/5248ad8ed955d22005b47a1dbfdfa720b0e77ecb/lint/utils.go#L79). The comment on source code, similarly to Golint, states that "Freudian code is rare". However this rule creates warnings for several projects, such as [here](https://github.com/golang/protobuf/issues/156)
3. Refactor initialisms into its own rule other than `var-naming`, so it can be enabled/disabled as whole.

I've tried disabling the `var-naming` rule, but other effective idiomatic rules get disabled as well, so that's not an option.

**Additional context**
The lack of flexibility in Golint rules is something that this project can offer, along the speed.

I'd more than happy to give it a try at this myself, if applicable.

mgechev:
This PR introduces a whitelist & blacklist of initialisms for the
`var-naming` rule.

Now the rule can be configured with:

```toml
[rule.var-naming]
  arguments = [["ID"], ["VM", "BAR"]]
```

This way, the linter will ignore `customId` but will throw on `customVm` or `customBar`.

Fix #41


chavacava:
Hi @mgechev, it looks good to me. :+1:

PR #62, opened 2018-09-17
New rule: return-argument-limit

xuri:
The rule `return-argument-limit` specifies the maximum number of arguments a function can return.


rule/return-argument-limit.go
```diff
@@ -0,0 +1,65 @@
...
+// ReturnArgumentsLimitRule lints given else constructs.
+type ReturnArgumentsLimitRule struct{}
+
+// Apply applies the rule to given file.
+func (r *ReturnArgumentsLimitRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	if len(arguments) != 1 {
+		panic(`invalid configuration for "return-argument-limit"`)
+	}
+
+	total, ok := arguments[0].(int64) // Alt. non panicking version
```
chavacava: `total` could be renamed to `max`
(the same for the field `total` of the `lintReturnArgumentsNum` struct)

rule/return-argument-limit.go
```diff
@@ -0,0 +1,65 @@
...
+
+import (
+	"fmt"
+	"go/ast"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// ReturnArgumentsLimitRule lints given else constructs.
+type ReturnArgumentsLimitRule struct{}
```
chavacava: Formally, functions do not return _arguments_ but _results_ (https://golang.org/ref/spec#Signature). 
What do you think about renaming the rule to something like `FunctionResultsLimit` ?

xuri:
Thanks for your suggestion, I have fixed it.

chavacava:
Hi @xuri, please include a test for the rule. You need to add a `return-value-limit_test.go` under `test` directory. You can take `argument-limit_test.go` as an example of a rule using an integer parameter.
Then you can check that test pass using `go test` (it is also safe to check that the test does not pass if it should not)

Please also update the description of the rule in `README.md` to remove the reference to _arguments_.

xuri:
@chavacava done.

rule/function-result-limit.go
```diff
@@ -0,0 +1,65 @@
...
+
+// Apply applies the rule to given file.
+func (r *FunctionResultsLimitRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	if len(arguments) != 1 {
+		panic(`invalid configuration for "function-result-limit"`)
+	}
+
+	max, ok := arguments[0].(int64) // Alt. non panicking version
+	if !ok {
+		panic(`invalid value passed as return results number to the "function-result-limit" rule`)
```
chavacava: We can provide more information about what the rule received as argument (the type?)

rule/function-result-limit.go
```diff
@@ -0,0 +1,65 @@
...
+type FunctionResultsLimitRule struct{}
+
+// Apply applies the rule to given file.
+func (r *FunctionResultsLimitRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	if len(arguments) != 1 {
+		panic(`invalid configuration for "function-result-limit"`)
+	}
+
+	max, ok := arguments[0].(int64) // Alt. non panicking version
+	if !ok {
```
chavacava: We can check that `max` is not negative

mgechev:
LGTM! Will let @chavacava take a final look & merge.

PR #64, opened 2018-09-17
Add support for go 1.11 dependency management

z0mbie42:


mgechev:
Can we drop `Gopkg.*`?

PR #65, opened 2018-09-17
New formatter: unix

z0mbie42:
here an example:

```bash
$ revive -config revive.toml -formatter unix
main.go:24:9: [errorf] should replace errors.New(fmt.Sprintf(...)) with fmt.Errorf(...)
```

I think it's better than default because it clearly shows which rule is violated.

Furthermore it would facilitate result parsing by other tools which would enable better linter composition with revive, for example with the following regexp:
```
^(.+):(\d+):(\d+): \[(.+)\] (.+)$
```

### DIFF
--- a/config.go
+++ b/config.go
@@ -72,6 +72,7 @@ var allFormatters = []lint.Formatter{
 	&formatter.JSON{},
 	&formatter.NDJSON{},
 	&formatter.Default{},
+	&formatter.Unix{},
 	&formatter.Checkstyle{},
 }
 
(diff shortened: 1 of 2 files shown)


mgechev:
Would you add it to the list of formatters in the readme?

z0mbie42:
Unfortunately I don't have a mac and I wasn't able to reproduce such beautiful screenshots.

I can add it, but without screenshot

mgechev:
No worries, I can add it to the list in a separate PR.

PR #68, opened 2018-09-17
Add the imports-blacklist rule

z0mbie42:


rule/imports-blacklist.go
```diff
@@ -0,0 +1,74 @@
...
+func (r *ImportsBlacklistRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+	blacklist := make([]string, len(arguments))
+
+	for i, arg := range arguments {
+		argStr, ok := arg.(string)
+		if !ok {
+			panic(fmt.Sprintf("Invalid argument to the imports-blacklist rule. Expecting a string, got %T", arg))
+		}
+		argStr = strings.ToLower(strings.TrimSpace(argStr))
```
mgechev: For a case-sensitive file system, upper cases will matter.

rule/imports-blacklist.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/mgechev/revive/lint"
+)
+
+// ImportsBlacklistRule lints given else constructs.
+type ImportsBlacklistRule struct{}
+
+// Apply applies the rule to given file.
+func (r *ImportsBlacklistRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+	blacklist := make([]string, len(arguments))
```
mgechev: We can store the blacklist as a set - `map[string]bool` so that we can have the algorithm in `Visit` below with `O(n)` instead of `O(n*m)`.

rule/imports-blacklist.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+type blacklistedImports struct {
+	file      *lint.File
+	fileAst   *ast.File
+	onFailure func(lint.Failure)
+	blacklist []string
+}
+
+func (w blacklistedImports) Visit(n ast.Node) ast.Visitor {
```
chavacava: Parameter `n` seems to be not used, then the utility of a struct implementing the visitor interface is lost.

Why not implementing the rule logic directly in the `Apply` method?

(@mgechev `blank-imports` has a similar implementation, may be there is a reason that I'm not seeing)

mgechev: What do you mean?

rule/imports-blacklist.go
z0mbie42: I've added this lately because I don't see any benefits to be case sensitive, but It can helps to be insensitive:
Imagine you have the package: `github.com/MySuperUser/pkg`

using my _limited_ memory I may add
```toml
[rule.imports-blacklist]
arguments = ["github.com/mysuperuser/pkg"]
```
to the configuration file which will result to the package not being blacklisted if we are case sensitive.

And as `github.com/mysupperuser/pkg` can't exist when `github.com/MySuperUser/pkg` exists I don't see how it can lead to situations where the bad package is blacklisted.

rule/imports-blacklist.go
chavacava: A `map[string]struct{}` is also an alternative with a smaller footprint (https://play.golang.org/p/Mv1SgehJvbh)

rule/imports-blacklist.go
mgechev: In a case-sensitive file system, both packages may exist. Sometime back, there were issues caused by [this](https://github.com/golang/dep/issues/1010).

rule/imports-blacklist.go
chavacava: I mean that this rule can be implemented as follows (without a visitor struct):

```go
func (r *ImportsBlacklistRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
	var failures []lint.Failure
        if  lint.File.IsTest() {
            return failures
        }
	blacklist := make([]string, len(arguments))
	for i, arg := range arguments {
		argStr, ok := arg.(string)
		if !ok {
			panic(fmt.Sprintf("Invalid argument to the imports-blacklist rule. Expecting a string, got %T", arg))
		}
		argStr = strings.ToLower(strings.TrimSpace(argStr))
		if len(argStr) > 2 && argStr[0] != '"' && argStr[len(argStr)-1] != '"' {
			argStr = fmt.Sprintf(`"%s"`, argStr)
		}
		blacklist[i] = strings.ToLower(strings.TrimSpace(argStr))
	}
	
        for _, is := range file.AST.Imports {
		if is.Path != nil {
			for _, blacklisted := range blacklist {
				if strings.ToLower(is.Path.Value) == blacklisted  {
					failures = append(failures,lint.Failure{
						Confidence: 1,
						Failure:    fmt.Sprintf("should not use the following blacklisted import: %s", is.Path.Value),
						Node:       is,
						Category:   "imports",
					})
				}
			}
		}
	}

	return failures
}

```

rule/imports-blacklist.go
z0mbie42: Wasn't aware, thanks for sharing.
Fixed in # 35e4c10

PR #69, opened 2018-09-18
Add missed max-public-structs rule to all list

psapezhko:
A just faced with `cannot find rule: max-public-structs` when I tried to use this rule.


mgechev:
Thanks for the catch!

Issue #59, opened 2018-09-12
panic: assertion failed [recovered]


l2dy:
**Describe the bug**
revive panics.

**To Reproduce**
1. I updated revive `go get -u github.com/mgechev/revive`
2. I run it with the following flags & configuration file:

```shell
# flags

~/.gotools/bin/revive -config=$HOME/.revive.toml ./...
```

```toml
# config file
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.8
errorCode = 0
warningCode = 0

[rule.blank-imports]
[rule.context-as-argument]
[rule.context-keys-type]
[rule.dot-imports]
[rule.error-return]
[rule.error-strings]
[rule.error-naming]
#[rule.exported]
[rule.if-return]
[rule.increment-decrement]
#[rule.var-naming]
[rule.var-declaration]
[rule.package-comments]
[rule.range]
#[rule.receiver-naming]
[rule.time-naming]
[rule.unexported-return]
[rule.indent-error-flow]
[rule.errorf]
```

**Expected behavior**
No panic.

**Logs**
```
[...]
vendor/github.com/apache/thrift/lib/go/thrift/server_socket.go:92:9: if block ends with a return statement, so drop this else and outdent its block (move short variable declaration to its own line if necessary)
panic: assertion failed [recovered]
	panic: assertion failed

goroutine 1315 [running]:
go/types.(*Checker).handleBailout(0xc0005b6870, 0xc008cab800)
	/usr/local/go/src/go/types/check.go:236 +0x98
panic(0x12b3600, 0x1370730)
	/usr/local/go/src/runtime/panic.go:513 +0x1b9
go/types.assert(...)
	/usr/local/go/src/go/types/errors.go:18
go/types.(*Checker).recordTypeAndValue(0xc0005b6870, 0x0, 0x0, 0x3, 0x1374960, 0xc02fead160, 0x0, 0x0)
	/usr/local/go/src/go/types/check.go:281 +0x279
go/types.(*Checker).exprInternal(0xc0005b6870, 0xc02fed4e00, 0x13755a0, 0xc00a1ca800, 0x1374960, 0xc02fead160, 0x100b3c3)
	/usr/local/go/src/go/types/expr.go:1164 +0x17c4
go/types.(*Checker).rawExpr(0xc0005b6870, 0xc02fed4e00, 0x13755a0, 0xc00a1ca800, 0x1374960, 0xc02fead160, 0x100ba38)
[...]

PR #70, opened 2018-09-21
fix issue [#59]

chavacava:
fix candidate for issue #59

### DIFF
--- a/lint/package.go
+++ b/lint/package.go
@@ -77,13 +77,29 @@ func (p *Package) TypeCheck() error {
 		anyFile = f
 		astFiles = append(astFiles, f.AST)
 	}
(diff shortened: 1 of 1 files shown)


mgechev:
Looks good. I'm slightly concerned that all rules swallow the error returned by `TypeCheck`. This way, rules using type checking may not report existing linting errors.

We should follow closely how the Go team handles the issue so we can react.

chavacava:
> I'm slightly concerned that all rules swallow the error returned by TypeCheck. 

Yes, this fix is somewhat brutal in that aspect

> We should follow closely how the Go team handles the issue so we can react.

Absolutely! I did not check the go repo to see if there is already an issue related to this problem

PR #72, opened 2018-09-22
Rule range var in closure

chavacava:
Warns if a range value is used in a closure dispatched as goroutine.

Inspired by http://devs.cloudimmunity.com/gotchas-and-common-mistakes-in-go-golang/index.html#closure_for_it_vars


chavacava:
duplicated with #75

PR #75, opened 2018-09-27
New rule: range-loop-var

xuri:
The rule `range-loop-var` disallows incorrect uses of range loop variables in closures. checking whether its index or value variables are used unsafely inside goroutines or deferred function literals.


mgechev:
Looks like @chavacava opened a PR with something similar https://github.com/mgechev/revive/pull/72.

chavacava:
> Looks like @chavacava opened a PR with something similar #72.

Yes, both rules address the same case.

@xuri's rule looks more complete because it includes `for` variables within `go func` and `defer func (mine only checks `range` values within `go func`). I'll close my PR.

@xuri: I think we can check all `go` and `defer` statements (not just when they are the last statement of the loop body). After linting some repos (kubernetes, tidb, prometheus) there are less false positive than missed problems.

README.md
```diff
@@ -271,6 +271,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | `redefines-builtin-id`|  n/a   | Warns on redefinitions of builtin identifiers                    |    no    |  no   |
 | `function-result-limit` |  int | Specifies the maximum number of results a function can return    |    no    |  no   |
 | `imports-blacklist`   | []string | Disallows importing the specified packages                     |    no    |  no   |
+| `range-loop-var`      | n/a    | Disallows incorrect uses of range loop variables in closures     |    no    |  no   |
```
chavacava: Rule's name and description do not truly reflect the scope of the rule (i.e. for loop values and defer statements are also checked)

fixtures/range-loop-var.go
```diff
@@ -0,0 +1,24 @@
...
+	myDict["A"] = 1
+	myDict["B"] = 2
+	myDict["C"] = 3
+	for key, value := range myDict {
+		defer func() {
+			fmt.Printf("Index: %d\n", key)   // MATCH /loop variable key captured by func literal/
+			fmt.Printf("Value: %s\n", value) // MATCH /loop variable value captured by func literal/
+		}()
+	}
+}
```
chavacava: We can add more test cases with examples like:
```go
for i, newg := range groups {
		go func(newg int) { 
			newg.run(m.opts.Context,i) / MATCH /loop variable i captured by func literal/
		}(newg)
}
```
```go
for i, newg := range groups {
		newg := newg
		go func() { 
			newg.run(m.opts.Context,i) / MATCH /loop variable i captured by func literal/
		}()
}
```

rule/range-loop-var.go
```diff
@@ -0,0 +1,111 @@
+package rule
```
chavacava: Please lint the code with `revive` using the recommended configuration (see README.MD) and remove all warnings/errors

rule/range-loop-var.go
```diff
@@ -0,0 +1,111 @@
...
+	lit, ok := last.Fun.(*ast.FuncLit)
+	if !ok {
+		return w
+	}
+	ast.Inspect(lit.Body, func(n ast.Node) bool {
+		id, ok := n.(*ast.Ident)
+		if !ok || id.Obj == nil {
+			return true
+		}
+		if lit.Type == nil {
```
chavacava: It seems that this condition (`lit.Type == nil`) evaluates to the same result all over the recursive traversal of the function literal body. If that is the case, we can do this check just before calling `ast.Inspect`

xuri:
Hi @chavacava, thanks for your advice, I have fixed it.

PR #76, opened 2018-09-29
waitgroup-by-value (new rule)

chavacava:
Warns on functions taking `sync.WaitGroup` as a by-value parameter.

Passing a `sync.WaitGroup` as a value to a function will create a copy of the `sync.WaitGroup` then all modifications (`wg.Add`, `wg.Done`...) performed at function level do not change the original `sync.WaitGroup`.


rule/waitgroup-by-copy.go
```diff
@@ -0,0 +1,65 @@
...
+)
+
+// WaitGroupByCopyRule lints sync.WaitGroup passed by copy in functions.
+type WaitGroupByCopyRule struct{}
+
+// Apply applies the rule to given file.
+func (r *WaitGroupByCopyRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	onFailure := func(failure lint.Failure) {
```
mgechev: This is a very repetitive pattern. I'm wondering if it's possible to abstract this closure somehow.

Issue #82, opened 2018-10-11
Edge cases in empty-lines rule


gsamokovarov:
**Describe the bug**

The following example fails:

```go
func f10(*int) bool {
	if x > 2 {
		return true
	}
}
```

This is because of the trailing rule taking the position if the `if` statement, not the closing of the block into account.

**Expected behavior**

This is a false-positive and should not happen.

PR #83, opened 2018-10-13
Handle multi-line statements edge-cases in the multi-line rule

gsamokovarov:
There were many cases where the multi-line statements at the end of a
block were marked as containing empty lines, while they didn't. Or vice
versa.

You may check-out the tests. They have lots of new examples.


mgechev:
Fixes https://github.com/mgechev/revive/issues/82

PR #86, opened 2018-10-18
New rule: line-length-limit

xuri:


untyped.toml
```diff
@@ -16,3 +16,6 @@
 [rule.range-val-in-closure]
 [rule.waitgroup-by-value]
 [rule.atomic]
+[rule.empty-lines]
+[rule.line-length-limit]
+    arguments = [200]
```
mgechev:
```suggestion
    arguments = [200]

```

mgechev: A trailing newline.

rule/line-length-limit.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(arguments) != 1 {
+		panic(`invalid configuration for "line-length-limit"`)
+	}
+
+	max, ok := arguments[0].(int64) // Alt. non panicking version
+	if !ok {
+		panic(fmt.Sprintf(`invalid value passed as return results number to the "line-length-limit" rule; need int64 but got %T`, arguments[0]))
+	}
+	if max < 0 {
+		panic(`the value passed as return results number to the "line-length-limit" rule cannot be negative`)
```
mgechev: Can we simplify the panic message?

xuri:
@mgechev @chavacava Thanks for your advice, I have fixed it.

chavacava:
It's looks good.

harleenarora:
Hi, This rule is not working with exclude parameter. I have add a pattern in exclude but its still show the error. Here is my code 
name: line-length-limit
          disabled: false
          severity: warning
          arguments: [120]
          exclude:
             - '^type\s+[A-Z][a-zA-Z0-9_]*\s*\{'

chavacava:
Hi @harleenarora, thanks for reporting the issue.

The rule hasn't an exclusion parameter so it's expected to not to take it into account. It seems you are using the rule through `golangci-lint`, you might open an issue there asking for help on how to use the exclusion patterns.

PR #88, opened 2018-10-26
docs: update the documentation

mgechev:
I did some formatting:

- There were new lines in the middle of a few sentences
- Added newlines before code blocks
- Removed trailing spaces

I also dropped statements in the form "This rule checks/does/warns...", when this was clear from the rule's explanation.


mgechev:
Btw, I'll check how I can make `https://revive.run/r/#hash` to redirect to the correct URL.

chavacava:
Great job. tnx

PR #90, opened 2018-10-31
call-to-gc (new rule)

### LINKED ISSUE #89: Add a rule to warn on explicit triggers of the garbage collector
Calling `runtime.GC()` to force garbage collection is rarely a good idea. In very few occasions triggering the GC is justified (when benchmarking for example) but most of the time is a symptom of the developer trying (and failing) to be smarter than the runtime.

We can add a rule to warn on the presence of `runtime.GC()`

chavacava:
Closes #89

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ Here's how `revive` is different from `golint`:
     - [Friendly](#friendly)
     - [Stylish](#stylish)
     - [Default](#default)
+    - [Plain](#plain)
+    - [Unix](#unix)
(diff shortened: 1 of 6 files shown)


mgechev:
As an improvement, we can type check and see if `GC` is from the correct type to prevent false negatives, but this should be fine!

chavacava:
> As an improvement, we can type check and see if GC is from the correct type to prevent false negatives, 

I've thought to do it but I've decided that the number of possible false positives does not worth the price of type-checking

PR #99, opened 2019-01-17
fixes issue #98

### LINKED ISSUE #98: disable-(next-)-line does not work for long comment lines
**Describe the bug**
`revive:disable-line` and `revive:disable-next-line` do not work with rule `line-length-limit` when applied to comments, although they work fine with lines of code.

**To Reproduce**

```shell
# flags

revive -config myconfig.toml
```

```toml
# config file

[rule.line-length-limit]
  arguments = [80]
```

```go
package main

import "fmt"

// revive:disable-next-line:line-length-limit
// The length of this comment line is over 80 characters, this is bad for readability.

func main() {
	// revive:disable-next-line:line-length-limit
	fmt.Println("This line is way too long. In my opinion, it should be shortened.")
}
```

**Expected behavior**
`revive` should output nothing, since the line length rule was disable for both offending lines.

 - OS: Mint 19.1 (based on Ubuntu 18.04)
 - Version of Go: 1.11.4

chavacava:
Closes #98

Warn: I've updated the test fixture for the `line-length-limit` rule but `revive`'s testing framework is not able to check failures on comments.

### DIFF
--- a/lint/file.go
+++ b/lint/file.go
@@ -127,7 +127,7 @@ type enableDisableConfig struct {
 }
 
 func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
-	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:|\s|$)`)
+	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:)?([^\s]*)?(\s|$)`)
 
 	enabledDisabledRulesMap := make(map[string][]enableDisableConfig)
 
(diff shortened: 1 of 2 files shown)


lint/file.go
```diff
@@ -127,7 +127,7 @@ type enableDisableConfig struct {
 }
 
 func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
-	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:|\s|$)`)
+	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:)?([^\s]*)?(\s|$)`)
```
mgechev: This would also match `  revive:disable-next-line:`, which I suppose is okay. We should be able to handle the case when the rule name is empty.

Let me know if I'm missing something.

chavacava: Hi, yes the regexp matches `revive:disable-next-line:` and `revive:disable-next-line`. 

The modification handles the case of an empty rule name.
For example, with:

```go
// revive:disable-next-line
// The length of this comment line is over 80 characters, this is bad for readability.

func toto() {
	// revive:disable-next-line:
	fmt.Println("This line is way too long. In my opinion, it should be shortened.")
}
```
it works as expected.

PR #103, opened 2019-01-23
Do not require additional newline for disabling multiple rules

### LINKED ISSUE #102: lint: Disabling multiple rules require additional newline
**Describe the bug**
After commit https://github.com/mgechev/revive/commit/c878d3090df36b2b57f9427ec7ea29be528e19ef disabling multiple rules at once require additional newline between comments. 

```go
// disable:revive:rule1
/* there should be newline, otherwise rule2 will be still enabled*/
// disable:revive:rule2
```

**To Reproduce**
2. Can be tested with docker only in isolated environment with default `config.toml`:

```shell
echo '
        FROM golang:alpine
        RUN apk add --no-cache git
        RUN go get -u -v github.com/mgechev/revive
        RUN mkdir -p /go/src/test
        RUN echo "package main" > /go/src/test/main.go && \
                echo "" >> /go/src/test/main.go && \
                echo "// revive:disable:var-naming" >> /go/src/test/main.go && \
                echo "// revive:disable:increment-decrement" >> /go/src/test/main.go && \
                echo "func main() {" >> /go/src/test/main.go && \
                echo "  var invalid_name = 0" >> /go/src/test/main.go && \
                echo "  invalid_name += 1" >> /go/src/test/main.go && \
                echo "  println(invalid_name)" >> /go/src/test/main.go && \
                echo "}" >> /go/src/test/main.go
        RUN revive test' | docker build --no-cache -
```

3. Also can be tested on local machine for this file:

```go
package main

// revive:disable:var-naming
// revive:disable:increment-decrement
func main() {
  var invalid_name = 0
  invalid_name += 1
  println(invalid_name)
}
```

```shell
revive ...
```

**Expected behavior**
Revive should exit with 0 RC and with empty output to STDERR and STDOUT. 

**Logs**
```
main.go:7:3: should replace invalid_name += 1 with invalid_name++
```

 - OS: 
   * darwin
   * linux
 - Version of Go: 
   * go version go1.11.4 darwin/amd64
   * go version go1.11.4 linux/amd64

**Additional context**
Probably caused by new regexp: https://github.com/mgechev/revive/blob/c878d3090df36b2b57f9427ec7ea29be528e19ef/lint/file.go#L130

chavacava:
Closes #102 
Added test on disabling annotations with `next-line` and `line` modifiers. These tests cope with the error case identified by the issue #102

### DIFF
--- a/lint/file.go
+++ b/lint/file.go
@@ -127,7 +127,7 @@ type enableDisableConfig struct {
 }
 
 func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
-	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:)?([^\s]*)?(\s|$)`)
+	re := regexp.MustCompile(`revive:(enable|disable)(?:-(line|next-line))?(?::([^\s]+))?`)
 
 	enabledDisabledRulesMap := make(map[string][]enableDisableConfig)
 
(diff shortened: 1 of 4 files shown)


lint/file.go
```diff
@@ -127,7 +127,7 @@ type enableDisableConfig struct {
 }
 
 func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
-	re := regexp.MustCompile(`^\s*revive:(enable|disable)(?:-(line|next-line))?(:)?([^\s]*)?(\s|$)`)
+	re := regexp.MustCompile(`revive:(enable|disable)(?:-(line|next-line))?(?::([^\s]+))?`)
```
mgechev: This will match: `//random text revive:enable` as well? I'm not sure if we should enable this.

I suppose you had to do this because for:

```
//revive:disable
//revive:enable
```

The comment in the AST will be: `revive:disable revive:enable`?

mgechev: @chavacava 
>Other possible solution is to iterate over all Comments in the CommentGroup. At the Comment level the text is as in the source code (including //)

 This sounds cleaner to me. It'll still be a `O(n)` check but with a bit more manual work, I guess.

chavacava: @mgechev 
I've pushed the fix version that iterates over each comment in the CommentGroup.
This version only supports revive annotations on mono-line comments starting with `//`

I can extend it to cope with annotation on multi-line comments (`/* ... */`) but I need some definitions on what to accept as valid annotations in such comments.

```go
/*  revive:disable */

/* This is a comment
   revive:disable
   that expands over 3 lines */

/*
revive:disable
*/
```

mgechev: I think this is good ATM. Great job ⚡️

Issue #100, opened 2019-01-18
Tagged version


akerl:
Howdy!

In https://github.com/mgechev/revive/issues/66 , it seems the determination was to tag releases w/o publishing release binaries. That said, it doesn't look like there's any tags as-yet.

I'm looking to use Revive in CI builds, where ideally I can lock the version of revive so that a given build is consistent. Right now, I'm locking to the git hash, but that's not ideal.

Is the intention to start tagging releases / provide a changelog, as per that issue?

mgechev:
Let's close this in favor of https://github.com/mgechev/revive/issues/66

Issue #66, opened 2018-09-17
Add SemVer releases


z0mbie42:
**Is your feature request related to a problem? Please describe.**
The idiomatic way to create software is to publish releases when required (fix, new feature, api break...).
There is currently no versioning for `revive`.

**Describe the solution you'd like**

I would like to suggest https://github.com/astrocorp42/rocket to automate GitHub releases.
Thus We can automate releasing and binaries publishing.
Furthermore it would enable automated **Docker** publishing which is great for users of services like drone.io. 

It would requires a `Makefile` (for example: https://github.com/astrocorp42/rocket/blob/master/Makefile), to build `revive`.

**Describe alternatives you've considered**
https://github.com/goreleaser/goreleaser but it mix too many things (build + publishing).

mgechev:
Let's collect some more feedback here before we proceed.

z0mbie42:
The case I encountered today was the following:
I use revive for one of my projects during the tests in the CI pipeline (in a docker container). The pipeline failed because revive was not at the good version and the `unix` formatter was not available because Docker cacher this `go get -u github.com/mgechev/revive` step.

`go get` unfortunately cannot enable to download a specific version of a go program, this is why I think providing binaries for each version is a *very good idea*. (thus we can pin our revive version in our CI pipelines). And there is absolutely no cost to do this.

@chavacava +1 for the changelog idea, but it's really a panacea and very error prone to do it manually and can be easily automated by extracting information from commit messages.
Do you have an idea about what tool to use to automate the changelog generation ?

mgechev:
Let's tag the individual versions without publishing binaries for now. If we see this feature request again, we can reconsider.

mgechev:
1. In the Angular team we have weekly patch releases monthly, major releases, and bi-weekly annual releases. We can introduce a similar schedule. Every new feature will be available in the master branch, so people who want to live on the edge can (feel free to propose a better release workflow)
2. I think we can keep it simple for now without LTS

chavacava:
1. Do you refer to [this schedule](https://angular.io/guide/releases#release-frequency)? LGTM
2. I agree.

Do we start with 1.0.0? When?

mgechev:
We can start tagging releases when https://github.com/mgechev/revive/pull/106 lands. I hope we're able to put some more thoughts into it by then.

albertorestifo:
Any update on this? Having tagged releases would be really helpful

mgechev:
[...]

PR #106, opened 2019-02-01
extract package comments form and exported comments form

### LINKED ISSUE #105: Separate checks for missing comments and comments format
It's OK when linter tells me I should document a package or a method, or a a constant. But I do not like when it tells me **how** I should document my code.

Would be nice to have separate checks for missing comments and comments format

ademenev:
I do not like when linter tells me how to write my documentation

Closes #105

### DIFF
--- a/config.go
+++ b/config.go
@@ -24,9 +24,11 @@ func fail(err string) {
 var defaultRules = []lint.Rule{
 	&rule.VarDeclarationsRule{},
 	&rule.PackageCommentsRule{},
+	&rule.PackageCommentsFormRule{},
 	&rule.DotImportsRule{},
 	&rule.BlankImportsRule{},
 	&rule.ExportedRule{},
+	&rule.ExportedFormRule{},
 	&rule.VarNamingRule{},
 	&rule.IndentErrorFlowRule{},
 	&rule.IfReturnRule{},
--- a/rule/exported-form.go
+++ b/rule/exported-form.go
@@ -1,0 +1,184 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+	"strings"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// ExportedFormRule lints given else constructs.
+type ExportedFormRule struct{}
+
+// Apply applies the rule to given file.
+func (r *ExportedFormRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	if isTest(file) {
+		return failures
+	}
+
+	fileAst := file.AST
+	walker := lintExportedForm{
+		file:    file,
+		fileAst: fileAst,
+		onFailure: func(failure lint.Failure) {
+			failures = append(failures, failure)
+		},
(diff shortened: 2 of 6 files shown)


config.go
```diff
@@ -24,9 +24,11 @@ func fail(err string) {
 var defaultRules = []lint.Rule{
 	&rule.VarDeclarationsRule{},
 	&rule.PackageCommentsRule{},
+	&rule.PackageCommentsFormRule{},
```
mgechev: These rules should be also documented in the readme.

rule/exported-form.go
```diff
@@ -0,0 +1,184 @@
...
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+	"strings"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// ExportedFormRule lints given else constructs.
```
mgechev: This comment need to be rewritten

ademenev:
Hmm, one thing I did not think about is breaking existing setups. Possibly a better approach would be to make `exported` and `package-comments` configurable

mgechev:
@ademenev that's a good point, somehow related to a better release process.

If we ignore backward compatibility of the config file, I'd prefer to go with two new rules, as you proposed. The change makes sense because it enforces better modularity and separation of concerns.

My suggestion would be to first address https://github.com/mgechev/revive/issues/100 and tag 0.1.0. After that merge your change (once there are tests for the new rules and we update the docs + the default set of rules to keep it compatible with golint) and tag 1.0.0-rc.0. After that, once 1.0.0-rc.0 has been well tested by developers, we can release 1.0.0.

What are your thoughts?

cc @chavacava

ademenev:
@mgechev I just started using revive, and warnings about my comment style were annoying, so I quickly put together some code to mute them. Now, after thinking more, I gather a better solution would be to not introduce new rules, but instead make existing ones configurable.

So I suggest that you drop this PR, and I will prepare another one, with tests and docs. After 0.1.0 I'ii submit another PR. Would that work?

mgechev:
Why you think making the rules configurable would be a better approach?

ademenev:
The reason is keeping BC.

Introducing separate rules will immediately make files that use disable:export comments generate warning - people expected that to suppress both checks, but only one will be suppressed.

mgechev:
Yes, backwards compatibility is the reason for my proposal: 0.1.0 and after that 1.0.0 after your PR goes in. @chavacava can share his opinion as well.

ademenev:
My point is : making the rules configurable adds desired feature without breaking BC

chavacava:
As pointed by @ademenev, making the rule configurable preserves BC. 
We need to work on #66 to have a clear definition of revive's releases policies

mgechev:
Just to add more context to this comment https://github.com/mgechev/revive/pull/106#issuecomment-460097897

If we introduce SemVer to revive, we can release the current set of changes as 0.1.0. We can release the rule refactoring (i.e., splitting into multiple rules) as 1.0.0. This is fine, because a major release in SemVer allows introducing breaking changes. Everyone who hasn't migrated yet, can rely on 0.1.0 and preserve the rule behavior.

PR #110, opened 2019-03-18
fix error return rule to allow multiple error return values

yangdiangzb:
please check golint related updates here: golang/lint@8379672

When returning multiple values with an error,
lint checks if the error is the last return value.
But the implementation actually is checking for all return values
except for the last one, and throw the alert if it found an error.

There is a (edge) case where some function returning more than one error
is getting a false positive even when the last return value is an error.

This patch adds an early check, to see if the last return value is an error
and if so, it will pass silently.

### DIFF
--- a/rule/error-return.go
+++ b/rule/error-return.go
@@ -47,6 +47,9 @@ func (w lintErrorReturn) Visit(n ast.Node) ast.Visitor {
 	if len(ret) <= 1 {
 		return w
 	}
+	if isIdent(ret[len(ret)-1].Type, "error") {
+		return true
+	}
 	// An error return parameter should be the last parameter.
 	// Flag any error parameters found before the last.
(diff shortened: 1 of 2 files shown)


mgechev:
@yangdiangzb thanks for the catch! Would you add the corresponding test changes as well?

rule/error-return.go
```diff
@@ -47,6 +47,9 @@ func (w lintErrorReturn) Visit(n ast.Node) ast.Visitor {
 	if len(ret) <= 1 {
 		return w
 	}
+	if isIdent(ret[len(ret)-1].Type, "error") {
+		return true
```
chavacava: This line does not compile.
Must be: `return nil`

yangdiangzb: Fixed.

yangdiangzb:
Test changes added.

PR #114, opened 2019-03-21
new rule: import shadowing

chavacava:
Closes #113


rule/import-shadowing.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (w importShadowing) Visit(n ast.Node) ast.Visitor {
+	switch n := n.(type) {
+	case *ast.AssignStmt:
+		if n.Tok == token.DEFINE {
+			return w // analyze variable declarations of the form id := expr
+		}
+
+		return nil
+	case *ast.CallExpr, *ast.ImportSpec, *ast.KeyValueExpr, *ast.ReturnStmt, *ast.SelectorExpr, *ast.StructType:
```
mgechev: This is for optimization to not traverse the subtree?

chavacava: Is to avoid failures on each reference to a shadowing identifier.

mgechev: Isn't this handled by the `alreadySeen` property?

chavacava: My previuos message was not clear. While `alreadySeen` avoids failing multiple times on a shadowing identifier, the case
```suggestion
	case *ast.CallExpr, *ast.ImportSpec, *ast.KeyValueExpr, *ast.ReturnStmt, *ast.SelectorExpr, *ast.StructType:
```
Avoids to analyze subtrees that can raise _false positives_; for example:

* skips analysis of the `import` sub-tree: we already have the list of imports
* skips analysis of key-val expressions (`{key:value}`): ids of such expressions, even the same of an import name, do not shadow the import name
* skips analysis of selector expressions (`anId.otherId`): because if `anId`shadows an import name, it was already detected, and `otherId` does not shadows the import name
* skips analysis of struct type because struct fields can not shadow an import name

skipping analysis of returns is an optimization but for the sake of clarity we can remove it from the case expression

mgechev: Thanks for the clarification! Would you add this as a comment in the rule? It'll be very useful to make sure we don't lose this information.

chavacava: done

PR #119, opened 2019-04-17
new rule: unused-receiver

### LINKED ISSUE #118: Warn on unused method receivers
In GO it is possible to define a method that does not use its receiver:

```go
func (ms *myStruct) Foo() { return }
```

The above method can be also written to explicitly state that the receiver is not used:

```go
func (_ *myStruct) Foo() { return }
```
or
```go
func (*myStruct) Foo() { return }
```

I propose to create a new rule (`unused-receiver`?) to cope with this pattern

chavacava:
Closes #118

### DIFF
--- a/fixtures/unused-receiver.go
+++ b/fixtures/unused-receiver.go
@@ -16,3 +16,28 @@ func (f *Unix) Format(failures <-chan lint.Failure, _ lint.RulesConfig) (string,
 	}
 	return "", nil
 }
+
+func (u *Unix) Foo() (string, error) { // MATCH /method receiver 'u' is not referenced in method's body, consider removing or renaming it as _/
+	for failure := range failures {
+		u := 1 // shadowing the receiver
+		fmt.Printf("%v\n", u)
+	}
+	return "", nil
+}
+
+func (u *Unix) Foos() (string, error) {
+	for failure := range failures {
+		u := 1 // shadowing the receiver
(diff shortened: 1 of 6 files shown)


fixtures/unused-receiver.go
```diff
@@ -0,0 +1,18 @@
+package fixtures
+
+import (
+	"fmt"
+
+	"github.com/mgechev/revive/lint"
+)
+
+func (f *Unix) Name() string { // MATCH /method receiver 'f' is not referenced in method's body, consider removing or renaming it as _/
```
mgechev: Can we also add a test case in which we have local variable named after the receiver and we use it?

Also one case when the receiver is used.

rule/unused-receiver.go
```diff
@@ -0,0 +1,77 @@
...
+			return nil // the receiver is anonymous: func (aType) Foo(...) ...
+		}
+
+		recID := rec.Names[0]
+		if recID.Name == "_" {
+			return nil // the receiver is already named _
+		}
+
+		// inspect the func body looking for references to the receiver id
+		fselect := func(n ast.Node) bool {
```
mgechev: We should handle local variables which are named after the receiver

chavacava: Using the [`Obj`](https://golang.org/pkg/go/ast/#Object) field of the receiver identifier to search references to it guarantees us no collision with other identifiers.

PR #121, opened 2019-04-20
simpler and more efficient implementation of unused-parameter

chavacava:
This PR proposes a simpler and more memory/time efficient implementation of the `unused-parameter` rule.

The new implementation replaces the ad-hoc handling of ids scopes by leveraging `Obj` field of `ast.Ident` nodes.

Benchmark comparison of the current (old) and the proposed (new) implementation:
```
benchdiff old.txt new.txt
benchmark                  old ns/op     new ns/op     delta
BenchmarkUnusedParam-4     1126516       1044633       -7.27%

benchmark                  old allocs     new allocs     delta
BenchmarkUnusedParam-4     7086           6434           -9.20%

benchmark                  old bytes     new bytes     delta
BenchmarkUnusedParam-4     295103        264038        -10.53%
```

:warning: There is a minor change in the semantics of the rule. While the current implementation considers a parameter _used_ if it is used (in the sense of data-flow analysis), now a parameter is considered used if it is _referenced_ somewhere in the function body (this includes LHS of assignments) In practice, this change has little impact because the case of parameters that are **only** redefined (assigned) in the function is very rare (and `revive already has a rule to cope with such cases). I've executed the current and the new version on kubernetes code base and the current rule identifies only one more failure than the new version (over ~1700)

### DIFF
--- a/rule/unused-param.go
+++ b/rule/unused-param.go
@@ -3,7 +3,6 @@ package rule
 import (
 	"fmt"
 	"go/ast"
-	"go/token"
 
 	"github.com/mgechev/revive/lint"
 )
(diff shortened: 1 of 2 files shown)


mgechev:
>now a parameter is considered used if it is referenced somewhere in the function body (this includes LHS of assignments)

This makes even more sense to me. If we reassign a reference, this shouldn't be considered as not unused.

That's a great catch @chavacava!

PR #124, opened 2019-04-22
unhandled-error (new rule)

### LINKED ISSUE #120: Warn on unhandled error.
Sometimes I forget to handle an error. Go linters usually don't warn me. At least those I know. I don't think I am the only one who is bothered by this. By implementing this everyone's code will greatly improve.
For example:

Bad code:
```go
os.Chdir("..")
```
But if we look into the docs, we can see that `os.Chdir` returns error.
Good code:
```go
err = os.Chdir("..")
if err != nil {
    // something
}
```
Actually, we don't need to check for the `if err != nil` part. It is handled by go compiler. We just need to check if err was assigned.

**Solution I would like**
Error like this:
`os.Chdir error not handled`

Thanks.

chavacava:
Closes #120

### DIFF
--- a/rule/unhandled-error.go
+++ b/rule/unhandled-error.go
@@ -1,0 +1,95 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"go/types"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// UnhandledErrorRule lints given else constructs.
+type UnhandledErrorRule struct{}
+
+// Apply applies the rule to given file.
+func (r *UnhandledErrorRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	walker := &lintUnhandledErrors{
+		pkg: file.Pkg,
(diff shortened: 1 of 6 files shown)


rule/unhandled-error.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		funcType := w.pkg.TypeOf(fCall)
+		if funcType == nil {
+			return nil // skip, type info not available
+		}
+
+		switch t := funcType.(type) {
+		case *types.Named:
+			if t.String() != errorTypeName {
```
mgechev: Wondering if we have a reference to the `error` type we can use?

chavacava: I'm not too happy with this way of checking for errors. I'll investigate more.

chavacava: I've did not find a better way of checking for `error` type other than using its string representation.
The new version of the checking use the fully-qualified type name (`_.error`) instead of the plain string representation (`error`)

I've encapsulated the checking inside a function, that way if in the future we find a better way, we just need to modify the function.

PR #126, opened 2019-04-25
proposal to add sklearn to "Who uses revive" in README

pa-m:
added line
- [`sklearn`](https://github.com/pa-m/sklearn) - A partial port of scikit-learn written in Go


pa-m:
Hi, test failure is related to travis authentication, not related to the change itself.

PR #130, opened 2019-05-03
too-many-indirections (new rule)

### LINKED ISSUE #129: Warn on presence of too many indirections
Expressions that have too many indirections are hard to understand.
For example (from Kubernetes):

```go
o.s.GA.ZoneOperations.Get(o.projectID, o.key.Zone, o.key.Name).Context(ctx).Do()
f.Signature.Receiver.Elem.Name
new(RouteBuilder).typeNameHandler(w.typeNameHandleFunc).servicePath(w.rootPath).Method("DELETE").Path(subPath)
```
Other than hard to read, this kind of code is fragile (calls and slice accesses are chained without checking for nil returns nor empty slices)

I propose to create a rule that warns when an expression has too many indirections (_too many_ must be configurable)

chavacava:
Closes #129

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -48,6 +48,7 @@ List of all available rules.
   - [struct-tag](#struct-tag)
   - [superfluous-else](#superfluous-else)
   - [time-naming](#time-naming)
+  - [too-many-indirections](#too-many-indirections)
   - [var-naming](#var-naming)
   - [var-declaration](#var-declaration)
   - [unexported-return](#unexported-return)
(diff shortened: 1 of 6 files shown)


RULES_DESCRIPTIONS.md
```diff
@@ -410,6 +411,19 @@ _Description_: Using unit-specific suffix like "Secs", "Mins", ... when naming v
 
 _Configuration_: N/A
 
+## too-many-indirections
```
mgechev: Thinking of it further, looks like this rule discourages chaining. I'm not particularly against this pattern.

I'd suggest to either rename the rule, to be more explicit what it enforces or change the confidence to something below 1.

What do you think?

chavacava:
I close this PR until I get more time to work on it

PR #193, opened 2019-07-27
Conf reason rule disabling

### LINKED ISSUE #190: Ability to provide a reason for rule disabling
I would like the ability to provide a reason for rule disabling (on the same line) as an option. Adding anything to the `revive:` comment breaks it (regex I guess?), so currently I must do it on a separate line.

I propose two things:

1. Ability to write comments like
`// revive:disable-line:RULE   This is intentional`
or even (for cleaner integration with other tools)
//  revive:disable:var-naming   TODO: fix it later

2. Global config flag (disabled by default) that would require to provide the reason (like `staticcheck` does by default)

chavacava:
Closes #190

### DIFF
--- a/lint/config.go
+++ b/lint/config.go
@@ -14,10 +14,11 @@ type RulesConfig = map[string]RuleConfig
 
 // Config defines the config of the linter.
 type Config struct {
-	IgnoreGeneratedHeader bool `toml:"ignoreGeneratedHeader"`
-	Confidence            float64
-	Severity              Severity
-	Rules                 RulesConfig `toml:"rule"`
-	ErrorCode             int         `toml:"errorCode"`
-	WarningCode           int         `toml:"warningCode"`
+	IgnoreGeneratedHeader  bool `toml:"ignoreGeneratedHeader"`
+	Confidence             float64
+	Severity               Severity
+	Rules                  RulesConfig `toml:"rule"`
+	ErrorCode              int         `toml:"errorCode"`
+	WarningCode            int         `toml:"warningCode"`
+	SpecifyDisableReason   bool `toml:"specifyDisableReason"`
 }
--- a/lint/config.go
+++ b/lint/config.go
@@ -12,13 +12,23 @@ type RuleConfig struct {
 // RulesConfig defines the config for all rules.
 type RulesConfig = map[string]RuleConfig
 
+// DirectiveConfig is type used for the linter directive configuration.
+type DirectiveConfig struct {
+	Severity  Severity
+}
+
+// DirectivesConfig defines the config for all directives.
+type DirectivesConfig = map[string]DirectiveConfig
+
+
 // Config defines the config of the linter.
 type Config struct {
-	IgnoreGeneratedHeader  bool `toml:"ignoreGeneratedHeader"`
+	IgnoreGeneratedHeader  bool             `toml:"ignoreGeneratedHeader"`
 	Confidence             float64
 	Severity               Severity
-	Rules                  RulesConfig `toml:"rule"`
-	ErrorCode              int         `toml:"errorCode"`
-	WarningCode            int         `toml:"warningCode"`
-	SpecifyDisableReason   bool `toml:"specifyDisableReason"`
+	Rules                  RulesConfig      `toml:"rule"`
+	ErrorCode              int              `toml:"errorCode"`
+	WarningCode            int              `toml:"warningCode"`
+	Directives             DirectivesConfig `toml:"directive"`
 }
+
--- a/lint/config.go
+++ b/lint/config.go
@@ -14,21 +14,19 @@ type RulesConfig = map[string]RuleConfig
 
 // DirectiveConfig is type used for the linter directive configuration.
 type DirectiveConfig struct {
-	Severity  Severity
+	Severity Severity
 }
 
 // DirectivesConfig defines the config for all directives.
 type DirectivesConfig = map[string]DirectiveConfig
 
-
 // Config defines the config of the linter.
 type Config struct {
-	IgnoreGeneratedHeader  bool             `toml:"ignoreGeneratedHeader"`
-	Confidence             float64
-	Severity               Severity
-	Rules                  RulesConfig      `toml:"rule"`
-	ErrorCode              int              `toml:"errorCode"`
-	WarningCode            int              `toml:"warningCode"`
-	Directives             DirectivesConfig `toml:"directive"`
+	IgnoreGeneratedHeader bool `toml:"ignoreGeneratedHeader"`
+	Confidence            float64
+	Severity              Severity
+	Rules                 RulesConfig      `toml:"rule"`
+	ErrorCode             int              `toml:"errorCode"`
+	WarningCode           int              `toml:"warningCode"`
+	Directives            DirectivesConfig `toml:"directive"`
 }
-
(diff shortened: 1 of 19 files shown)


chavacava:
@mgechev I'm not sure if exiting with error from file's `disabledIntervals` is the best idea...

lint/file.go
```diff
@@ -202,24 +209,28 @@ func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
...
+			for _, name := range tempNames {
+				name = strings.Trim(name, "\n")
+				if len(name) > 0 {
+					ruleNames = append(ruleNames, name)
 				}
 			}
 
+			mustCheckDisablingReason := specifyDisableReason && match[directivePos] == "enable"
+			if  mustCheckDisablingReason && strings.Trim(match[reasonPos]," ") == "" {
+				fmt.Fprintln(os.Stderr, fmt.Sprintf("%s:%d: reason of lint disabling not found.",filename,line))
```
mgechev:
```suggest
				fmt.Fprintln(os.Stderr, fmt.Sprintf("%s:%d: reason of lint disabling not found.", filename,line))
```

lint/file.go
```diff
@@ -202,24 +209,28 @@ func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
...
-					}
+			tempNames := strings.Split(match[rulesPos], ",")
+			for _, name := range tempNames {
+				name = strings.Trim(name, "\n")
+				if len(name) > 0 {
+					ruleNames = append(ruleNames, name)
 				}
 			}
 
+			mustCheckDisablingReason := specifyDisableReason && match[directivePos] == "enable"
```
mgechev: Isn't this `disable`, or I'm missing something?

lint/file.go
```diff
@@ -126,8 +128,13 @@ type enableDisableConfig struct {
...
 
-func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
-	re := regexp.MustCompile(`^//[\s]*revive:(enable|disable)(?:-(line|next-line))?(?::([^\s]+))?[\s]*$`)
+const directiveRE = `^//[\s]*revive:(enable|disable)(?:-(line|next-line))?(?::([^\s]+))?[\s]*(?: (.+))?$`
+const directivePos = 1
+const modifierPos = 2
+const rulesPos = 3
+const reasonPos = 4
+func (f *File) disabledIntervals(rules []Rule, specifyDisableReason bool) disabledIntervalsMap {
+	re := regexp.MustCompile(directiveRE)
```
mgechev: Maybe we can also compile outside of the function since we're not changing the regex?

lint/file.go
```diff
@@ -126,8 +128,13 @@ type enableDisableConfig struct {
 	position int
 }
 
-func (f *File) disabledIntervals(rules []Rule) disabledIntervalsMap {
```
mgechev: Good point to move it outside of `disabledIntervals`.

lint/file.go
chavacava: ooops. Fixed

lint/file.go
chavacava: Done

jBugman:
`staticcheck` just ignores disable comments without provided reason as if they aren't there. While it is a perfectly valid behaviour, I think it would be better to treat them as lint errors (just like staticcheck does for not applicable comments)

chavacava:
My understanding is that `staticcheck` does not recognize directives without a reason (the directive regexp does not match)
@mgechev What do you think of that approach?
I can modify the code to add a disabling range if the flag `specifyDisableReason` is not set or the directive has a reason.

mgechev:
> I think it would be better to treat them as lint errors

This makes sense, but it opens few questions:
- What should be the severity of this
- What should be the name of the rule

I'd vote for keeping the error that @chavacava implemented in the PR.

jBugman:
Unfortunately exiting with error silently breaks the tooling, i.e. vscode integration (just tested it), no errors or warnings are reported at all.

chavacava:
Hi, I've pushed a commit that implements my understanding of @jBugman's requirement wrt "treat them as lint errors" 

If `specify-disable-reason` is required, when `revive` founds linter directives lacking disable reason it will generate a failure (as with code not respecting a rule). 
Activation and severity of this check is done by configuration:
```toml
[directive.specify-disable-reason]
    severity = "error"
```

Notes:
1. I think the implementation can be better, this commit is just for check if it is the desired behaviour
1. I didn't find something better than `directive`.

lint/file.go
```diff
@@ -99,9 +97,12 @@ func (f *File) isMain() bool {
 	return false
 }
 
+const directiveSpecifyDisableReason = "specify-disable-reason"
+
 func (f *File) lint(rules []Rule, config Config, failures chan Failure) {
 	rulesConfig := config.Rules
-	disabledIntervals := f.disabledIntervals(rules, config.SpecifyDisableReason)
+	_, mustSpecifyDisableReason := config.Directives[directiveSpecifyDisableReason]
```
mgechev: Do we handle the case when `directiveSpecifyDisableReason` is not specified in the config?

lint/config.go
```diff
@@ -12,13 +12,23 @@ type RuleConfig struct {
...
 	Confidence             float64
 	Severity               Severity
-	Rules                  RulesConfig `toml:"rule"`
-	ErrorCode              int         `toml:"errorCode"`
-	WarningCode            int         `toml:"warningCode"`
-	SpecifyDisableReason   bool `toml:"specifyDisableReason"`
+	Rules                  RulesConfig      `toml:"rule"`
+	ErrorCode              int              `toml:"errorCode"`
+	WarningCode            int              `toml:"warningCode"`
+	Directives             DirectivesConfig `toml:"directive"`
```
mgechev: How can we disable it?

lint/file.go
chavacava: Yes, after the execution of this line, if `directiveSpecifyDisableReason` is not in the config, the variable `mustSpecifyDisableReason` is `false`

lint/config.go
chavacava: If `[directive.specify-disable-reason]` is not present in the config file ==> `Directives` is empty ==> it is disabled (see my previous comment above)

chavacava:
@mgechev if the setting is not specified in the config file `revive` will behave as usual.

If you are OK with this implementation, please **do not** accept it yet, I'll continue to work on it tomorrow. Tnx

mgechev:
@chavacava LGTM! Looking forward for the final implementation.

@jBugman, let us know if you notice anything here.

chavacava:
@mgechev I've pushed a modification to the README to document this new feature, please check if OK to merge

mgechev:
LGTM. @jBugman would you give it a try?

jBugman:
My results:
- [x] Comments with and without reason work perfectly
- [x] directive.specify-disable-reason creates new `specify-disable-reason  reason of lint disabling not found` warning and turns off corresponding comment, triggering underlying lint warning too
- [ ] `severity = "error"` does not seem to work with root revive severity set to either `warning` or `error` (it stays at warning)

chavacava:
Nice catch @jBugman 
Please check the last commit

jBugman:
It works now, thank you for your work!

PR #198, opened 2019-07-29
fix: adhere to "Rule of Silence"

### LINKED ISSUE #165: Revive violates the Rule of Silence
When revive finds no errors it prints:

> 0 problems (0 errors) (0 warnings)

This violates the Rule of Silence:

> The rule of silence, also referred to as the silence is golden rule, is an important part of the Unix philosophy that states that when a program has nothing surprising, interesting or useful to say, it should say nothing. It means that well-behaved programs should treat their users' attention and concentration as being valuable and thus perform their tasks as unobtrusively as possible. That is, silence in itself is a virtue. (http://www.linfo.org/rule_of_silence.html)

What would you say to a PR that made revive silent by default?

markelog:
Makes stylish formatter respect "Rule of Silence"
See http://www.linfo.org/rule_of_silence.html

Fixes #165

### DIFF
--- a/formatter/stylish.go
+++ b/formatter/stylish.go
@@ -82,7 +82,8 @@ func (f *Stylish) Format(failures <-chan lint.Failure, config lint.RulesConfig) 
 	} else if total > 0 && totalErrors == 0 {
 		suffix = color.YellowString("\n ✖" + suffix)
 	} else {
-		suffix = color.GreenString("\n" + suffix)
+		suffix, output = "", ""
 	}
+
 	return output + suffix, nil
 }


markelog:
As it turns out `friendly` already does this, so only `stylish` is changed

chavacava:
Hi @markelog, thanks for the PR.
Reading the code of `checkstyle` formatter, I think it will generate an XML result even in case of no errors were found. But, from my understanding of the _rule of silence_, it is not a problem because an XML output is not for human users. I'm OK?

markelog:
> But, from my understanding of the rule of silence, it is not a problem because an XML output is not for human users

Yes, my understanding is the same and the description of it also mentiones stuff like 

> The rule of silence does not mean that any less information is available. Rather, by default it just does not provide information that is likely to be unnecessary in most situations. However, it allows the user to use options, also referred to as switches, to view detailed information if desired. Many command line programs have a verbose option, and it is typically -v.

So I believe `json`/`checkstyle` formatters could be considered as these types of `switches`. Back when I was maintaining [jscs](https://github.com/jscs-dev/node-jscs), we also followed same approach.

So we should be good here :)

PR #225, opened 2019-08-18
Update go.mod and go.sum

mgechev:
@Clivern would you double check if everything looks fine?


Clivern:
Looks perfect, Thanks for the fix!

PR #231, opened 2019-09-10
fixes issue #229

### LINKED ISSUE #229: False positives for "exported" rule
**Describe the bug**
revive reports "comment on exported type T should be of the form "T ..." (with optional leading article)" when the comment on that type is of the form "/* T ... */" (with a leading space).  Likewise for exported functions, and presumably ValueSpec entries as well.

**To Reproduce**
1. I updated revive `go get -u github.com/mgechev/revive` (commit fa5acbc1f03, from 31 Aug)
2. I run it with the following flags, configuration file, and Go source file:

```shell
revive -config exported.toml exported.go
```

```toml
[rule.exported]
```

```golang
package exported

/* Exported is an exported structure. */
type Exported struct {
	p int
}
```

**Expected behavior**
I expected no warning.  If I change the comment style to //, the warning does not occur.

 - OS: Ubuntu 19.04
 - Tested with go1.10.4 and go1.13.

**Additional context**
It is easy to fix with strings.TrimSpace() in the obvious spots in exported.go, but I could not figure out how to add a regression test, so I did not file this as a PR.  I would be happy to do that, but the problem is small enough that it may be easier to do it than to walk me through adding a suitable regression test.

chavacava:
Closes #229

### DIFF
--- a/fixtures/golint/exported.go
+++ b/fixtures/golint/exported.go
@@ -13,3 +13,23 @@ type (
 )
 
 type Foo struct{} // MATCH /exported type Foo should have comment or be unexported/
+
+// The following cases are no-regression tests for issue 229
+
+/* Bar something */
+type Bar struct{}
+
+/* Toto something */
+func Toto() {}
+
+/* FirstLetter something */
+const FirstLetter = "A"
+
+/*Bar2 something */
+type Bar2 struct{}
+
+/*Toto2 something */
+func Toto2() {}
+
+/*SecondLetter something */
+const SecondLetter = "B"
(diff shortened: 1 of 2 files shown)


fixtures/golint/exported.go
```diff
@@ -13,3 +13,23 @@ type (
...
+/* Toto something */
+func Toto() {}
+
+/* FirstLetter something */
+const FirstLetter = "A"
+
+/*Bar2 something */
+type Bar2 struct{}
+
+/*Toto2 something */
```
mgechev: This should match as well, right?

fixtures/golint/exported.go
```diff
@@ -13,3 +13,23 @@ type (
...
+func Toto2() {}
+
+/*SecondLetter something */
```
mgechev: And here.

Issue #232, opened 2019-09-16
Support Global Configuration


johnrichardrinehart:
It would be nice to support a global (possibly default) configuration for `revive`.

johnrichardrinehart:
Closed with #233

PR #233, opened 2019-09-16
support global config (e.g. for company-wide settings)

johnrichardrinehart:
Addresses #232.


mgechev:
Would you also document it in the README? Otherwise, LGTM!

@chavacava, would you take a look as well?

config.go
```diff
@@ -233,12 +235,22 @@ func init() {
...
+	// command line help strings
 	const (
-		configUsage    = "path to the configuration TOML file (i.e. -config myconf.toml)"
+		configUsage    = "path to the configuration TOML file, defaults to $HOME/revive.toml, if present (i.e. -config myconf.toml)"
 		excludeUsage   = "list of globs which specify files to be excluded (i.e. -exclude foo/...)"
 		formatterUsage = "formatter to be used for the output (i.e. -formatter stylish)"
 	)
-	flag.StringVar(&configPath, "config", "", configUsage)
+
+	var globalConfigPath string
```
chavacava: I propose to rename `globalConfigPath` into `defaultConfigPath`.

To simplify the reading of the `init()` function we can also factor out the calculation of the default path into a helper function like

```go
func buildDefaultConfigPath() string {
	var result string
	if homeDir, err := homedir.Dir(); err == nil {
		result = filepath.Join(homeDir, "revive.toml")
		if _, err := os.Stat(result); err != nil {
			result = ""
		}
	}

	return result
}
```

And just write

```go
defaultConfigPath := buildDefaultConfigPath()
```

johnrichardrinehart: I had similar sentiments. But, I didn't have a good idea about where to put the helper function.

Let me know what you think.

@chavacava Your changes have been implemented (logic moved to helper function).
@mgechev I have added documentation to the `README`. Thanks for reminding me.

johnrichardrinehart:
@mgechev I'm not sure how long you want to wait on @chavacava but I vote for closing #232 and merging #233.

chavacava:
LGTM
(CI not working)

PR #282, opened 2019-12-20
fixes issue #280

### LINKED ISSUE #280: Bare return detected in an anynomous function inside a function only if the latter function’s return value is named
**Describe the bug**
With `[rule.bare-return]` on, Revive emits an error message in the case detailed below, and not in another case, also detailed below, that I think should be equivalent.

**To Reproduce**

```shell
# flags

revive ...
```

```toml
[rule.bare-return]
```

**Expected behavior**
No output.

**Logs**
`main.go:5:3: avoid using bare returns, please add return expressions`

 - OS: Linux Mint 19.2 (based on Ubuntu 18.04)
 - Version of Go: 1.13.4

**Additional context**

Here is the minimum example that triggers the Revive error:
```go
package main

func f() (err error) {
	func() {
		return
	}()

	return nil
}
```

No errors are produced if the return value is not named:
```go
package main

func f() error {
	func() {
		return
	}()

	return nil
}
```

I think there should be no difference between the two cases, since the incriminated `return` is in an anonymous function that is identical in both cases.

chavacava:
Closes #280

### DIFF
--- a/fixtures/bare-return.go
+++ b/fixtures/bare-return.go
@@ -31,3 +31,38 @@ func bare4(a string, b int) string {
 func bare5(a string, b int) {
 	return
 }
+
+// NR tests for issue #280
+func f280_1() (err error) {
+	func() {
+		return
+	}()
+
+	return nil
+}
+
+func f280_2() (err error) {
+	func() (r int) {
+		return // MATCH /avoid using bare returns, please add return expressions/
+	}()
+
+	return nil
+}
+
+func f280_3() (err error) {
+	func() (r int) {
+		return 1
+	}()
+
+	return // MATCH /avoid using bare returns, please add return expressions/
+}
+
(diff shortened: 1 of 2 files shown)


fixtures/bare-return.go
```diff
@@ -31,3 +31,38 @@ func bare4(a string, b int) string {
 func bare5(a string, b int) {
 	return
 }
+
+// NR tests for issue #280
+func f280_1() (err error) {
+	func() {
+		return
```
mgechev: I'm missing something here. Why we don't report an error?

chavacava: A _bare return_ is a `return` that makes _implicit_ reference to the _named_ return values of the function 
The scope of the `return` at line 38 is the function literal. That func literal does not have return values thus the `return` it is not a bare return

PR #285, opened 2019-12-25
t.Error doesn't return a value

quasilyte:
Go compiler would complain about using void-result
function value as a return expression.


mgechev:
Thanks for the PR!

PR #342, opened 2020-02-14
String of int

### LINKED ISSUE #336: New rule: warn when converting int into string with string(int)
`go vet` added a rule to warn on [string(int) conversion](https://github.com/golang/go/issues/32479), I think we could add the same rule in `revive`

chavacava:
Closes #336

### DIFF
--- a/rule/string-of-int.go
+++ b/rule/string-of-int.go
@@ -1,0 +1,95 @@
+package rule
+
+import (
+	"go/ast"
+	"go/types"
+
+	"github.com/mgechev/revive/lint"
+)
+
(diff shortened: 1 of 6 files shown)


rule/string-of-int.go
```diff
@@ -0,0 +1,95 @@
...
+	if !w.isCallStringCast(ce.Fun) {
+		return w
+	}
+
+	if !w.isIntExpression(ce.Args) {
+		return w
+	}
+
+	w.onFailure(lint.Failure{
+		Confidence: 1,
```
mgechev: Maybe we can drop the confidence here? There's still a chance they want to get the codepoint.

PR #353, opened 2020-02-25
Add rule 'range-val-address'

### LINKED ISSUE #352: New Rule: range-val-address
Hallo.

Taking the address of the "loop variable" when ranging over a slice is a sneaky mistake, a developer can make. e.g.: https://play.golang.org/p/d6h81KqZiCa

Lint assignments in loops, where the name on the rhs matches the "loop variable" and is prefixed by &

**Additional context**
i have already created the code for the rule (will follow in a pull request shortly), and for the given test cases, it does work. But i have very little knowledge about the golang ast, so please review it in detail. (especially performance-wise)

bernhardreisenberger:
See issue #352 for details

this rule will add the functionality to lint unwanted address assignments in range-loops.
i added a few test cases and these tests are passing. 
as i stated in issue #352, i have no experience with golang ast, so please review in detail.

Closes #352

### DIFF
--- a/rule/range-val-address.go
+++ b/rule/range-val-address.go
@@ -1,6 +1,7 @@
 package rule
 
 import (
+	"fmt"
 	"go/ast"
 	"go/token"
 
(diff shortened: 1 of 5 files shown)


chavacava:
Hi @bernhardreisenberger thanks for your contribution. I'll try to review it during this week, meanwhile you can test the rule against big/complex repositories like kubernetes

bernhardreisenberger:
i tested the rule against kubernetes, and found a few occurrences. from my point of view, these warnings are all valid. (checked all in /cmd (3), /plugin (11), and some in /staging (42))
e.g.
- cmd/phases/init/certs_test.go:66:21
- cmd/kubeadm/app/cmd/upgrade/diff.go:139:23
- cmd/kubeadm/app/cmd/token.go:391:17

i will apply your change requests today.

rule/range-val-address.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	for _, exp := range asgmt.Rhs {
+		switch e := exp.(type) {
+		case *ast.UnaryExpr: // e.g. ...&value
+			if bw.isAccessingRangeValueAddress(e) {
+				bw.onFailure(bw.newFailure(e))
+			}
+		case *ast.CallExpr:
+			if fun, ok := e.Fun.(*ast.Ident); ok && fun.Name == "append" { // e.g. ...append(arr, &value)
```
bernhardreisenberger: limited to 'append' functions to reduce false positives. 
another option would be to reduce the confidence for non 'append' functions.

chavacava: I was wandering, why limiting to `append` and assignments? Is using the address of the range value variable not always suspicious?

bernhardreisenberger: i summarized the problem with an example: https://play.golang.org/p/uAJAgz7apQv 

if the dev is using the address of the value after the loop (by adding it to a list, map or variable), there will be problems.
if the dev is using the address of the value right there in the loop (not storing it for later processing, e.g. adding it to a list, map, variable in the outer scope, everything is fine)

chavacava:
@bernhardreisenberger everything looks good to me, please update README.md with the new rule. Thanks!

mgechev:
I merged this one already. A new PR for the readme update works as well.

PR #406, opened 2020-05-04
Late return rule

### LINKED ISSUE #404: new rule: late-return
This new rule is inspired by [this question on twitter](https://twitter.com/francesc/status/1255620804705808384): 

> Is there a static analyzer for #golang that will warn about possible early returns?
```go
if foo {
 .... 
 return a
} else {
  return b
}
```
Should be:
```go
if ! foo {
  return b
}
...
return a
```
Contrary to other `revive` rules this is more a style-related rule. I'm not sure if it has its place in `revive`

chavacava:
Closes #404


README.md
```diff
@@ -349,7 +349,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | [`unhandled-error`](./RULES_DESCRIPTIONS.md#unhandled-error)   | []string   | Warns on unhandled errors returned by funcion calls    |    no    |  yes   |
 | [`cognitive-complexity`](./RULES_DESCRIPTIONS.md#cognitive-complexity)          |  int   | Sets restriction for maximum Cognitive complexity.              |    no    |  no   |
 | [`string-of-int`](./RULES_DESCRIPTIONS.md#string-of-int)          |  n/a   | Warns on suspicious casts from int to string            |    no    |  yes   |
-
+| [`late-return`](./RULES_DESCRIPTIONS.md#late-return)          |  n/a   | Spots if-then-else statements that can be refactored to simplify code reading            |    no    |  no   |
```
mgechev, line 352: I've heard referring to this as early return more often, although I understand why you named it `late-return`, that's what we're aiming against.

rule/late-return.go
```diff
@@ -0,0 +1,79 @@
...
+			return w
+		}
+
+		lenThenBlock := len(n.Body.List)
+		if lenThenBlock < 1 {
+			// then block is empty thus the stmt can be rewritten
+			w.onFailure(lint.Failure{
+				Confidence: 1,
+				Node:       n,
+				Failure:    "if c { } else {... return} can be rewritten as if !c {... return }",
```
mgechev, line 60: I'd change it from "rewritten" to "simplified".

rule/late-return.go
```diff
@@ -0,0 +1,79 @@
...
+			})
+
+			return w
+		}
+
+		elseIsShorterThanThen := lenElseBlock < lenThenBlock
+		_, lastThenStmtIsReturn := n.Body.List[lenThenBlock-1].(*ast.ReturnStmt)
+		_, lastElseStmtIsReturn := elseBlock.List[lenElseBlock-1].(*ast.ReturnStmt)
+		if lastElseStmtIsReturn && (elseIsShorterThanThen || !lastThenStmtIsReturn) {
```
mgechev, line 69: I'm not sure I understand this condition `lastElseStmtIsReturn && (elseIsShorterThanThen || !lastThenStmtIsReturn)`. How's the number of statements in the block related to the outcome?

chavacava: You are right, it is a code from an experiment ("spot `then` blocks larger than `else` ones") I've forgot to delete.

PR #411, opened 2020-05-08
Identical branches

### LINKED ISSUE #410: new rule: identical-branches
An `if-then-else` conditional with identical implementations in both branches is an error.
I've developed a prototype of a rule that detects such cases (testing the rule against real code has [already found bugs](https://github.com/gogo/protobuf/issues/687))

chavacava:
Closes #410

### DIFF
--- a/rule/identical-branches.go
+++ b/rule/identical-branches.go
@@ -1,0 +1,98 @@
+package rule
+
+import (
+	"go/ast"
+	"reflect"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// IdenticalBranchesRule warns on constant logical expressions.
+type IdenticalBranchesRule struct{}
+
+// Apply applies the rule to given file.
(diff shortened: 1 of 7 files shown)


rule/identical-branches.go
```diff
@@ -0,0 +1,82 @@
...
+		if gofmt(branches[i]) != ref {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (w lintIdenticalBranches) newFailure(node ast.Node, msg string) {
+	w.onFailure(lint.Failure{
```
mgechev, line 76: Do you think there's a case when we'll return a false positive? If we have a variable declaration with the same name, but different types, for example, are we going to catch this?

Since you're using `gofmt`, I don't think we can hit such a case, but just pointing this out.

chavacava: The rule uses (inefficient) string comparison to find identical branches so false positives are not possible.
In the near future I'll try to provide a, I hope, more efficient implementation based on AST comparison.

PR #413, opened 2020-05-09
Unconditional recursion

### LINKED ISSUE #412: new rule: unconditional-recursion
Unconditional recursive calls will produce infinite recursion, thus program stack overflow.
Even if it seems that the frequency of such errors is low, autocompletion functionality of IDEs facilitates them. A typical case is:

```go
type myType struct {
  field int
}

func (mt myType) Field() int {
  return mt.Field()          // /!\ recursive call in place of accessing mt.field
} 
```

chavacava:
Closes #412

### DIFF
--- a/rule/unconditional-recursion.go
+++ b/rule/unconditional-recursion.go
@@ -1,0 +1,143 @@
+package rule
+
+import (
+	"go/ast"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// UnconditionalRecursionRule lints given else constructs.
+type UnconditionalRecursionRule struct{}
+
+// Apply applies the rule to given file.
+func (r *UnconditionalRecursionRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	onFailure := func(failure lint.Failure) {
+		failures = append(failures, failure)
+	}
+
+	// err := file.Pkg.TypeCheck()
+	// if err != nil {
+	// 	panic(fmt.Sprintf("Error while type-checking file %s: %v", file.Name, err))
+	// }
+	w := lintUnconditionalRecursionRule{onFailure: onFailure}
+	ast.Walk(w, file.AST)
+	return failures
+}
+
+// Name returns the rule name.
+func (r *UnconditionalRecursionRule) Name() string {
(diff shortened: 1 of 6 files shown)


rule/unconditional-recursion.go
```diff
@@ -0,0 +1,187 @@
...
+
+// Apply applies the rule to given file.
+func (r *UnconditionalRecursionRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	onFailure := func(failure lint.Failure) {
+		failures = append(failures, failure)
+	}
+
+	// err := file.Pkg.TypeCheck()
```
mgechev, line 20: We can remove the type checking here?

rule/unconditional-recursion.go
```diff
@@ -0,0 +1,187 @@
...
+		default:
+			return w
+		}
+
+		if w.currentFunc != nil && // not in a func body
+			!w.currentFunc.seenConditionalExit && // there is a conditional exit in the function
+			w.currentFunc.funcDesc.equal(&funcDesc{selector, funcId}) {
+			w.onFailure(lint.Failure{
+				Category:   "logic",
+				Confidence: 1,
```
mgechev, line 98: Maybe we can decrease the confidence a bit? A construct with a for loop when we're draining a channel may still throw?

rule/unconditional-recursion.go
chavacava: sure, type-checking is not necessary for this rule

rule/unconditional-recursion.go
chavacava: The rule is all about control dependencies.
`for-range`'s bodies are analysed looking for stmt taking the control out of the function (`return`, `os.Exit`, ...) to check if subsequent statements depend upon the loop
A channel-draining loop is not different from other loop.

About the confidence: I do not see cases where we can generate a false positive (do you have examples in mind). Contrary we can pass silently recursive calls, for example:
```go
func foo() {
  for false { // or other arbitrary complex condition evaluating to false
    return
  }
  foo()
``` 
will not trigger a failure.

mgechev:
@chavacava there are some conflicts with `master`. We can merge after that.

PR #430, opened 2020-05-25
Defer rule

### LINKED ISSUE #429: rule warning on defer gotchas
It could be useful to have a rule that warns on [`defer` gotchas](https://blog.learngoprogramming.com/5-gotchas-of-defer-in-go-golang-part-iii-36a1ab3d6ef1)

chavacava:
Closes #429

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -19,6 +19,7 @@ List of all available rules.
   - [context-keys-type](#context-keys-type)
   - [cyclomatic](#cyclomatic)
   - [deep-exit](#deep-exit)
+  - [defer](#defer)
   - [dot-imports](#dot-imports)
   - [duplicated-imports](#duplicated-imports)
   - [early-return](#early-return)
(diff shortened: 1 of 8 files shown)


mgechev:
I'll look at the PR next week.

RULES_DESCRIPTIONS.md
```diff
@@ -192,6 +193,25 @@ _Description_: Packages exposing functions that can stop program execution by ex
 
 _Configuration_: N/A
 
+## defer
+
+_Description_: This rule warns on some common mistakes when using `defer` statement. It currently alerts on the following situations:
```
mgechev, line 198: To make them configurable and allow folks to disable some of these checks, maybe it makes sense to have 4 separate rules?

rule/defer.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		switch fn := n.Call.Fun.(type) {
+		case *ast.CallExpr:
+			w.newFailure("prefer not to defer chains of function calls", fn, 1.0, "bad practice", "call-chain")
+		case *ast.SelectorExpr:
+			if id, ok := fn.X.(*ast.Ident); ok {
+				isMethodCall := id != nil && id.Obj != nil && id.Obj.Kind == ast.Typ
+				if isMethodCall {
+					w.newFailure("be careful when deferring calls to methods without pointer receiver", fn, 0.8, "bad practice", "method-call")
```
mgechev, line 108: In what cases can we return false positive?

chavacava: The rule only knows that it is a call to a method but it has no info about the method receiver (pointer or not)
If the receiver is a pointer receiver then deferring is safe

Example:

```go
func (mt myType) foo() {...}
func (mt *myType) bar() {...}
...
{
...
defer foo() // revive alerts 
defer bar() // revive alerts but deferring it is without risk because bar has a pointer receiver
...
}
```

chavacava:
@mgechev About splitting this rule in many rules: it is something I've thought. It seems you deleted your comment about the subject.

mgechev:
>@mgechev About splitting this rule in many rules: it is something I've thought. It seems you deleted your comment about the subject.

Yes, originally I didn't see the rule is configurable. What I was thinking is maybe change the configuration format a bit:

```toml
[["call-chain","loop"]]
```

This can mean that `call-chain` and `loop` are enabled. We can have everything enabled by default and folks can change the allowlist. Allowlist seems more intuitive to me.

PR #443, opened 2020-07-16
new rule: unexported-naming

### LINKED ISSUE #442: Warn on exported symbols within private scopes
In GO a symbol is exported (outside a package) if its name begins with a capital letter. 
Declaring an exported symbol within a private scope is useless, the symbol is not visible outside its scope ([example](https://play.golang.org/p/CWKw5KWKj3X)) Even if these kind of declarations are not a bug, they might difficult program understanding.

`revive` could warn on such cases

chavacava:
Closes #442

### DIFF
--- a/rule/unexported-naming.go
+++ b/rule/unexported-naming.go
@@ -1,0 +1,115 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// UnexportedNamingRule lints wrongly named unexported symbols.
+type UnexportedNamingRule struct{}
+
+// Apply applies the rule to given file.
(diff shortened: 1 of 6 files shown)


rule/unexported-naming.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+func (unl unexportablenamingLinter) lintIDs(ids []*ast.Ident) {
+	for _, id := range ids {
+		if id.IsExported() {
+			unl.onFailure(lint.Failure{
+				Node:       id,
+				Confidence: 1,
+				Category:   "naming",
+				Failure:    fmt.Sprintf("the symbol %s can not be exported, its name should start with a lowercase letter", id.String()),
```
mgechev, line 111: This might be a little confusing. My guess is that folks would not want to export a symbol if they declare it in a function's body. I'd suggest to focus on not starting names of local identifiers with a capital letter.

chavacava: Done

PR #449, opened 2020-08-02
raster optimizations

mathieu-aubin:
images sizes reduction in assets (documentation-related).

i understand it's not essential but to me, it feels considerable enough of savings to justify it. Your call! Thank you


mgechev:
It's definitely a great contribution. All images are more than 40% smaller now.

Issue #500, opened 2021-02-26
Don't consider blank import of "embed" (Go 1.16) an error


grongor:
I don't think that it should be necessary to comment on a blank import of "embed" everywhere, as the "why" is obvious.

Of course, it only makes sense to ignore this import if the embed is actually used in the file.

PR #501, opened 2021-02-26
Ignore blank import of embed if embed is actually used in the file

grongor:
Closes https://github.com/mgechev/revive/issues/500

### DIFF
--- a/rule/blank-imports.go
+++ b/rule/blank-imports.go
@@ -10,55 +10,38 @@ import (
 // BlankImportsRule lints given else constructs.
 type BlankImportsRule struct{}
 
-// Apply applies the rule to given file.
-func (r *BlankImportsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
-	var failures []lint.Failure
-
-	fileAst := file.AST
-	walker := lintBlankImports{
-		file:    file,
-		fileAst: fileAst,
-		onFailure: func(failure lint.Failure) {
-			failures = append(failures, failure)
-		},
-	}
-
-	ast.Walk(walker, fileAst)
-
-	return failures
-}
-
 // Name returns the rule name.
 func (r *BlankImportsRule) Name() string {
 	return "blank-imports"
 }
 
-type lintBlankImports struct {
-	fileAst   *ast.File
-	file      *lint.File
-	onFailure func(lint.Failure)
-}
-
-func (w lintBlankImports) Visit(_ ast.Node) ast.Visitor {
-	// In package main and in tests, we don't complain about blank imports.
-	if w.file.Pkg.IsMain() || w.file.IsTest() {
+// Apply applies the rule to given file.
+func (r *BlankImportsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	if file.Pkg.IsMain() || file.IsTest() {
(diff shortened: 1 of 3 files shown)


rule/blank-imports.go
```diff
@@ -52,14 +55,20 @@ func (w lintBlankImports) Visit(_ ast.Node) ast.Visitor {
...
 			prev := w.fileAst.Imports[i-1]
 			prevPos := w.file.ToPosition(prev.Pos())
-			if isBlank(prev.Name) && prevPos.Line+1 == pos.Line {
+
+			if isBlank(prev.Name) && prevPos.Line+1 == pos.Line && prev.Path.Value != embedImportPath {
 				continue // A subsequent blank in a group.
 			}
 		}
 
+		if imp.Path.Value == embedImportPath && w.fileHasValidEmbedComment() {
```
chavacava, line 68: we do not need to call `w.fileHasValidEmbedComment()` inside the loop. In fact we can call it just once at the beginning of the analysis. 
If we do this, and the file effectively has a valid embed comment, do we need to check the imports?

rule/blank-imports.go
```diff
@@ -52,14 +55,20 @@ func (w lintBlankImports) Visit(_ ast.Node) ast.Visitor {
 		if !isBlank(imp.Name) {
 			continue // Ignore non-blank imports.
 		}
+
 		if i > 0 {
 			prev := w.fileAst.Imports[i-1]
 			prevPos := w.file.ToPosition(prev.Pos())
-			if isBlank(prev.Name) && prevPos.Line+1 == pos.Line {
+
+			if isBlank(prev.Name) && prevPos.Line+1 == pos.Line && prev.Path.Value != embedImportPath {
```
chavacava, line 63: we could add a variable to document the semantics of this condition. Something like:
 ```suggestion
			isSubsequentBlancInAGroup := isBlank(prev.Name) && prevPos.Line+1 == pos.Line && prev.Path.Value != embedImportPath 
			if  isSubsequentBlancInAGroup {
```

chavacava (changes requested):
Hi @grongor, thanks for the PR. I've made a couple of remarks.

rule/blank-imports.go
grongor: Sure, why not :)

rule/blank-imports.go
grongor: Well, the embed import should be present in the file only once, so this method is effectively called only once. And for files that do not use embed (which will be majority of files) we don't do unnecessary comment analysis. From performance point of view I think my approach is better. But this is my first time working with Go AST so I might be wrong...

chavacava: @grongor you ar right, my bad. The rule logic is not based on AST traversal and can be directly implemented in the `Apply` method without requiring to use a tree walker object.

grongor: I think it's out of the scope for this MR/issue, but sure, it makes sense to not use the `ast.Walk` here ... I added the commit that fixes that.

chavacava (approved):

chavacava:
@grongor thanks for the PR and sorry for the late response (it was a busy week)

grongor:
No worries, thanks for merging :)

PR #502, opened 2021-03-10
Add information about GoLand support on README

s0xzwasd:


chavacava:
@s0xzwasd thanks for the contribution

Issue #489, opened 2020-12-17
Revive docker image


inhuman:
**Is your feature request related to a problem? Please describe.**
No docker image for CI

**Describe the solution you'd like**
Create please docker image for use in CI

doniacld:
Hey
I would be happy to work on this issue.

doniacld:
Hey @inhuman 
When revive will be released, you will be able to pull the docker image `ghcr.io/mgechev/revive:<tag-version>`. It will be found under the assets section.

If you really need it in the docker hub. I'll let @mgechev or @chavacava creates a repository in [dockerhub](https://docs.docker.com/docker-hub/repos/) under their name for revive if they want to. Then, I could update the goreleaser file to push also on the dockerhub when revive is released.

Also, I am curious why you do not use the binary and you need a docker image ? :)

PR #503, opened 2021-03-16
Add goreleaser & a gh workflow

Clivern:
**Why**
It should be handy to get revive binary without installing go.  also I want to add revive inside dockerfile without a dedicated stage to build the binary. So I will have a place to get binaries from (related to https://github.com/mgechev/revive/issues/489):

```zsh
curl -sL https://github.com/mgechev/revive/releases/download/vx.x.x/revive_x.x.x_OS.tar.gz | tar xz
```

**How**
With this PR, every time you create a tag, release github workflow runs to build binaries (linux, windows, mac) and upload them to releases section on github https://github.com/mgechev/revive/releases. It also add the changelog from commits.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,29 @@
+name: Release
+
+on:
+  push:
+    tags:
+      - '*'
+
+jobs:
+  goreleaser:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      -
+        name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+      -
+        name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v2
+        with:
+          version: latest
+          args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


mgechev (approved):

mgechev:
@Clivern looks like the recent actions have been failing. Interesting why without changes the secret seems to no longer work? https://github.com/mgechev/revive/runs/5102785873?check_suite_focus=true

Clivern:
Hey @mgechev,

I think this is related to this PR https://github.com/mgechev/revive/pull/590. It doesn't authenticate with gitHub docker registry to push the docker image. I think it can be fixed by adding this step in https://github.com/mgechev/revive/blob/master/.github/workflows/release.yml

```
      - name: Log into registry
        run: echo "${{ secrets.GITHUB_TOKEN }}" | docker login docker.pkg.github.com -u ${{ github.actor }} --password-stdin
```

cc @doniacld

mgechev:
Thanks for looking into this!

mgechev:
Getting a similar failure [here](https://github.com/mgechev/revive/runs/5118448013?check_suite_focus=true). For some reason docker doesn't consume the token via stdin properly?

Clivern:
yeah, goreleaser push to `ghcr.io` while i used `docker.pkg.github.com`. final fix https://github.com/mgechev/revive/pull/631

mgechev:
The release is green now. Thank you for the fixes :)

PR #507, opened 2021-03-30
Update the list of projects using revive

petethepig:
Just wanted to add pyroscope to the list of happy users.

As a side-effect I also removed a trailing space


mgechev (approved):
❤️

PR #511, opened 2021-04-06
Check string literals against regular expressions

### LINKED ISSUE #510: New rule: check string literals against regular expressions
In some Go programs, string literals will be primarily used for messages that will face users (such as error messages sent from an API). In these cases, it makes sense to have the ability to enforce a specific kind of formatting for these string literals, such as forced (un)capitalization, forcing or forbidding the usage of certain punctuation, etc. This enforcement can be achieved with the `string-regex` rule.

The rule can be configured with multiple regular expressions, along with custom error messages for each expression (optional). String literals are matched against each expression, creating failures for each expression a string *doesn't* match with.

Example configuration:

```toml
[rule.string-regex]
  arguments = [
    ["/^([^A-Z]|$)/", "must not start with a capital letter"],
    ["/(^|[^\\.!?])$/", "must not end in punctuation"],
    ["/^[^\\n]*$/", "must not contain line breaks"]]
```

very-amused:
Closes #510. See the issue for detailed rule information/reasoning. Tests are implemented and passing, docs are updated. I've tested `string-regex` in a real development environment, working on my repository [CSplan-API](https://github.com/very-amused/CSplan-API) in Visual Studio Code.

`string-regex` works properly linting a real project in Visual Studio Code


rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+			rule.Apply(call)
+		}
+	}
+
+	return w
+}
+
+// Return the name of a call expression in the form of package.Func or Func
+func (w lintStringRegexRule) getCallName(call *ast.CallExpr) (callName string, ok bool) {
+	if ident, ok := call.Fun.(*ast.Ident); ok {
```
chavacava, line 153: we could rewrite into something more in the GO style
```go
if ident, ok := call.Fun.(*ast.Ident); ok {
	// Local function call
	return ident.Name, true
} 

if selector, ok := call.Fun.(*ast.SelectorExpr); ok {
	// Scoped function call
	scope, ok := selector.X.(*ast.Ident)
	if !ok {
		return "", false
	}

	return scope.Name + "." + selector.Sel.Name, true
} 

return "", false
```

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+		w.rules = append(w.rules, stringRegexSubrule{
+			Parent:       w,
+			Scope:        scope,
+			Regexp:       regex,
+			ErrorMessage: errorMessage,
+		})
+	}
+}
+
+func (w lintStringRegexRule) parseArgument(argument interface{}, ruleNum int) (scope stringRegexSubruleScope, regex *regexp.Regexp, errorMessage string) {
```
chavacava, line 79: the complexity of the functions related with the parsing of arguments might deserve a dedicated set of unitary test

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+// Apply a single regex rule to a call expression (should be done after verifying the that the call expression matches the rule's scope)
+func (rule stringRegexSubrule) Apply(call *ast.CallExpr) {
+	if len(call.Args) <= rule.Scope.Argument {
+		// TODO: should cases where calls are incompatible with scope cause failures?
+		return
+	}
+
+	arg := call.Args[rule.Scope.Argument]
+	var lit *ast.BasicLit
+	if len(rule.Scope.Field) > 0 {
```
chavacava, line 181: we could encapsulate this logic into a method that retrieves the `*ast.BasicLit` corresponding to the `rule.Scope.Field`

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+}
+
+func (rule stringRegexSubrule) lintMessage(s string, node ast.Node) {
+	// Fail if the string doesn't match the user's regex
+	if rule.Regexp.MatchString(s) {
+		return
+	}
+	var failure string
+	if len(rule.ErrorMessage) > 0 {
+		failure = fmt.Sprintf("string literal doesn't match user defined regex (%s)", rule.ErrorMessage)
```
chavacava, line 223: why not just report the rule error message?
i.e. in place of
`string literal doesn't match user defined regex (must start with a capital letter)`
we could just emit
`must start with a capital letter`

up to the user to provide a nice error message

test/string-regex_test.go
```diff
@@ -0,0 +1,21 @@
...
+func TestStringRegex(t *testing.T) {
+	testRule(t, "string-regex", &rule.StringRegexRule{}, &lint.RuleConfig{
+		Arguments: []interface{}{
+			[]interface{}{
+				"stringRegexMethod1", // The first argument is checked by default
+				"/^[A-Z]/",
+				"must start with a capital letter"},
+
+			[]interface{}{
+				"stringRegexMethod2[2].d",
```
chavacava, line 19: the ability to refer fields of structs might be "too powerful" and maybe not sufficiently useful for the complexity its implementation requires

chavacava (commented):
hi thanks for the new commits, I left some comments

README.md
```diff
@@ -357,6 +357,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | [`unhandled-error`](./RULES_DESCRIPTIONS.md#unhandled-error)   | []string   | Warns on unhandled errors returned by funcion calls    |    no    |  yes   |
 | [`cognitive-complexity`](./RULES_DESCRIPTIONS.md#cognitive-complexity)          |  int   | Sets restriction for maximum Cognitive complexity.              |    no    |  no   |
 | [`string-of-int`](./RULES_DESCRIPTIONS.md#string-of-int)          |  n/a   | Warns on suspicious casts from int to string            |    no    |  yes   |
+| [`string-regex`](./RULES_DESCRIPTIONS.md#string-regex)          |  map   | Warns on string literaly that fail one or more user-configured regular expressions            |    no    |  no   |
```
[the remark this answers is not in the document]
very-amused: Just fixed this typo. I've also updated the rest of the docs for the rule to cover the new, scoped format for arguments. Please let me know if there's anything unclear or worth revising.

RULES_DESCRIPTIONS.md
```diff
@@ -488,6 +489,23 @@ _Description_:  explicit type conversion `string(i)` where `i` has an integer ty
 
 _Configuration_: N/A
 
+## string-regex
+
+_Description_: This rule allows you to configure a list of regular expressions that string literals are checked against.
```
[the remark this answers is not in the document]
very-amused: Answered here: https://github.com/mgechev/revive/pull/511#discussion_r615160684

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+	"fmt"
+	"go/ast"
+	"go/token"
+	"regexp"
+	"strconv"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// #region Revive API
```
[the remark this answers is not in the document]
very-amused: These are #region comments, which specify regions of code that I can collapse in VSCode. I use them to help outline/organize source files by grouping relevant sections of code/data structures.

test/string-regex_test.go
very-amused: One of the main reasons I started working on this feature was because I knew it would be helpful for personal projects I'm working on, as well as hopefully others that do similar work. In an API I've been developing, [I use a struct to describe error responses](https://github.com/very-amused/CSplan-API/blob/2f0b07a6d4d56e8423bfe561e8007e29131fb1ce/core/errors.go#L13), which are then shown to users by a client, so it's desirable to be able to lint `core.WriteError[1].Message`.

I believe that this is a fairly common scenario, but it does have a tradeoff in terms of complexity, which is why I've limited the feature to top level fields.

rule/string-regex.go
very-amused: I've tried to work this out, but it just results in more complexity stemming from the fact that we now need to return both the `*ast.BasicLit` and the success of parsing. By keeping this logic in `Apply`, we can simply return early if parsing fails, as opposed to needing to communicate that need upwards.

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+type stringRegexSubrule struct {
+	Parent       *lintStringRegexRule
+	Scope        stringRegexSubruleScope
+	Regexp       *regexp.Regexp
+	ErrorMessage string
+}
+
+type stringRegexSubruleScope struct {
+	Func     string // Function name the rule is scoped to
+	Argument int    // (optional) Which argument in calls to the function is checked against the rule (the first argument is checked by default)
```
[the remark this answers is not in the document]
very-amused: Yes, I've mentioned this in the documentation I added.

rule/string-regex.go
```diff
@@ -0,0 +1,233 @@
...
+	for _, rule := range w.rules {
+		if rule.Scope.Func == callName {
+			rule.Apply(call)
+		}
+	}
+
+	return w
+}
+
+// Return the name of a call expression in the form of package.Func or Func
```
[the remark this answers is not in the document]
very-amused: Just an escape route in case both casts fail. This shouldn't happen in practice, but I try to never perform casts in Go without checking for success, which would increase the overall risk of undefined/unexpected behavior.

rule/string-regex.go
very-amused: I agree that a set of tests for argument parsing would be good. I don't have the time to implement this tonight, but I can get it done tomorrow.

very-amused:
> hi thanks for the new commits, I left some comments

I've resolved or responded to each of your comments.

chavacava:
Hi @very-amused, I've played around with the new rule, its works OK, nice job. 

Some minor remarks:

1. **action**: in the RULES_DESCRIPTIONS, add an example of configuring a check on struct fields

1. check that parameter indexes in configuration are positive:
  `["errors.Errorf[-5]", "/(^|[^\\.!?])$/", "....."]`
   rises a runtime error
   **action**: fix and add test

1. empty regexp rises a runtime error:
  `["errors.Errorf[0]","", "......"]`
   **action**: fix and add test

1. struct field identifiers containing underscores (_) are not accepted. For example:
  `["errors.Errorf[0].a_b", "/(^|[^\\.!?])$/", "error message must not end in punctuation"]`
  raises an error
  **action**: fix and add test

1. the error messages from the rule configuration parsing need to be more explicit
  `panic: unable to compile /(^|[^\.!?]).($/ as regexp (argument 0, option 1)`
  `panic: unable to parse rule scope (argument 0, option 0)`
  adding information saying that the problem was encountered in when parsing string-regex 
  configuration will help to better understand the error message
  **action**: enrich error messages of the rule

very-amused:
Would it be preferred to call log errors that can be recovered from (i.e: couldn't parse scope for a single argument) with `fmt.Println`, and only panic for those that cannot be recovered from (i.e: the entire set of arguments isn't a slice)? Also, does revive have an internal logger that would be relevant to this?

very-amused:
> Hi @very-amused, I've played around with the new rule, it works OK, nice job.
> 
> Some minor remarks:
> 
>     1. **action**: in the RULES_DESCRIPTIONS, add an example of configuring a check on struct fields
Done, this has already been there since I updated the docs yesterday evening.
>     2. check that parameter indexes in configuration are positive:
>        `["errors.Errorf[-5]", "/(^|[^\\.!?])$/", "....."]`
>        rises a runtime error
>        **action**: fix and add test
This doesn't warrant its own specific error, it's an invalid scope config (which should be simple enough to figure out, given that scopes are not allowed to be excessively complex by design).
>     3. empty regexp rises a runtime error:
>        `["errors.Errorf[0]","", "......"]`
>        **action**: fix and add test
Done, both scope and regex options are checked for length before proceeding in `parseArgument`, so invalid config errors can be produced instead of cases like this resulting in index collisions.
>     4. struct field identifiers containing underscores (_) are not accepted. For example:
>        `["errors.Errorf[0].a_b", "/(^|[^\\.!?])$/", "error message must not end in punctuation"]`
>        raises an error
>        **action**: fix and add test
Done, underscores are accepted in the package/struct name, function name, and/or struct field name in scopes.
>     5. the error messages from the rule configuration parsing need to be more explicit
>        `panic: unable to compile /(^|[^\.!?]).($/ as regexp (argument 0, option 1)`
>        `panic: unable to parse rule scope (argument 0, option 0)`
>        adding information saying that the problem was encountered in when parsing string-regex
>        configuration will help to better understand the error message
>        **action**: enrich error messages of the rule
Done, panic messages now either begin with `invalid configuration for string-regex` if we know it's a user configuration error, or `failed to parse configuration for string-regex` if it's a parsing error we can't blame on the user for certain. All error handling for `parseArgument` has been moved into the methods `w.configError` and `w.parseError`, which makes error reporting more consistent and easy to modify.

chavacava:
> Would it be preferred to log errors that can be recovered from (i.e: couldn't parse scope for a single argument) with `fmt.Println`, and only panic for those that cannot be recovered from (i.e: the entire set of arguments isn't a slice)? Also, does revive have an internal logger that would be relevant to this?
> 
> For now I'm sticking with panicking, but I'm currently separating out panicking with parsing and misconfiguration errors into methods. This way, I could easily add an `isFatal` parameter and log nonfatal errors differently if that is preferred.

there is no log infrastructure. The strategy is to panic at the first problem in the configuration of rules (this assures that the linter either executed under the user defined conditions or not at all

chavacava:
@very-amused , a last proposition before merging: what about renaming the rule as `string-format` ?

Thanks a lot for your work!

very-amused:
I think `string-format` is better suited to the rule in its current state than `string-regex`, so I've gone ahead and renamed it. `string regex`, `string-regex`, and `stringRegex` all turn up no results in a project wide search, which leads me to believe I didn't miss renaming anything.

Working on this has been a great experience for my first major PR. Thank you and everyone else who's worked on revive for the work that you do! This linter is incredibly useful to myself and thousands of others, I'm glad I was able to help make it even better.

PR #512, opened 2021-04-18
Bump revive-action to v2

morphy2k:
I just released a [new major version](https://github.com/morphy2k/revive-action/releases/tag/v2.0.0) which, among other things, fixes the permission issue with PRs.

GitHub recently further restricted the permissions of the token, which severely limits the use when annotations are enabled. I switched to a [new alternative](https://github.com/morphy2k/revive-action/pull/48), this no longer requires direct access to the API.


chavacava:
Thanks @morphy2k!

PR #514, opened 2021-04-30
range-val-address: improve detection

### LINKED ISSUE #513: range-val-address: does not detect specific case
**Describe the bug**
range-val-address does not detect an address reuse

**To Reproduce**
see https://play.golang.org/p/Dp4jBn9d0Ci for details.

**Expected behavior**
Revive should tell the user that this code is bad

**Additional context**
pull request ready.

bernhardreisenberger:
This will improve the range-val-address rule to detect more issues in the code.

Closes #513

### DIFF
--- a/rule/range-val-address.go
+++ b/rule/range-val-address.go
@@ -77,7 +77,7 @@ func (bw rangeBodyVisitor) Visit(node ast.Node) ast.Visitor {
 
 	for _, exp := range asgmt.Rhs {
 		switch e := exp.(type) {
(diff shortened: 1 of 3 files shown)


bernhardreisenberger:
not able to fix travis in this case

chavacava:
Hi @bernhardreisenberger , thanks for the issue and the PR. 
It looks good to me, if you can fix the 2 linter errors in the PR it will be nice.

PR #521, opened 2021-05-20
Adds configuration option to enable all available rules

### LINKED ISSUE #517: Expose the ability to enable all rules
Hello, and thanks for the linter. I have a feature request.

Right now, it appears that the only way that users can enable all rules and disable one rule is to manually specify every single existing rule into the config except for the one specific one that they don't want.

This is tedious, and not very future proof.
Meaning that when you guys add a new useful rule, I don't want to have to manually do research and find out which specific new rule that you added, and then manually add that new rule to my config in order to stay up-to-date. I just want to upgrade my version of `golangci-lint` and automatically get the latest and greatest linting without having to change my config.

As an analogy, golangci-lint allows this kind of workflow with the following YML config:

```yml
linters:
  enable-all: true
  disable:
    # These linters are deprecated
    - interfacer
    - maligned
    - scopelint
```

chavacava:
Closes #517

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,6 @@ enableAllRules = true
     Arguments = [7]
 [rule.function-result-limit]    
     Arguments = [3]
-
 ```
 
 ### Default Configuration
@@ -595,4 +594,3 @@ REVIVE_FORCE_COLOR=1 revive -formatter friendly ./... | tee revive.log
 ## License
 
 MIT
-
--- a/README.md
+++ b/README.md
@@ -240,6 +240,60 @@ warningCode = 0
   severity = "error"
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -240,6 +240,60 @@ warningCode = 0
...
+# Rule tunning
+[rule.argument-limit]
+    Arguments = [5]
+[rule.cyclomatic]    
+    Arguments = [10]
+[rule.cognitive-complexity]    
+    Arguments = [7]
+[rule.function-result-limit]    
+    Arguments = [3]
+
```
mgechev, line 294:
```suggestion
```

PR #528, opened 2021-05-31
simplify: redundant map composite literal

sina-devel:


chavacava:
Hi @sina-devel thanks for the PR

PR #530, opened 2021-06-03
new rule: nested-structs

### LINKED ISSUE #529: Disallow nested structs
No.

Nested structs in projects are annoying and ugly. I'd like a rule that disallows them entirely.

I haven't really considered any alternatives. I'm not sure if there are any.

**Additional context**
For example:

```go
type Parameters struct {
	Options []struct {
		Value string `json:"value"`
	} `json:"options,omitempty"`
}
```

Is awful to look at. This is a 1000x better:

```go
type Option struct {
	Value string `json:"value"`
}

type Parameters struct {
	Options []Option `json:"options,omitempty"`
}
```

rdeusser:
### Please, describe in details what's your motivation for this PR
Nested structs are awful to look at and I want to disallow them entirely in my projects.

### Did you add tests?
Yes.

### Does your code follows the coding style of the rest of the repository?
Yes.

### Does the Travis build passes?
TBD

Closes #529

### DIFF
--- a/rule/nested-structs.go
+++ b/rule/nested-structs.go
@@ -7,12 +7,10 @@ import (
 	"github.com/mgechev/revive/lint"
 )
 
+// NestedStructs lints nested structs.
 type NestedStructs struct{}
 
-func (r *NestedStructs) Name() string {
-	return "nested-structs"
-}
-
+// Apply applies the rule to given file.
 func (r *NestedStructs) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
 	var failures []lint.Failure
 
@@ -39,6 +37,11 @@ func (r *NestedStructs) Apply(file *lint.File, arguments lint.Arguments) []lint.
 	}
 
 	return failures
+}
+
+// Name returns the rule name.
+func (r *NestedStructs) Name() string {
+	return "nested-structs"
 }
 
 type lintNestedStructs struct {
--- a/rule/nested-structs.go
+++ b/rule/nested-structs.go
@@ -41,6 +41,8 @@ type lintNestedStructs struct {
 
 func (l *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
 	switch v := n.(type) {
+	case *ast.FuncDecl:
+		return nil
 	case *ast.Field:
 		if _, ok := v.Type.(*ast.StructType); ok {
 			l.onFailure(lint.Failure{
--- a/rule/nested-structs.go
+++ b/rule/nested-structs.go
@@ -42,6 +42,9 @@ type lintNestedStructs struct {
 func (l *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
 	switch v := n.(type) {
 	case *ast.FuncDecl:
+		if v.Body != nil {
+			ast.Walk(l, v.Body)
+		}
 		return nil
 	case *ast.Field:
 		if _, ok := v.Type.(*ast.StructType); ok {
--- a/rule/nested-structs.go
+++ b/rule/nested-structs.go
@@ -1,7 +1,6 @@
 package rule
 
 import (
-	"fmt"
 	"go/ast"
 
 	"github.com/mgechev/revive/lint"
@@ -27,15 +26,6 @@ func (r *NestedStructs) Apply(file *lint.File, arguments lint.Arguments) []lint.
 
 	ast.Walk(walker, file.AST)
 
-	if walker.count > 0 {
-		walker.onFailure(lint.Failure{
-			Failure:    fmt.Sprintf("no nested structs are allowed, got %d", walker.count),
-			Confidence: 1,
-			Node:       file.AST,
-			Category:   "style",
-		})
-	}
-
 	return failures
 }
 
@@ -45,18 +35,22 @@ func (r *NestedStructs) Name() string {
 }
 
 type lintNestedStructs struct {
-	count     int64
 	fileAST   *ast.File
 	onFailure func(lint.Failure)
 }
 
-func (w *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
+func (l *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
 	switch v := n.(type) {
 	case *ast.Field:
 		if _, ok := v.Type.(*ast.StructType); ok {
-			w.count++
+			l.onFailure(lint.Failure{
+				Failure:    "no nested structs are allowed",
+				Category:   "style",
+				Node:       v,
+				Confidence: 1,
+			})
 			break
 		}
 	}
-	return w
+	return l
 }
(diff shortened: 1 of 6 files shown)


rdeusser:
Tests pass for me locally. CI failed with this error: `The command "openssl aes-256-cbc -K $encrypted_ff954c97a679_key -iv $encrypted_ff954c97a679_iv -in .docs-deploy-key.pem.enc -out .docs-deploy-key -d" failed and exited with 1 during .`

chavacava:
Hi (again) @rdeusser and thanks for the PR.

To add a new rule to the revive machinery it must be added to the [list of rules in the configuration](https://github.com/mgechev/revive/blob/0ade5dc9f879b05a905d583eb87e53a6d1cb370a/config/config.go#L37). Updates of the `README.md` and the ` RULES_DESCRIPTIONS.md ` are also welcome.

rule/nested-structs.go
```diff
@@ -0,0 +1,62 @@
...
+		},
+	}
+
+	ast.Walk(walker, file.AST)
+
+	if walker.count > 0 {
+		walker.onFailure(lint.Failure{
+			Failure:    fmt.Sprintf("no nested structs are allowed, got %d", walker.count),
+			Confidence: 1,
+			Node:       file.AST,
```
chavacava, line 34: During the analysis we have enough information to spot the exact struct that do not respect the rule, therefore we can provide a more precise failure, i.e. a failure attached to the struct and not to the whole file.

Imagine a file with hundreds of lines and one nested struct declaration, as user I expect `revive` to point me the single offensive struct and not the whole file.

rdeusser:
@chavacava Hi!

Added the rule to the list of rules in the config and updated the README as well as the RULES_DESCRIPTION.

testdata/nested-structs.go
```diff
@@ -0,0 +1,13 @@
+package fixtures // MATCH /no nested structs are allowed, got 1/
+
+type Foo struct {
+	Bar struct {
```
chavacava, line 4: As user I expect a failure referring to this line thus the test should read as:

```go
type Foo struct {
  Bar struct { // MATCH /no nested structs are allowed/
  }
}
```

chavacava (changes requested):

testdata/nested-structs.go
rdeusser: Oh I didn't know you could match on individual lines. Nice, I'll fix that.

rule/nested-structs.go
```diff
@@ -0,0 +1,62 @@
...
+	count     int64
+	fileAST   *ast.File
+	onFailure func(lint.Failure)
+}
+
+func (w *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
+	switch v := n.(type) {
+	case *ast.Field:
+		if _, ok := v.Type.(*ast.StructType); ok {
+			w.count++
```
chavacava, line 57: Here we can create a new failure instead of incrementing the counter. Here we have the pointer to the offending struct (`v`)

chavacava:
> @chavacava Hi!
> 
> Added the rule to the list of rules in the config and updated the README as well as the RULES_DESCRIPTION.

Nice, thanks.
I left some comments on the code

rule/nested-structs.go
rdeusser: Good catch. I put that there as I was learning how this worked and forgot to change it.

rule/nested-structs.go
```diff
@@ -0,0 +1,62 @@
...
+func (r *NestedStructs) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	if len(arguments) > 0 {
+		panic(r.Name() + " doesn't take any arguments")
+	}
+
+	walker := &lintNestedStructs{
+		fileAST: file.AST,
+		onFailure: func(failure lint.Failure) {
```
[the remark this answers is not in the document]
rdeusser: Looks like I didn't end up doing anything with that. I'll remove it.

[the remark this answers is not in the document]
rdeusser: Actually I just saw your comment in another thread. It seems I do have a use for this.

rule/nested-structs.go
rdeusser: Yes good call.

rdeusser:
Hey @chavacava I think I got everything. Can you re-review?

chavacava:
I've tested the rule against some codbases (kebernetes, docker), it works well.

The rule spots cases like the following (from `github.com/google/go-cmp/cmp/internal/diff/diff.go:62:31`)
```go
func (es EditScript) stats() (s struct{ NI, NX, NY, NM int }) {
```
and like (from `kubernetes/pkg/volume/git_repo/git_repo_test.go:283:19`)
```go
func doTestPlugin(scenario struct {
	name              string
	vol               *v1.Volume
	expecteds         []expectedCommand
	isExpectedFailure bool
}, t *testing.T) []error {
```
Beyond that it seems to me very hard to read, it is not a _nested_ struct. Should we warn on cases like that? If yes, the rule is not about _nested_ structs but something else. We can also let as is (the above cases are very rare)

In the test cases we could add a case for multiple nests:
```go
type Two struct {
	Bar struct { // MATCH /no nested structs are allowed/
		Baz struct { // MATCH /no nested structs are allowed/
			b   boolean
			Foo struct { // MATCH /no nested structs are allowed/
				b boolean
			}
		}
	}
}
```

rdeusser:
I went ahead and made the change to ignore structs in func declarations and added some additional test cases for that until I hear back from you.

rule/nested-structs.go
```diff
@@ -0,0 +1,58 @@
...
+
+type lintNestedStructs struct {
+	fileAST   *ast.File
+	onFailure func(lint.Failure)
+}
+
+func (l *lintNestedStructs) Visit(n ast.Node) ast.Visitor {
+	switch v := n.(type) {
+	case *ast.FuncDecl:
+		return nil
```
chavacava, line 45:
```suggestion
	case *ast.FuncDecl:
		if v.Body != nil {
			ast.Walk(l, v.Body)
		}
		return nil
```

chavacava:
Hi @rdeusser thanks for the new commit and sorry for the late response.

The new version skips function declarations completely (function bodies are not analyzed) therefore we will not detect nested structs declared in function bodies (they were a large portion of failures detected by the previous version in code bases like kubernetes and docker) I've left a suggestion in the code to let the rule analyze function bodies while skipping function signatures.

rule/nested-structs.go
rdeusser: Oof, yeah didn't see that. Thanks.

rdeusser:
@chavacava Thanks for the review! I got that fixed and added an additional test case for what you mentioned.

chavacava (approved):

chavacava:
Thanks @rdeusser

PR #536, opened 2021-06-29
fix 534

### LINKED ISSUE #534: import-shadowing shouldn't check current package name
**Describe the bug**
```go
// File: github.com/powerman/go-monolith-example/ms/auth/internal/config/config.go
package config
import "github.com/powerman/go-monolith-example/internal/config"
```
```
ms/auth/internal/config/config.go:11:9: import-shadowing: The name 'config' shadows an import name (revive)
package config
        ^
```

**To Reproduce**
I use revive bundled with golangci-lint 1.41.1.

**Expected behavior**
Within package named "x" we should be able to import/create identifiers with same name "x" - there is no shadowing in this case.

 - OS: Gentoo Linux
 - go version go1.16.5 linux/amd64

chavacava:
Closes #534

### DIFF
--- a/rule/import-shadowing.go
+++ b/rule/import-shadowing.go
@@ -23,7 +23,8 @@ func (r *ImportShadowingRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fa
 
 	fileAst := file.AST
 	walker := importShadowing{
-		importNames: importNames,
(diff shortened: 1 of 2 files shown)


mgechev:
There seems to be a linting error. Aside from that LGTM.

mgechev (approved):

PR #542, opened 2021-07-09
Fix testeability typo

derekperkins:


chavacava:
@derekperkins, thanks for the PR!

PR #548, opened 2021-07-28
add tests on common methods handling

chavacava:
Add tests for the rule `exported` to cover the handling of common methods


mgechev (approved):
Good catch!

PR #554, opened 2021-08-09
range-val-address: fix false positive

### LINKED ISSUE #553: range-val-address false positive
**Describe the bug**
A false positive occurs when we're taking the value of a field on a pointer struct.

**To Reproduce**
This is a valid case, but it will get picked up by the linter:
https://play.golang.org/p/DcTOwoJ5cmZ

bernhardreisenberger:
This PR fixes the false positive detected by archseer. i added a few more tests too.

Closes #553

### DIFF
--- a/rule/range-val-address.go
+++ b/rule/range-val-address.go
@@ -16,11 +16,13 @@ func (r *RangeValAddress) Apply(file *lint.File, _ lint.Arguments) []lint.Failur
 	var failures []lint.Failure
 
 	walker := rangeValAddress{
+		file: file,
 		onFailure: func(failure lint.Failure) {
 			failures = append(failures, failure)
 		},
 	}
 
+	file.Pkg.TypeCheck()
 	ast.Walk(walker, file.AST)
 
 	return failures
@@ -32,6 +34,7 @@ func (r *RangeValAddress) Name() string {
 }
 
 type rangeValAddress struct {
+	file      *lint.File
 	onFailure func(lint.Failure)
 }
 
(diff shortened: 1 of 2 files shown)


archseer (approved):

archseer:
Thanks for the quick fix! This looks good to me.

chavacava:
Hi @bernhardreisenberger, thanks for the bug fix (and sorry for the late response)

Testing the fix against Kubernetes code base (`kubernetes/vendor/...`) raised a nil pointer dereference:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x6bbcd2]

goroutine 2022 [running]:
github.com/mgechev/revive/rule.rangeValAddress.Visit(0xc009fe3400, 0xc00a2ff1d0, 0x7d1f60, 0xc009fc5db0, 0x7ceec0, 0xc00b8fd510)
	/workspaces/revive/rule/range-val-address.go:56 +0x72
```

https://github.com/bernhardreisenberger/revive/blob/4f55759aa2d003b1efc1c58623ed2fd0f616bbb3/rule/range-val-address.go#L53-L56
```go
	t := w.file.Pkg.TypeOf(value)

	ast.Walk(rangeBodyVisitor{
		valueIsStarExpr: strings.HasPrefix(t.String(), "*"),
```

It seems that `t` is `nil` in some cases thus `t.String()` fails. The origin of this problem might be that `file.Pkg.TypeCheck()` is returning an error but we do not check for it.

nit: the README.md need to be updated to indicate that the rule is _typed_

PR #570, opened 2021-09-10
fix: minor markdown issue with rules headers

butuzov:
Fix for "headers in rules" description doc, which I am using to create and validate automatic docs and checks for revive integration with `golangci-lint`


chavacava:
Thanks @butuzov !

PR #580, opened 2021-09-25
fixed some complex code

sina-devel:


chavacava:
Hi @sina-devel, thanks for the improvements.

PR #584, opened 2021-09-27
add new rule time equal 

### LINKED ISSUE #583: add new rule to suggest using time.Time.Equal instead of == and !=
according to this [link](https://pkg.go.dev/time#Time) is better used `time.Time.Equal` method instead of == and != in most codes

sina-devel:
Closes #583

### DIFF
--- a/rule/time-equal.go
+++ b/rule/time-equal.go
@@ -20,7 +20,9 @@ func (*TimeEqualRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 	}
 
 	w := &lintTimeEqual{file, onFailure}
-	w.file.Pkg.TypeCheck()
+	if w.file.Pkg.TypeCheck() != nil {
+		return nil
+	}
 
 	ast.Walk(w, file.AST)
 	return failures
--- a/rule/time-equal.go
+++ b/rule/time-equal.go
@@ -1,0 +1,69 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+
+	"github.com/mgechev/revive/lint"
+)
+
+type TimeEqualRule struct{}
+
+// Apply applies the rule to given file.
+func (*TimeEqualRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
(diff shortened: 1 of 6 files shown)


rule/time-equal.go
```diff
@@ -0,0 +1,70 @@
...
+// Apply applies the rule to given file.
+func (*TimeEqualRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	onFailure := func(failure lint.Failure) {
+		failures = append(failures, failure)
+	}
+
+	w := &lintTimeEqual{file, onFailure}
+	w.file.Pkg.TypeCheck()
```
chavacava, line 23: `TypeCheck` might return an error, need to check for it

testdata/time-equal.go
```diff
@@ -0,0 +1,27 @@
+package pkg
```
chavacava, line 1: After configuring `revive` to only apply `time-equal` rule, when doing
```
./revive -config defaults.toml -formatter friendly testdata/time-equal.go
```
No failure is found.
This might be due to (silent) type checking errors.

chavacava (changes requested):
Hi @sina-devel thanks for the PR! 
I've left some comments in the code.
Please add an entry for the new rule in the README.md, and one in the RULES_DESCRIPTIONS.md

rule/time-equal.go
sina-devel: Hi, thanks for the comments, I agree but saw that this error has not been checked in other files too.

chavacava: >this error has not been checked in other files too.
This is something that need to be fixed

sina-devel: > > this error has not been checked in other files too.
> 
> This is something that need to be fixed

well, i thinks `return nil` is good , because the `Apply` method not returns any error.

testdata/time-equal.go
sina-devel: interesting, it works for me :thinking: 😬

mgechev (commented):
LGTM, but I'll let @chavacava give final approval.

PR #587, opened 2021-10-02
Fix config initialisation

### LINKED ISSUE #585: cannot set confidence to 0
**Describe the bug**
It's impossible to set revive's `confidence` setting to `0.0` because this is the default value and is overwritten here:
https://github.com/mgechev/revive/blob/389ba853b0b3587f0c3b71b5f0c61ea4e23928ec/config/config.go#L133

mihaitodor:
Closes #585. Also fixes the "default config" test, which I believe should fail when `reflect.DeepEqual(cfg, tc.wantConfig) == false`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -4,12 +4,17 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"math"
 
 	"github.com/mgechev/revive/formatter"
 
 	"github.com/BurntSushi/toml"
 	"github.com/mgechev/revive/lint"
 	"github.com/mgechev/revive/rule"
+)
+
+const (
+	defaultConfidence = 0.8
 )
 
 var defaultRules = []lint.Rule{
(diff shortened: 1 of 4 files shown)


chavacava:
Hi @mihaitodor, thanks for the PR, I've made some modifications on it with the intention to simplify the configuration management (the need of 3 references to the expression `math.Inf(1)` make me think that something was not right with the whole logic of the original -pre PR- implementation of the configuration management). Let me know what do you think

mihaitodor:
Hi @chavacava, you're welcome! I'm quite new to this codebase and was trying my best to reuse that `defaultConfig()` helper, which, unfortunately, makes things difficult :) I'm used to config loader libraries which look at the `Config` struct tags for the default values (see [caarlos0/env](https://github.com/caarlos0/env) for example), but I know not everyone likes that approach.

mgechev (approved):

PR #588, opened 2021-10-07
Fix typo for gorush description

abeltay:
Fix a typo in the description for https://github.com/appleboy/gorush. Copied the `.` over also as it looks like there's no standard regarding whether we should end with `.`. Please correct me if you want it removed.


chavacava:
@abeltay thanks for the PR

Issue #241, opened 2019-10-17
Is it possible for revive to return a non-zero exit code when there are issues and zero if not?


inliquid:
I'd like to run `revive` in some sort of automated test during build. However I found that it returns zero in any case. This makes it more complicated to check result of running the tool.

chavacava:
Hi @inliquid, in the configuration you can set 

```
# Sets the error code for failures with severity "error"
errorCode = 0
# Sets the error code for failures with severity "warning"
warningCode = 0
```
to the values you need.

mgechev:
@inliquid I'm closing the issue. Feel free to comment/open a new issue in case I misunderstand.

jan-xyz:
That would be my favorite solution 🤗 I would also offer to provide the PR for it, if you are ok with adding it

jan-xyz:
I filed #589 , let me know what you think 🤗

PR #589, opened 2021-10-10
overwrite error status code from config with set_exit_status

jan-xyz:
This PR is a result from [a discussion](https://github.com/mgechev/revive/issues/241#issuecomment-939269291) in an already closed issue. It copies the functionality from golang/lint to set _only_ the error exit code to 1 and leave everything else untouched.

I didn't add tests for this change, but I can do so, if you point me in a good place to add it.

relates to #241


chavacava:
Hi @jan-xyz, thanks for the PR, I've left a couple of comments

PR #604, opened 2021-10-19
Add debug logger

### LINKED ISSUE #592: Revive in mode debug

During the development of a new rule, I had to add `Println` to debug my code when running the revive binary.

I would like to add logs and activate a debug mode to revive when the corresponding flag is activated.
So I can see the logs and verify the steps of the analyse.

```
Example:
  revive -config c.toml -formatter friendly -exclude a.go -exclude b.go ./...

Usage of revive:
  -config string
        path to the configuration TOML file, defaults to $HOME/revive.toml, if present (i.e. -config myconf.toml)
  -exclude value
        list of globs which specify files to be excluded (i.e. -exclude foo/...)
  -formatter string
        formatter to be used for the output (i.e. -formatter stylish)
  -version
        get revive version
  -debug
        activate debug mode
```

dvejmz:
Provide logger implementation that allows writing application logs to a file if `DEBUG` mode environment variable is enabled.

Closes #592.

Example usage

```
DEBUG=1 revive
```

This will write a log file called `revive.log` created in the current working directory, where all subsequent application logs will be appended.

The implementation is very basic as I thought we could start with something simple and build from there, if necessary. I decided to use an environment variable instead of a command-line flag to make it easier to activate this functionality in CI environments. If people have a strong preference to use the command-line option instead, I'm happy to change it (or allow both).

Another assumption I made was that we'd want to write application logs to a file instead of stdout, so as not to mix up application output with debug information.

Feedback welcome on the above or anything else I may have missed!

### DIFF
--- a/logging/logger.go
+++ b/logging/logger.go
@@ -9,6 +9,8 @@ import (
 
 var logger *log.Logger
 
+// GetLogger retrieves an instance of an application logger which outputs
+// to a file if the debug flag is enabled
 func GetLogger() (*log.Logger, error) {
 	if logger != nil {
 		return logger, nil


dvejmz:
Yeah, I see where you're coming from with that @chavacava . I think perhaps what this PR actually does is not introduce a "debug" mode per se which alters the behaviour of revive, but rather give devs a simple, self-contained logger that can be dropped into any package they happen to be working on at the time and allow them to trace application state wherever it's needed. I added a call to the log in `main.go` as an example of how it could be used but for the most part, I think it's up to people to decide how much information they want to record via this new verbose mode and perhaps remove those logging statements before merging back into master if they have no further use for them.

I guess the other nice thing about a logger that can be transparently switched on and off is that if people forgot to remove a debug `log.Println` statement somewhere before merging in a change, it would be suppressed anyway.

PR #612, opened 2021-11-20
refactor: use os and io packages intead of ioutil

Clivern:
- `io/ioutil` is deprecated so i replaced it according to https://golang.org/doc/go1.16#ioutil
- Updated goreleaser action to use latest go version `1.17`.

Closing the PR since some of these methods supported only on go >= 1.16. so we will have to drop older go versions.


chavacava:
Thanks for the PR anyway!

PR #616, opened 2021-12-15
Allow whitelist for the context parameter check

### LINKED ISSUE #605: Allow `*testing.T` before `context.Context` in the context parameter check
This issue mirrors the upstream one [here](https://github.com/golang/lint/issues/422), which ended up stalling out due to upstream abandoning golint.

The specific case I'm referring to looks like so:

```go
func TestFoo(t *testing.T) {
  ctx, cancel := context.WithCancel(context.Backgrdoun())
  defer cancel()
  t.Run("subtest1", func(t *tesitng.T) {
    ctx, cancel := context.WithCancel(ctx)
    fooSubtest1(t, ctx)
  })
}

func fooSubtest1(t *testing.T, ctx context.Context) {
  // do things that require a context
}
```

Currently, the `context-as-argument` lint will complain that `fooSubtest1` should be written with the signature `func(context.Context, *testing.T)` instead.

To me, it's idiomatic for `*testing.T` to remain the first argument of `fooSubtest` since it's effectively being curried in. Ideally, if this language supported currying, we'd do something more akin to `subtest1 := t.Run("subtest1", fooSubtest1); subtest1(ctx)`, but without language support for currying, it seems like manually passing through the `t` first makes more sense.

So, that leads me to the question for this issue:
Should we update revive to allow this? Is there some other hoops we should jump through first?

I'm happy to provide the code changes necessary to make the current context lint accept this (and have them written up [over here](https://github.com/euank/revive/commit/9a452fbac3ed67c59ca37fe73404b99729d192ba) in fact.)

euank:
This allows users to configure a set of types that may appear before
`context.Context`.

Notably, I think this rule is useful for allowing the `*testing.T` type
to come before `context.Context`, though there may be other uses (such
as putting a tracer before it, etc).

See #605 for a little more context on this.

Fixes #605

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -186,7 +186,16 @@ _Configuration_: N/A
 
 _Description_: By [convention](https://github.com/golang/go/wiki/CodeReviewComments#contexts), `context.Context` should be the first parameter of a function. This rule spots function declarations that do not follow the convention.
 
-_Configuration_: N/A
+_Configuration_:
+
+* `allowTypesBefore` : (string) comma-separated list of types that may be before 'context.Context'
+
+Example:
+
+```toml
+[rule.context-as-argument]
+  arguments = [{allowTypesBefore = "*testing.T,*github.com/user/repo/testing.Harness"}]
+```
 
 ## context-keys-type
 
@@ -252,7 +261,7 @@ _Configuration_: N/A
 _Description_: In GO it is idiomatic to minimize nesting statements, a typical example is to avoid if-then-else constructions. This rule spots constructions like
 ```go
 if cond {
-  // do something 
+  // do something
 } else {
   // do other thing
   return ...
@@ -262,7 +271,7 @@ that can be rewritten into more idiomatic:
 ```go
 if ! cond {
   // do other thing
-  return ... 
+  return ...
 }
 
 // do something
@@ -314,8 +323,8 @@ _Description_: Exported function and methods should have comments. This warns on
 
 More information [here](https://github.com/golang/go/wiki/CodeReviewComments#doc-comments)
 
-_Configuration_: ([]string) rule flags. 
-Please notice that without configuration, the default behavior of the rule is that of its `golint` counterpart. 
+_Configuration_: ([]string) rule flags.
+Please notice that without configuration, the default behavior of the rule is that of its `golint` counterpart.
 Available flags are:
 
 * _checkPrivateReceivers_ enables checking public methods of private types
@@ -425,8 +434,8 @@ Example:
 ```
 ## import-shadowing
 
-_Description_: In GO it is possible to declare identifiers (packages, structs, 
-interfaces, parameters, receivers, variables, constants...) that conflict with the 
+_Description_: In GO it is possible to declare identifiers (packages, structs,
+interfaces, parameters, receivers, variables, constants...) that conflict with the
 name of an imported package. This rule spots identifiers that shadow an import.
 
 _Configuration_: N/A
@@ -503,7 +512,7 @@ _Configuration_: N/A
 
 ## range-val-address
 
-_Description_: Range variables in a loop are reused at each iteration. This rule warns when assigning the address of the variable, passing the address to append() or using it in a map. 
+_Description_: Range variables in a loop are reused at each iteration. This rule warns when assigning the address of the variable, passing the address to append() or using it in a map.
 
 _Configuration_: N/A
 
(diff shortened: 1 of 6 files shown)


rule/context-as-argument.go
```diff
@@ -1,22 +1,49 @@
...
 // ContextAsArgumentRule lints given else constructs.
-type ContextAsArgumentRule struct{}
+type ContextAsArgumentRule struct {
+}
 
 // Apply applies the rule to given file.
-func (r *ContextAsArgumentRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+func (r *ContextAsArgumentRule) Apply(file *lint.File, args lint.Arguments) []lint.Failure {
+	var allowTypesBefore []string
+	if len(args) > 1 {
```
mgechev, line 17: Can we have this panic if `len(args) != 1` so we can skip the second if statement?

rule/context-as-argument.go
```diff
@@ -33,20 +60,48 @@ func (r *ContextAsArgumentRule) Name() string {
...
 }
 
 func (w lintContextArguments) Visit(n ast.Node) ast.Visitor {
 	fn, ok := n.(*ast.FuncDecl)
 	if !ok || len(fn.Type.Params.List) <= 1 {
 		return w
 	}
+	allowTypesLookup := make(map[string]struct{}, len(w.allowTypesBefore))
+	for _, v := range w.allowTypesBefore {
+		allowTypesLookup[v] = struct{}{}
```
mgechev, line 76: Curious why not a boolean value?

rule/context-as-argument.go
```diff
@@ -33,20 +60,48 @@ func (r *ContextAsArgumentRule) Name() string {
...
+	}
+
+	fnArgs := fn.Type.Params.List
+	// trim off all types we've been configured to skip
+	for len(fnArgs) > 0 {
+		typeStr, ok := astExprTypeStr(fnArgs[0].Type)
+		if !ok {
+			// assume we're done. This can happen, for example, with a function type
```
mgechev, line 84: Shouldn't we just `continue` here to go to the next argument?

mgechev (commented):
Thank you for the feature! Left a few comments. Will let @chavacava have a look as well.

rule/context-as-argument.go
euank: Sure, done

rule/context-as-argument.go
euank: I find struct{} is more clear for sets. If it's a string -> bool map, I have to wonder if both true/false in the bool have meaning, or only true (i.e. if it's a set or a map). If the value's a struct{}, I immediately know it's a set of strings.

rule/context-as-argument.go
```diff
@@ -33,20 +57,48 @@ func (r *ContextAsArgumentRule) Name() string {
...
+	}
+
+	if len(fnArgs) <= 1 {
+		return w
+	}
 	// A context.Context should be the first parameter of a function.
 	// Flag any that show up after the first.
-	previousArgIsCtx := isPkgDot(fn.Type.Params.List[0].Type, "context", "Context")
-	for _, arg := range fn.Type.Params.List[1:] {
+	previousArgIsCtx := isPkgDot(fnArgs[0].Type, "context", "Context")
```
chavacava, line 100: We might simplify the check with something like:
```go
ctxIsAllowed := true
for _, arg := range fnArgs {
		argIsCtx := isPkgDot(arg.Type, "context", "Context")
		if argIsCtx && !ctxIsAllowed {
			w.onFailure( ... )
		}
		// check if the type is in the allowTypesLookup  map or is context.Context
		// if is not the case, make ctxIsAllowed = false
}
```

This lets us avoid doing all the parameter list trimming stuff

RULES_DESCRIPTIONS.md
```diff
@@ -186,7 +186,16 @@ _Configuration_: N/A
 
 _Description_: By [convention](https://github.com/golang/go/wiki/CodeReviewComments#contexts), `context.Context` should be the first parameter of a function. This rule spots function declarations that do not follow the convention.
 
-_Configuration_: N/A
+_Configuration_:
```
chavacava, line 189: the README.md must be also updated to document this new configuration for the rule

chavacava (changes requested):
Hi @euank, thanks for the PR, I`ve left some comments

RULES_DESCRIPTIONS.md
```diff
@@ -186,7 +186,16 @@ _Configuration_: N/A
...
+
+* `allowTypesBefore` : (string) comma-separated list of types that may be before 'context.Context'
+
+Example:
+
+```toml
+[rule.context-as-argument]
+  arguments = [{allowTypesBefore = "*testing.T,*github.com/user/repo/testing.Harness"}]
```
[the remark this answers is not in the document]
chavacava: I've modified the config parsing code to accept a comma separated list of type names (like in the example above)

mgechev (approved):

PR #618, opened 2021-12-26
var-naming rule: note about initialisms

kulti:
It is not clear, why var-naming rule force users to write HTTP instead of Http. I've updated documentation about that.


mgechev (approved):

mgechev:
Thanks for the PR!

PR #625, opened 2022-01-24
Limit concurrency to avoid hitting file descriptor limit

### LINKED ISSUE #624: too many open files
When running revive over a medium size codebase (~1500 files) it occasionally hits the system file descriptor limit.

Originally we suspected Chrome eating up FDs, but this still happens even with no other software running on the system, and with high system wide limits set.

After some profiling it seems that revive will spawn a new goroutine for every file upon calling `revive.Lint`, all at the same time.

**To Reproduce**

```
/go/.bin/revive -formatter=stylish -config=.revive.toml ./cmd/... ./pkg/...
open pkg/service/flags/flags.go: too many open files
```
<details>
<summary>.revive.toml</summary>

```toml
ignoreGeneratedHeader = false
severity = "error"
confidence = 0.8
errorCode = 1

# https://github.com/mgechev/revive#available-rules
[rule.atomic]
[rule.blank-imports]
[rule.bool-literal-in-expr]
[rule.confusing-results]
[rule.constant-logical-expr]
[rule.context-keys-type]
[rule.dot-imports]
[rule.empty-block]
[rule.error-naming]
[rule.error-return]
[rule.errorf]
[rule.get-return]
[rule.if-return]
[rule.increment-decrement]
[rule.indent-error-flow]
[rule.modifies-parameter]
[rule.range-val-address]
[rule.range-val-in-closure]
[rule.range]
[rule.receiver-naming]
[rule.redefines-builtin-id]
[rule.string-of-int]
[rule.struct-tag]
[rule.superfluous-else]
[rule.time-naming]
[rule.unexported-return]
[rule.unreachable-code]
[rule.var-declaration]
[rule.waitgroup-by-value]
```

</details>

**Expected behavior**

Should lint the full repo without incident.

**Desktop:**
 - OS: MacOS
 - go 1.17.5

I have a proposed solution - PR will follow.

ricardobeat:
The current implementation spawns a new goroutine for every package,
all at the same time.

This will eventually cause it to hit the system max FD limits, even
when they are set to very high values (50k/100k+)

Setting max = number of CPUs is arbitrary; regardless of attempts at manually tuning this number, running time remains essentially the same. Memory usage is
significantly reduced - from 2GB down to 300MB in my test project
with around 1500 files.

Note: I honestly haven't written a lot of async code in Go, let me know if the approach here is wrong.

The change has been tested on two different developer machines that were displaying the `too many open files` issue.

Closes #624

### DIFF
--- a/lint/linter.go
+++ b/lint/linter.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"os"
 	"regexp"
+	"runtime"
 	"strconv"
 	"sync"
 )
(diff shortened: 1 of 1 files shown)


mgechev:
The change makes sense to me. Since chances are that @chavacava has written way more Go code recently, I'll let him have a final look before we get it in.

chavacava:
Hi @ricardobeat, thanks for filling the issue and proposing a solution.
I will check your proposal. I not convinced that we should hardcode concurrency limits in `revive`. It might be necessary to add a cmd line flag to set the concurrency limit.
Setting limits, hardcoded or not, has also performance consequences. I need to check how these limits affect the linting perf in order to provide guidance to users.
I'll also try to find other potential solutions to the "too many open files" problem

mgechev:
@ricardobeat, @chavacava fixed the issue with a command line flag https://github.com/mgechev/revive/pull/628.

Thank you for pointing this out and thanks for the PR!

PR #639, opened 2022-02-24
Fix module and add `recover` for `panic` recovery

subham-deepsource:
revive's some not-so-well-written rules miss out on many corner cases leading to runtime panics. These runtime panics are a problem for the dependent packages that revive as an API. Not only that, the panics occur in individual goroutine spawned from a method of revive, i.e., that cannot be controlled by the dependent packages and hence can't be recovered.

We could have recovered from these panics by using the built-in `recover` but it only works if used in the same goroutine that panics.

**Example**:
```
main.f() -> main.g() -> revive.h() // example call heirarchy
```

Suppose `f` and `g` belong to main packages and `f` calls `g` and `g` calls `h`, which is a part of the revive. Now, `h` spawns one goroutine per file, it runs diagnostics, and in those goroutines, one of them panics, but we cannot control them as the control of the goroutine isn't with the `main` package.

That's why forking and applying a patch. Also made necessary changes to the module to make it importable by the dependent package.


subham-deepsource:
My bad. Was supposed to open a PR against my fork.

PR #643, opened 2022-02-26
Format sources w/ gofumpt

shmsr:
Format sources w/ gofumpt, i.e., no change to tests, testdata, etc., where formatting could've been a problem. Safe to merge.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -180,7 +180,7 @@ const defaultConfidence = 0.8
 
 // GetConfig yields the configuration
 func GetConfig(configPath string) (*lint.Config, error) {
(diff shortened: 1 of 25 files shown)


chavacava:
Hi @shmsr thanks for the PR.
I prefer to stick to official tooling as much as possible. 
Moreover, switching to `gofumpt` in place of `gofmt` will require all `revive` contributors to also switch.

(To avoid PR that are finally not merged we ask to open an issue before making the PR)

shmsr:
But gofmt is a subset of gofumpt, so if everyone keeps using gofmt, it is not a problem because it'd be a no-op!

PR #644, opened 2022-02-26
Use Go 1.17 and upgrade dependencies

shmsr:
**Changes include**

- Update go version in go.mod to 1.17 to benefit from improvements related to module pruning, etc.
- Upgrade dependencies

From the commit message:
```
Up to Go 1.16, a module graph might contain module requirements that have no
impact on building the main module.

Go 1.17 prunes the module graph from those "impact-less" dependencies, except
for cases that require backward compatibility to module files created with
Go 1.16 or earlier.

If a module specifies "go 1.17" or higher, the module graph includes only the
immediate dependencies of other go 1.17 modules, not their full transitive
dependencies.

But, changing the "go.mod" (which could also change "go.sum") could break
the builds for revive who are not using Go 1.17 or later i.e., users who try
to build revive from source. For that, we will use:

"""
go mod tidy -compat=1.12
"""

The "-compat" flag allows that version to be overridden to support older
(or only newer) versions, up to the version specified by the go directive in
the go.mod file. So, "-compat=1.12" keeps the builds working for Go 1.12
to the latest Go release (for Go 1.17 or above, module pruning also exists).

So, this is safe change.

Signed-off-by: subham sarkar <sarkar.subhams2@gmail.com>
```


shmsr:
>  @shmsr thanks for the PR. First at all, I think `revive` should update its `go` version to something newer than `1.12`. It is something that should be discussed with other users to check if the upgrade will break their builds (personally I don't think there are a lot of people building `revive` with go 1.12)

As per [Go's release policy](https://go.dev/doc/devel/release):

```
Each major Go release is supported until there are two newer major releases. For example, Go 1.5 was 
supported until the Go 1.7 release, and Go 1.6 was supported until the Go 1.8 release. We fix critical 
problems, including [critical security problems](https://go.dev/security), in supported releases as 
needed by issuing minor revisions for example, Go 1.6.1, Go 1.6.2, and so on).
```

So, I think tools should have a similar policy like this. There should ideally be a threshold after which we should upgrade. So, based on Go's policy, we should just think of Go 1.15+.

People getting the executables from [releases](https://github.com/mgechev/revive/releases) are not affected at all and those who are building it from source (go install, go get, go build) already know how to fix things if something happens — `go mod tidy` to the rescue.

So, I think it is safe to use: 
```
go mod tidy -compat=1.15
```

---

> What I'm not sure is about the _safety_ of the approach proposed by the PR: rely on commiters discipline to run `go mod tidy` always including the `-compat=1.12` flag.


I think this needs to be followed by only Go 1.17+ users because:

1. Go 1.17+ only knows how to prune the "impact-less" dependencies
2. Go version before Go 1.17 can continue using `go mod tidy` because `go mod tidy` for Go 1.17+ will fix it (if necessary).

Also, if any commit going forward changes `go.{mod,sum}` (which can only happen in dependency upgrades, downgrades, etc. or addition or removal of dependencies), we can always check if the checksums for Go 1.15+ exists by using:
```
go mod tidy -compat=1.15
```

See, someday, we have to upgrade. Why not now? Also, why not draft a release policy like Go?

chavacava:
Hi @shmsr sorry for the late response.
I've checked the GO version of `golangci-lint`, it seems the most popular way of using `revive` nowadays, and it [compiles with GO 1.16](https://github.com/golangci/golangci-lint/blob/master/go.mod) thus I'm not sure if upgrading to 1.17 from our side will be without impact for `golangci-lint`. 
We could "safely" upgrade to 1.16 or bet into an upgrade to `1.17` without breaks

PR #647, opened 2022-03-08
range-val-address: improve detection

bernhardreisenberger:
This pr will improve the range-val-address rule. i added tests to display the issue.

#513


chavacava:
Hi @bernhardreisenberger, thanks for th PR

PR #650, opened 2022-03-16
Allow revive to be called with extra linters

heynemann:
This change allows revive to be called from other libraries and pass in a list of custom linters to be added to the built-in linters found in config.

After this change, users who want to extend Revive can do it without forking. It's a simple matter of creating a new folder with `main.go`:
```go
package main

import "github.com/mgechev/revive/cli"

func main() {
    cli.RunRevive(&CustomRule{}, &OtherRule{})
}

// Rules ommitted for clarity, but they could be in this file,
// or in another package. Could even be packs of rules
// added as dependency of the project
```

Then in their .toml file:

```toml
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.8
errorCode = 0
warningCode = 0

[rule.blank-imports]
[rule.context-as-argument]
[rule.context-keys-type]
[rule.dot-imports]
[rule.error-return]
[rule.error-strings]
[rule.error-naming]
[rule.exported]
[rule.if-return]
[rule.increment-decrement]
[rule.var-naming]
[rule.var-declaration]
[rule.package-comments]
[rule.range]
[rule.receiver-naming]
[rule.time-naming]
[rule.unexported-return]
[rule.indent-error-flow]
[rule.errorf]
[rule.custom-rule]
  arguments = [0]
```

As soon as they remove [rule.custom-rule] from their config, it is not run (even though it is passed as an extra rule.

The motivation for this PR is that I intend to base the custom linters for my company on revive and do not wish to maintain a fork of it. It feels much safer to just use a function that does everything and receives the extra rules as arguments.

The change in revive's codebase is so small that it felt like a good way moving forward to enable even easier extensibility for this amazing project.

If you feel like this is a valuable contribution I'm more than happy to add tests and update the docs on how to create and use custom rules without forking.


heynemann:
Can confirm that using my fork I already set up revive with additional linters in our repo. It is working great :D That way I can build and distribute internal versions of our linter to every eng in the company (using brew probably). 

Thanks once again for such a great tool!

chavacava:
Hi @heynemann thanks for the PR. Great idea!

I'll check it this we.

chavacava:
@heynemann, I did a small refactoring in the `RunRevive` function. could you check it please.

If you are Ok with the refactoring, I'll go on with the merge.

heynemann:
Definitely ok with me :) Already using this PR to write new rules and it is a bliss :)

heynemann:
Thank you very much for merging this. Is there any rule about being in the readme as a contributor?

chavacava:
Just modify the README.md

Issue #654, opened 2022-03-21
Allow configuration to be passed in to library mode


heynemann:
I was checking how golangci-lint [uses revive](https://github.com/golangci/golangci-lint/blob/30c6166b08117b0b19ec44c26c7e97c1506656ba/pkg/golinters/revive.go) and they seem to have copied a substantial part of the codebase.

Now that it is possible to run revive as a library I was thinking of making a PR to golangci-lint to improve revive's usage there. The only thing missing is passing in configuration and being able to call revive without the CLI arguments (why they copied I believe).

Do you think it is a good step forward? Decoupling the CLI part with the linting part of revive? I can get a draft PR...

chavacava:
After merging your last PR I immediatly thinked on golang-ci and how we could facilitate the integration of revive.

heynemann:
Will get a draft PR as soon as possible for us to work on this...

chavacava:
Closed by #655

PR #655, opened 2022-03-22
Separating lib from cli

heynemann:
This PR corresponds to issue #654. 

It extends Revive usage as a library to more than just cli usage. Clients can now use parts of the revive workflow (Lint, Format) and pass configuration to the flow themselves, instead of loading from a toml file.

After this PR users can use the library with:

```go
        // removed error handling for clarity
	conf, _ := config.GetConfig("./defaults.toml")

	revive, _ := revivelib.New(
		conf,  // config
		true,  // set exit code
		2048,  // max open files
		[]string{"./tests"},  // exclude
		revivelib.NewExtraRule(&mockRule{}, lint.RuleConfig{}),  // extra rules
	)

	// Now they can lint
     	failuresChan, _ := revive.Lint("../testdata/if-return.go")

	// Get the []lint.failure list
        failureList := revive.GetLintFailures(failuresChan)

        // Or just get the formatted output and exit code
	failures, exitCode, _ := revive.Format("stylish", failuresChan)
```


heynemann:
@chavacava can you review this PR? If you want I can change revivelib to just lib. I just felt like doing:

```go
revive, err := revivelib.New(...)
```
Feels nicer than `lib.New(...)`.

chavacava:
Hi @heynemann, I did not forget your PR, I'll try to check it this we. Thanks!

cli/main.go
```diff
@@ -29,155 +26,44 @@ func fail(err string) {
...
 	}
 
-	log.Println("Config loaded")
-
-	if len(excludePaths) == 0 { // if no excludes were set in the command line
-		excludePaths = conf.Exclude // use those from the configuration
-	}
-
-	packages, err := getPackages(excludePaths)
+	failures, err := revive.Lint(flag.Args()...)
```
chavacava, line 47: here I'm not convinced with only passing `includePatterns` and not also `excludePatterns`.
my feeling is both should be handled in the same way: either they are both a property of the `revive` lib instance or both are parameters of the `Lint` method.
I prefer to pass them as arguments to the `Lint` methods because it will allow a flexible usage patter: create a `revive` instance and then call `Lint` many times with different  include & exclude patterns. 

What  do you think?

revivelib/core.go
```diff
@@ -0,0 +1,210 @@
...
+	failures, err := revive.Lint(packages, r.lintingRules, *r.config)
+	if err != nil {
+		return nil, errors.Wrap(err, "linting - retrieving failures channel")
+	}
+
+	return failures, nil
+}
+
+// GetLintFailures gets the list of failures for a given failures channel from Lint.
+func (r *Revive) GetLintFailures(failuresChan <-chan lint.Failure) []lint.Failure {
```
chavacava, line 101: Not sure about the added value of this method.
We do not use it and other users of the lib can implement it very easily.

chavacava (approved):
LGTM, I've left a couple of comments/questions but nothing relevant.
Let me know when you are ok to merge

cli/main.go
heynemann: I was thinking more on the lines of "Lint these packages". The excludes part applies to all of them. But I can see your point and have no strong feelings. Will change!

revivelib/core.go
heynemann: I wasn't sure either, but feels like it's a good helper method. If you prefer I can remove.

heynemann:
Sorry for the initial low quality of this PR. It feels much better now. If you feel like it is up to par, just merge it :D

chavacava:
@heynemann thanks for the PR! Nice work.

Issue #685, opened 2022-05-17
More initialisms


scop:
**Is your feature request related to a problem? Please describe.**

I can think of a bunch of common acronyms that would be good to have built in as initialisms in revive.

**Describe the solution you'd like**

BGP, DB (as in database), MAC (as in address), SCSI, SDN, VRF added.

MAC could be a bit controversial, but I don't see the other ones causing false positives. Not all of them are _that_ common though, but I'd say at least DB is.

**Describe alternatives you've considered**

Local config.

**Additional context**

None.

chavacava:
Hi @scop 
The rule `var-naming` already accepts as parameter a list of initialisms to check and a list of initialisms to ignore.
You can check the [rule's documentation](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming)
For technical details check the [related commit](https://github.com/mgechev/revive/commit/90f51530cc4b7bc0aafc9f8a28c874f1a43143fa)

scop:
I know, that's what I referred to by "local config" in the initial commit. The point here was to add some common built in ones that people would not have to do that.

PR #686, opened 2022-05-17
Add some common initialisms

scop:
Closes https://github.com/mgechev/revive/issues/68


chavacava:
Hi @scop thanks for the PR

revive's list of initialisms is [that of `golint`](https://github.com/golang/lint/blob/6edffad5e6160f5949cdefc81710b2706fbcd4f6/lint.go#L770) because `revive` is a _drop-in replacement of golint_, thus hard-coding new initialisms will break the promise of "drop-in replacement"

What about adding an (optional) configuration for the rule in order to enrich the list of initialisms?

scop:
I understand being a drop-in replacement may have been valuable at some point, but given that golint is deprecated and frozen since at least a year, quite likely several, I'm not sure it's a good strategy to be bound by its features or lack of them indefinitely. I'd personally just let go of that promise whenever it holds back progress (such as IMHO here).

Could you elaborate on how the optional configuration for this rule you're thinking of would be used by end users? If it's easy and clear enough, then I guess I might be interested. But then again if one needs to configure something in the first place, one could go ahead and just configure the missing initialisms...

chavacava:
Please check the comment on the related issue https://github.com/mgechev/revive/issues/685#issuecomment-1129888672

PR #689, opened 2022-05-19
fix(var-naming): set node to package name for underscore in package name

### LINKED ISSUE #688: Enther AST as node for underscore in package name var-naming issues
**Describe the bug**
Underscore in package name var-naming violations get the entire source of the file printed as context in golangci-lint output.

**To Reproduce**

```
$ golangci-lint run --disable-all --enable revive /tmp/foo.go
../../tmp/foo.go:1:1: var-naming: don't use an underscore in package name (revive)
package foo_bar

func main() {
}
```

**Expected behavior**

```
$ golangci-lint run --disable-all --enable revive /tmp/foo.go
../../tmp/foo.go:1:1: var-naming: don't use an underscore in package name (revive)
package foo_bar
        ^
```

**Logs**
See reproducer and expected behavior.

 - OS: Ubuntu 20.04
 - Version of Go: 1.17.10

**Additional context**
N/A

scop:
Setting the entire file AST as the node causes golangci-lint to print the entire file source as the context, and line and column numbers set to 1. Point to the package name node instead.

Closes #688

### DIFF
--- a/rule/var-naming.go
+++ b/rule/var-naming.go
@@ -56,7 +56,7 @@ func (r *VarNamingRule) Apply(file *lint.File, arguments lint.Arguments) []lint.
 		walker.onFailure(lint.Failure{
 			Failure:    "don't use an underscore in package name",
 			Confidence: 1,
-			Node:       walker.fileAst,
+			Node:       walker.fileAst.Name,
 			Category:   "naming",
 		})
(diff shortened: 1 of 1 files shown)


chavacava:
@scop, thanks for the catch and the PR

PR #692, opened 2022-05-30
fix(receiver-naming): distinguish types with parameters

### LINKED ISSUE #669: Distinguish generic types
**Describe the bug**
Revive reports receiver name error at first and only declaration.

**To Reproduce**
Run `revive` on this code:
```go
package example

type Foo[T any] struct{}

func (f *Foo[T]) Foo() {}

type Bar[T comparable] struct{}

func (b *Bar[T]) Bar() {}
```

```shell
 $ revive --version && revive
version 1.1.4
example.go:9:1: receiver name b should be consistent with previous receiver name f for invalid-type
example.go:3:6: exported type Foo should have comment or be unexported
example.go:7:6: exported type Bar should have comment or be unexported
```

**Expected behavior**
1. I expected to *not* see any errors regarding receiver types.
2. The receiver error should not report "invalid-type".

 - OS: Arch Linux
 - go version go1.18 linux/amd64

tie:
This PR adds a new `internal/typeparams` package that provides support for type parameters when revive is built with Go 1.18 and higher.

It fixes an issue with `receiver-naming` rule being unable to parse receiver type from AST method declarations the receiver has type parameters.

Closes #669

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,6 @@
-name: Build and Test
+name: Build
 on:
+  workflow_dispatch:
   pull_request:
     types: [opened, edited, synchronize, reopened]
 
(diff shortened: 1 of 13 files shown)


chavacava:
Hi @tie, thanks for the PR (and sorry for the late response)

PR #694, opened 2022-06-01
Make package comment more confident

### LINKED ISSUE #693: Package comment is not confidence
Currently, a file without doc is reported by rule `package-comments` with confidence 0.2 .
but it is not check is there other files in package have doc.
I want make this rule more confident

check package comment in package level

**Additional context**
I have build a [branch](https://github.com/okhowang/revive/tree/feat/package-comment) for this issue

okhowang:
close #693

### DIFF
--- a/rule/package-comments.go
+++ b/rule/package-comments.go
@@ -5,6 +5,7 @@ import (
 	"go/ast"
 	"go/token"
 	"strings"
+	"sync"
 
 	"github.com/mgechev/revive/lint"
 )
@@ -14,10 +15,12 @@ import (
 // This has a notable false positive in that a package comment
 // could rightfully appear in a different file of the same package,
 // but that's not easy to fix since this linter is file-oriented.
-type PackageCommentsRule struct{}
+type PackageCommentsRule struct {
+	checkPackageCommentCache sync.Map
+}
 
 // Apply applies the rule to given file.
-func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+func (r *PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 	var failures []lint.Failure
 
 	if file.IsTest() {
@@ -29,7 +32,7 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
 	}
 
 	fileAst := file.AST
-	w := &lintPackageComments{fileAst, file, onFailure}
+	w := &lintPackageComments{fileAst, file, onFailure, r}
 	ast.Walk(w, fileAst)
 	return failures
 }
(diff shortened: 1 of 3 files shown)


rule/package-comments.go
```diff
@@ -31,9 +31,48 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
 	fileAst := file.AST
 	w := &lintPackageComments{fileAst, file, onFailure}
 	ast.Walk(w, fileAst)
+	failures = append(failures, checkPackageComment(file.Pkg)...)
 	return failures
 }
 
+var checkPackageCommentCache = map[*lint.Package]struct{}{}
```
chavacava, line 38: warn: the same rule is concurrently applied to multiple files, then accessing this map might produce race conditions

rule/package-comments.go
```diff
@@ -31,9 +31,48 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
 	fileAst := file.AST
 	w := &lintPackageComments{fileAst, file, onFailure}
 	ast.Walk(w, fileAst)
+	failures = append(failures, checkPackageComment(file.Pkg)...)
```
chavacava, line 34: the previous line is the original "walk" (not a real one in fact) of the file AST to check for failures. How that analysis relates with the one performed by ` checkPackageComment` ? They are complementary? Or the new check totally replaces the original one?

rule/package-comments.go
```diff
@@ -31,9 +31,48 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
...
+var checkPackageCommentCache = map[*lint.Package]struct{}{}
+
+func checkPackageComment(p *lint.Package) []lint.Failure {
+	// deduplicate warnings in package
+	if _, ok := checkPackageCommentCache[p]; ok {
+		return nil
+	}
+	checkPackageCommentCache[p] = struct{}{}
+	var docFile *ast.File
+	var firstFile *ast.File
```
chavacava, line 47: what is the purpose of these `first*` variables? _first_ in which sense?

rule/package-comments.go
```diff
@@ -31,9 +31,48 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
...
+	// prefer warning on doc.go
+	if docFile == nil {
+		docFile = firstFile
+	}
+	if docFile != nil {
+		return []lint.Failure{{
+			Category:   "comments",
+			Node:       docFile,
+			Confidence: 1,
+			Failure:    "should have a package comment, unless it's in another file for this package",
```
chavacava, line 70: the failure's message is not very _confident_: it can be interpreted as "there is a problem **unless** there is not" 
Creating a failure with confidence = 1 and a not so confident message seems to be somewhat contradictory

rule/package-comments.go
okhowang: I though there is not concurrency.
replaced with sync.Map

rule/package-comments.go
okhowang: maybe we should do checking in `Visit`

rule/package-comments.go
okhowang: for a package with many source file, and all of them have no package doc.
I think we should only raise one failure to warning lacking of pacakge doc
rather than warning on every file. (so there is checkPackageCommentCache)
so the problem is warning on which file.
currently I perfer warning on `doc.go` if it is exists.
else warning on then `firstFile` which is the first file in lexicographic order.

rule/package-comments.go
okhowang: maybe we can adjust message to `should have a package comment`.
the implementation should be confidence now.

rule/package-comments.go
tie: @okhowang, can we also add a special case for files named after the package in addition to `doc.go`? E.g. it’s common to have package `foo` docs in `foo.go`.

okhowang: > @okhowang, can we also add a special case for files named after the package in addition to `doc.go`? E.g. it’s common to have package `foo` docs in `foo.go`.

of course

rule/package-comments.go
```diff
@@ -34,6 +35,43 @@ func (*PackageCommentsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Fail
 	return failures
 }
 
+var checkPackageCommentCache sync.Map
```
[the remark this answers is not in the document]
okhowang: > Is there any reason this can’t be defined as a field in PackageCommentsRule with a pointer field in lintPackageComments?

it sound more reasonable.

chavacava (approved):

chavacava:
Hi @okhowang thanks for the PR updates.
In its current state, the PR do not pass rule's tests

okhowang:
@chavacava just because outdate matching pattern. updated

PR #703, opened 2022-07-05
Allow to customize user functions in rule `error-strings`

### LINKED ISSUE #697: error-strings function list should be user configurable

Our code-base has error strings that are not being checked by `error-strings`.

Currently, `error-strings` uses a [hardcoded list](https://github.com/mgechev/revive/blob/04728cf0de86ab99fb3327a1ea48f7d51a411352/rule/error-strings.go#L20-L32) of functions that it checks. We use an internal package for generating errors that is not covered by the list. It would be nice there was a configuration option to add additional functions.

Potentially, we could write our own Ruleguard rule or use another linting package. We could also rename our internal package to `errors`, which I think would end up working given the current error functions we have and my understanding of how `error-strings` works.

git-hulk:
**Please, describe in details what's your motivation for this PR**
As mentioned in #697, we may use other error packages instead of builtin or pkg
and also want to lint those error messages with `revive`, so it will be good to
allow to customize the function name.

### CHECKLIST
- [x] Did you add tests?
- [ ] Does the Travis build passes?

Closes #697

### DIFF
--- a/README.md
+++ b/README.md
@@ -350,6 +350,8 @@ enableAllRules = true
     Arguments = [7]
 [rule.function-result-limit]
     Arguments = [3]
+[rule.error-strings]
+    Arguments = ["mypackage.Error"]
 ```
 
 ### Default Configuration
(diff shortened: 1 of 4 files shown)


chavacava:
Thanks @git-hulk

PR #704, opened 2022-07-08
doc: add devlake to README usage

likyh:
add devlake to README usage


chavacava:
@likyh thanks for the PR.
Could you please add a comment to this PR pointing where `revive` is used in the devlake source code

likyh:
> @likyh thanks for the PR. Could you please add a comment to this PR pointing where `revive` is used in the devlake source code

I use revive in https://github.com/apache/incubator-devlake/pull/2453 and it will be merged after the committer of devlake reviewed.

likyh:
Now it merged. Revive is used in linter: https://github.com/apache/incubator-devlake/blob/main/.golangci.yaml#L60

PR #706, opened 2022-07-13
Check whether the tag name is duplicate or not

### LINKED ISSUE #705: Does it make sense to check the duplicate struct tag name?
Duplicate tag name should make no sense in the same struct and most of
cases should be mistaken, so I think we can check the struct tag.

Check whether has the duplicate struct tag key or not

No

**Additional context**
No

git-hulk:
### CHECKLIST ###

- [x] Please, describe in details what's your motivation for this PR
- [x] Did you add tests?

This closes #705

### DIFF
--- a/rule/struct-tag.go
+++ b/rule/struct-tag.go
@@ -73,16 +73,31 @@ func (w lintStructTagRule) checkTagNameIfNeed(tag *structtag.Tag) (string, bool)
 		return "", true
 	}
 
+	tagName := w.getTagName(tag)
 	// We concat the key and name as the mapping key here
 	// to allow the same tag name in different tag type.
-	key := tag.Key + ":" + tag.Name
+	key := tag.Key + ":" + tagName
 	if _, ok := w.usedTagName[key]; ok {
-		return fmt.Sprintf("duplicate tag name: '%s'", tag.Name), false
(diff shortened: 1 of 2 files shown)


git-hulk:
> Hi @git-hulk, thanks for the PR I've made minor refactoring and added some test cases. One of the test cases fails. When a field has a tag with two keys (for example json and yaml) and these keys have the same name... the rule generates a failure (but it should not) For example, in the following struct, the rule will identify `chartName` as duplicated.
> 
> ```go
> type HelmChartArgs struct {
> 	ChartName        string                 `json:"chartName,omitempty" yaml:"chartName,omitempty"`
> }
> ```

Got it, my bad. Will fix this case.

chavacava:
@git-hulk I've fixed a problem with the check on `protobuf` tags

chavacava:
@git-hulk thanks for the PR

If you are motivated ... there are still some tag formats to add to the rule :)

https://github.com/golang/go/wiki/Well-known-struct-tags#list-of-well-known-struct-tags

git-hulk:
> @git-hulk thanks for the PR
> 
> If you are motivated ... there are still some tag formats to add to the rule :)
> 
> https://github.com/golang/go/wiki/Well-known-struct-tags#list-of-well-known-struct-tags

Sure, I’ll be happy to do that

PR #719, opened 2022-07-23
Capture yet more bad defer / recover patterns

### LINKED ISSUE #718: Capture more bad / risky defer+recover patterns
While cleaning up some code in a repo I work on, and attempting to lock down our `defer` rules, the first round of fixing the problem led to code like this:
```go
func something() {
  defer logPanics()
}
func logPanics() {
  if r := recover() ...  // we want to enforce the lint that blocks this
}

// because this does not work
func helper() {
  logPanics()
}
func other() {
  defer helper()
}
```
turning into code like this:
```go
func something() {
  defer logPanics(recover())  // oops
}
```
... which trades one issue for another.

---

Since our underlying goal is to block all non-deferred recovers because it's error-prone (there are a few flaws in our codebase motivating this cleanup), and helper funcs are great for many reasons, the _intent_ was to do stuff like this:
```go
defer func() { logPanics(recover()) }()  // verbose but impossible to screw up helper funcs
```
Since `defer logPanics(recover())` is simpler looking, the eternal battle with working on a tired brain combined with the eternal seduction of seemingly-straightforward code led to the simpler-but-wrong one being selected.

Eh, it happens.  
But this is a _great_ signal for something worth locking down.

With reliable "must defer" and "must recover in deferred func, not arguments" lints, I think ^ both of those bad patterns can be made impossible.

Groxx:
`defer` + `recover()` is an eternal font of excitement.

This adds a new category under `defer`, "immediate-recover", which warns about cases like these:
```go
defer recover()
defer helper(recover())
```
because they do not do what they appear to do.  Neither will recover a panic (in normal use).

Tests and lint-tests include full demonstrations of ^ those, as well as some fairly exotic example where those _do_ work.
I have not ever seen these in practice, so for the moment I've set the confidence level to `1` - IMO both of these are very strong signals of flawed code, and worth running by default.

I'm fairly sure I'm missing some things from this PR, e.g. documentation.
I'll try to look tomorrow, but I'd be happy for any pointers :)  I'm happy to flesh it out.

Closes #718, which also contains motivation and other details.


test/defer_test.go
```diff
@@ -23,3 +23,82 @@ func TestDeferOthersDisabled(t *testing.T) {
 		Arguments: []interface{}{[]interface{}{"loop"}},
 	})
 }
+
+func TestDemonstrateIneffectiveDefers(t *testing.T) {
```
chavacava, line 27: Usually this kind of tests are not part of rules tests. You could add all these cases to `testdata/defer.go`

chavacava:
Hi @Groxx thanks for the PR, I've left one comment on it. 
If you are motivated enough, you could enrich the documentation of the rule at `[RULES_DESCRIPTIONS.md](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md)` with some detail on why those ways of using `defer` are not ok

test/defer_test.go
Groxx: yea, my goal was to provide _executable_ evidence that the patterns being detected behave the way the code / warnings describe.  these aren't linted, they're run as part of the test suite.

I'm entirely fine just removing them if you'd prefer to keep the tests more focused on running the linter or something, I wrote them half to re-prove it to myself :)

RULES_DESCRIPTIONS.md
```diff
@@ -233,12 +233,14 @@ _Configuration_: N/A
 ## defer
 
 _Description_: This rule warns on some common mistakes when using `defer` statement. It currently alerts on the following situations:
+
```
Groxx, line 236: I think this will fix the current rendering issue:
Screen Shot 2022-07-23 at 8 33 11 PM

chavacava:
Thanks @Groxx !

PR #756, opened 2022-10-07
fix #751 [rule.add-constant] add ignoreFuncs to exclude constants in …

### LINKED ISSUE #751: [rule.add-constant] add ''ignoreFuncs' to exclude constants in functions
I want to add 'ignoreFuncs' in configuration
to have a possibility to exclude constants in functions

```
[rule.add-constant]
  arguments = [{'ignoreFuncs="println,os.Chmod"}]
```

I want to create PR with changes, which add this feature

rawen17:
Closes #751

### DIFF
--- a/rule/add-constant.go
+++ b/rule/add-constant.go
@@ -3,6 +3,7 @@ package rule
 import (
 	"fmt"
 	"go/ast"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -32,8 +33,9 @@ func (wl whiteList) add(kind, list string) {
 
 // AddConstantRule lints unused params in functions.
 type AddConstantRule struct {
-	whiteList   whiteList
-	strLitLimit int
+	whiteList       whiteList
+	ignoreFunctions []*regexp.Regexp
+	strLitLimit     int
 	sync.Mutex
 }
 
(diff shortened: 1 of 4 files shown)


rule/add-constant.go
```diff
@@ -158,6 +214,18 @@ func (r *AddConstantRule) configure(arguments lint.Arguments) {
 						panic(fmt.Sprintf("Invalid argument to the add-constant rule, expecting string representation of an integer. Got '%v'", v))
 					}
 					r.strLitLimit = limit
+				case "ignoreFuncs":
+					excludes, ok := v.(string)
+					if !ok {
+						panic(fmt.Sprintf("Invalid argument to the add-constant rule, string expected. Got '%v' (%T)", v, v))
+					}
+
+					for _, exclude := range strings.Split(excludes, ",") {
```
chavacava, line 223: trimming blanks can help in avoiding subtle config problems.
For example configuring the rule with 
```go
"os.*,.Println,ignoredFunc"
```
will not behave as if it is configured with
```go
"os.*, .Println,ignoredFunc"
```

chavacava (changes requested):
I've left some comments
Please fill the PR description by using the PR description template

In general I think regexps are too powerful (or complicated?) to be used without the risk of ignoring things you don't know you are ignoring. For example, the regexp `.Println` used in the tests will not ignore just "methods named `Println` of any package" but **also** any method or function named such as `Println` is a suffix of the name (`myPrintln`, `ThisIsAnotherPrintln`, `dPrintlln`, ...)

chavacava:
My point is that RE are very error prone (at the point that your own test case `.Println` was wrong)
I'm not against the use of RE and I agree that providing clear doc will help in avoiding subtle errors; for example in the rule documentation it should be interesting to highlight that `fmt.Println` in the given example, will make `revive` ignore `fmt.Println`, `fmtXPrintln`, `fmt1Println`, an so on.

chavacava (approved):

chavacava:
> It was my fault, i had to write regexp `os\\.*,fmt\\.Println,make` in documentation, but i wrote `os.*,fmt.Println,make ` I will fix it

@rawen17 thanks for the PR, I've already merged it. 
Sorry @xobotyi I did not notice that I was answering to you (and not to the author of the PR)

rawen17:
@chavacava Ok, thanks, i will create new PR for this changes

PR #757, opened 2022-10-07
fix #754 [rule.unhandled-error] change arguments to regexp

### LINKED ISSUE #754: [rule.unhandled-error] change 'arguments' (ignored function names) in configuration to regexp
I want to change 'arguments' in configuration to have a possibility to add regexp function names

Example:
```
[unhandled-error]
  arguments =[".Chdir", "unhandledError1", "net.*""]
```

I want to create PR with changes, which add this feature

rawen17:
**Issue**
https://github.com/mgechev/revive/issues/754

**Reason to PR**
Change 'arguments' in configuration to have a possibility to add regexp function names
configuration arguments - Ignoring function names

**Example**
```
[unhandled-error]
  arguments =["os\.(Create|WriteFile|Chmod)", "fmt\.Print", "myFunction", "net\..*]
```


chavacava (changes requested):
Hi, thanks for  the PR.
I've left some comments.
Please provide a description of the the PR following the guidelines of the description template

chavacava:
Replacing plain function names to regexps in the conf breaks backward compatibility (i.e. old revive versions will behave differently than the new version even with the same configuration)
The test cases I've added show the problem.

xobotyi:
Welp, there are major versions for that🙃

As an example where it would be useful:
- `fmt\.Print.*`
- `github.com/gin-gonic/gin\.Error\.*`

chavacava:
> Welp, there are major versions for that🙃

The question is not if there are means to alert about backward compat breaks (semver for example) but "does the new functionality benefits are greater than the cost of breaking backward compatibility?" I do not have an answer.

(Please also notice that in this particular case, backward compatibility will break **silently**: old configurations will be valid but their semantic different)

> As an example where it would be useful:
> 
> * `fmt\.Print.*`
> * `github.com/gin-gonic/gin\.Error\.*`

Nice to see this time RE examples have `\`

chavacava (commented):
LGTM

@rawen17 please add a description for the PR

@mgechev this PR modifies the semantics of this rule's configuration, now the arguments are interpreted as regexps. This might silently "break" old configurations, for example an argument "myFunc" is currently interpreted as "_ignore calls to myFunc_" while this PR interprets it as "_ignore calls to all functions matching the regex 'myFunc'_" then, for example, calls to function `myFunc2` will be not ignored by the current implementation but ignored by the implementation.in this PR

wojtekxtx:
Disfaced FuckFuckMoz() fuckers

!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!

xobotyi:
@chavacava have an idea of making regexes to strictly match full string, thus there wont be any silent changes, previous confgurations wont work (since patterns has to contain package name now) and authors will be forced to go read the changelog.

chavacava (approved):

mgechev:
>@chavacava have an idea of making regexes to strictly match full string, thus there wont be any silent exclusions, previous confgurations wont work at all (since patterns has to contain package name now) and users will be forced to go read the changelog.

This sounds good to me. I'd rather break the config than silently break. Can we do this and I'll release next major?

xobotyi:
@rawen17 implement this proposal please.
rather than silently add symbols of line start and end - check match length against source string

mgechev:
I'll let @chavacava merge and I can cut a release.

PR #764, opened 2022-10-22
Speedup branch comparison in `identical-branches` rule

chavacava:
This PR optimizes branch comparison by avoiding unnecessary calls to the expensive `gofmt`
That is done by checking the number of statements on each branch before calling `gofmt`. If branches have not the same number of statements then they are necessarily not identical.


mgechev (approved):

mgechev:
Nice catch!

PR #765, opened 2022-10-22
Allows inversing the semantics of `string-format` rule configurations

### LINKED ISSUE #762: string-format: Missing possibility to consider matches as linting errors
Since Go (or at least the RE2 parsing engine) does not support negative lookaheads in regular expressions, it's hard for me to exclude certain strings in the string format.

Right now, I want to check for the things inside the `fmt.Errorf` call:

1. Does not contain phrases like "could not", "error" or similar.
2. Verbs are using the present tense instead of the continuous tense (e.g. "do" instead of "doing").

Both solutions are fairly easy to implement with regular expressions.

1. `/(could not|error|failed)/`
2. `/(\w+ing)/`

Examples:
- valid: `create API client: %w`
- invalid: `creating API client failed: %w:`

Unfortunately, the current revive rule asserts: If the pattern matches, this is valid code.
But I need: If the pattern matches, this is a linting error.

There should be a way of defining expression rules that should not appear in certain strings. So, if the pattern matches, this is considered a failure.

"Invert" the above regular expressions. In my opinion that's way too complicated for such a simple case.

chavacava:
Adds a syntactic mean to _negate_ the regular expressions in the configuration to pass from a "accept if match" to "fail if match" mode.

Now, using a `!` the configuration meaning is inverted

with "/^[A-Z].*$/" the rule will accept strings starting with capital letters
with "!/^[A-Z].*$/" the rule will a fail on strings starting with capital letters

Closes #762

### DIFF
--- a/rule/string-format.go
+++ b/rule/string-format.go
@@ -151,7 +151,7 @@ func (w lintStringFormatRule) parseArgument(argument interface{}, ruleNum int) (
 	negated = rule[1][0] == '!'
 	offset := 1
 	if negated {
-		offset += 1
+		offset++
 	}
 	regex, err := regexp.Compile(rule[1][offset : len(rule[1])-1])
 	if err != nil {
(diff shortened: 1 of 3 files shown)


rule/string-format.go
```diff
@@ -261,21 +268,40 @@ func (r *stringFormatSubrule) Apply(call *ast.CallExpr) {
...
-		return
-	}
-	var failure string
-	if len(r.errorMessage) > 0 {
-		failure = r.errorMessage
-	} else {
-		failure = fmt.Sprintf("string literal doesn't match user defined regex /%s/", r.regexp.String())
+	switch r.negated {
+	case false:
+		// Fail if the string doesn't match the user's regex
```
mgechev, line 273:
```suggestion
		// Fail if the string does NOT match the user's regex
```

mgechev: To make it easier to read :)

mgechev (approved):
LGTM. Left few nits, feel free to merge.

PR #815, opened 2023-04-28
Prevent empty-block when looping over call expr

### LINKED ISSUE #810: False positive on `for` loop
**Describe the bug**

`for` loops where the condition has side-effects and the block is empty can be valid.

**To Reproduce**

```
package main

import "fmt"

func main() {
	next := 0
	iter := func(v *int) bool {
		*v = next
		next++
		fmt.Println(*v)
		return next < 10
	}

	z := 0
	for iter(&z) {
	}
}
```

```
$ revive .
main.go:1:1: should have a package comment
main.go:15:15: this block is empty, you can remove it
```

Such code may be present in locations where you have an iterator interface and want to skip over rest-of the entries, but still read them.

**Expected behavior**

No warning.

 - OS: Windows 11
 - go version go1.20.1 windows/amd64

avorima:
Fix false positive for common use-case of looping over a function with side-effects mentioned in https://github.com/mgechev/revive/issues/810.

closes #810

### DIFF
--- a/rule/empty-block.go
+++ b/rule/empty-block.go
@@ -43,6 +43,13 @@ func (w lintEmptyBlock) Visit(node ast.Node) ast.Visitor {
 	case *ast.SelectStmt:
 		w.ignore[n.Body] = true
 		return w
+	case *ast.ForStmt:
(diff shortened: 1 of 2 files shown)


chavacava:
Thanks @avorima!

Issue #816, opened 2023-05-01
Idea: minimum block size for indent-error-flow/early-return


mdelah:
**Is your feature request related to a problem? Please describe.**
I find the `indent-error-flow` and `early-return` rules quite useful for finding large chunks of code that can lose an indentation level, but I'd prefer they didn't fire if the block in question was small (e.g. 1-2 lines), _especially_ if it requires moving an `if`-initializer to a separate line. I'm talking in particular about constructions of the form:

```
if foo, ok := Foo(); ok {
  foo.Bar()
} else {
  panic("!")
}
```
and
```
if foo, ok := Foo(); !ok {
  panic("!")
} else {
  foo.Bar()
}
```

The rules suggest reducing the indentation of the line `foo.Bar()` by one level by doing:

```
foo, ok := Foo()
if !ok {
  panic("!")
}
foo.Bar()
```

In terms of readability it does not seem like a clear win to me. The overall line count is unchanged (at 5), we increase the scope of the `foo` and `ok` variables in order to do it, and we only end up deleting a single tab character. Constructions like these are far more common (at least in my code base) than those where the `indent-error-flow` and `early-return` suggestions are helpful to me.

**Describe the solution you'd like**
I was wondering if the maintainers would consider configuration of some form to suppress such cases, to improve the signal-to-noise ratio. For example, a minimum count of lines that will be unintended to justify splitting an `if`-initializer onto a separate line. I'm happy to the supply code if desired.

**Describe alternatives you've considered**
- Could just bake in a reasonable minimum block size (say 2 or 3), and not have any extra configuration.
- Could have an option to simply never make such suggestions if there's an `if`-initializer.
- Could have separate block size thresholds for the case when there is an `if`-initializer and when there isn't.

**Additional context**
N/A

chavacava:
Hi @mdelah, thanks for the proposal.
I think we could: 
1. add an option to do not suggest the refactoring when there is in if-initializer, and 
2. (in absence of this option) let the rule calculate itself if the potential refactoring provides a decrease in lines count and suggest the modification only if it is the case.

What do you think?

mdelah:
Hi @chavacava, thanks for taking a look.

[...]

PR #821, opened 2023-05-08
refactor: extract shared code for linting if-else chains

mdelah:
The rules `early-return`, `indent-error-flow` and `superfluous-else` have a similar structure. This moves the common logic for classifying if-else chains to a common package.

A few side benefits:
- "early-return" now handles os.Exit/log.Panicf/etc
- "superfluous-else" now handles (builtin) panic
- "superfluous-else" and "indent-error-flow" now handle if/else chains with 2+ "if" branches

Motivation: preliminary refactor to make [#816](https://github.com/mgechev/revive/issues/816) easier to implement. Tidying up code I added to `early-return` rule previously.

I added tests wherever I noticed a change in behaviour (noted above).

I added the new package at `internal/ifelse`. If there's a better home, let me know.


internal/ifelse/kind.go
```diff
@@ -0,0 +1,79 @@
+package ifelse
+
+// Kind is a classifier for if-else branches. It says whether the branch is empty,
+// and whether the branch ends with a statement that can deviate control flow.
+type Kind int
```
chavacava, line 5: Rename to `BranchKind` ?

chavacava (changes requested):
Hi @mdelah  thanks for the PR! 
It LGTM, just need to fix linting errors (mainly public identifiers without doc) before merging.
Thanks again.

PR #827, opened 2023-05-15
Fix prefix trimming in struct-tag

### LINKED ISSUE #813: getTagName function uses strings.TrimLeft instead of TrimPrefix
**Describe the bug**

In the code below the "name=" string is a cutset (a set of characters) to be removed, and not a prefix.
https://github.com/mgechev/revive/blob/b03e54f617491984b41f0411b69326f94d03a76f/rule/struct-tag.go#L139-L140

`getTagName` function will return `o` instead of `ameno` for the this struct tag: `protobuf:"varint,name=ameno"`.

mgechev:
Fix #813


chavacava (approved):
LGTM

Issue #824, opened 2023-05-14
nested-structs false positives with "struct{}"


sbunce:
I found some instances of "struct{}" causing the nested-structs check to fail. Based on other issues it looks like the intention is to not include struct{} in this check.
```
type test struct {
  foo []chan struct{}
  bar map[string]struct{}
}
```

If this is working as intended feel free to resolve.

p.s. Revive works great! Found multiple bugs in my code.

mgechev:
Thank you!

Can't reproduce it in https://github.com/mgechev/revive/pull/826. Would you share another example which gives you the false positive?

sbunce:
I created a draft pull request which shows the issue. https://github.com/mgechev/revive/pull/828

I think it's an issue in the "nested-structs" rule.

Running "make test" shows the following.
```
    utils.go:120: Unexpected problem at nested-structs.go:50: no nested structs are allowed
    utils.go:120: Unexpected problem at nested-structs.go:51: no nested structs are allowed
--- FAIL: TestNestedStructs (0.00s)
```

chavacava:
Hi @sbunce, thanks for the PR with the counter example. I'll check for a solution.

chavacava:
@sbunce please let me know if the fix works for your cases. If that is the case, you can close the PR #828 
Thanks again!

sbunce:
Fix works. Thank you!

I found one more tiny one with generics.
```
type test struct {
	baz generic[struct{}]   // Must not match
}

type generic [T any] struct {}
```

The error points at the "baz" field.
```
=== RUN   TestNestedStructs
    utils.go:119: Unexpected problem at nested-structs.go:64: no nested structs are allowed
--- FAIL: TestNestedStructs (0.00s)
```

I added the example to my draft PR. https://github.com/mgechev/revive/pull/828/files

PR #828, opened 2023-05-16
nested-structs: show issue with "struct{}" detection

sbunce:
Background
=========
This pull request is to demonstrate issue in https://github.com/mgechev/revive/issues/824

It seems like the intent is to not fail on "struct{}".

### DIFF
--- a/testdata/nested-structs.go
+++ b/testdata/nested-structs.go
@@ -61,4 +61,7 @@ type mySetInterface interface {
 type test struct {
 	foo []chan struct{}     // Must not match
 	bar map[string]struct{} // Must not match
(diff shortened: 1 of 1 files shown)


mgechev (approved):

mgechev (changes requested):
Looks like this file contains broken syntax

sbunce:
The syntax is go generics.

I put it on go playground here.
https://go.dev/play/p/XSgqTLsjaS_N

I think the 1.17 test is failing because generics was introduced in go1.18.

chavacava:
Closing this PR. The example is already mentioned in #824 
Thanks!

PR #834, opened 2023-05-29
Fix typo in example documentation for early-return

meanguy:
The example for configuring `early-return` arguments mistakenly refers to the `exported` rule.


chavacava:
Thanks @meanguy !

PR #840, opened 2023-06-12
fix: support method call from structures for string-format

### LINKED ISSUE #839: string-format rule not working on method called when obj is inside an struct.
**Describe the bug**
The string-format rule search only for function calls and methods, but if the object is
inside a structure, the method call is not found by the rule.
This is a typical pattern when we want a generic Logger interface and different implementations. In this case `Logger` is a port and `MyLogger` would be an adapter from a project with hexagonal architecture. This implementation is injected in main to a service.

**To Reproduce**
1. Create go file with the following code:
```golang
package main

import "fmt"

type Logger interface {
	Error(msg string, keysAndValues ...any)
}
type MyLogger struct {
}

func (m *MyLogger) Error(msg string, _ ...any) {
	fmt.Println(msg)
}

type Service struct {
	logger Logger
}

func main() {
	s := Service{
		logger: &MyLogger{},
	}
	s.logger.Error("Test")
}
```

```shell
revive main.go
```

_config_
```toml
[rule.string-format]
  arguments = [
    ["logger.Error", "/^([^A-Z]|$)/",  "must not start with a capital letter"]]
```

**Expected behavior**
I spect it to fail with 

`./tests/main.go:23:17: must not start with a capital letter`

as the logger.Error call start with a capital letter

 - OS: Gentoo
 - Version of Go: 1.20

**Additional context**
MR is created with working solution (I don't know if the best)

rmarku:
This MR fix issue #839. Here the called object method is inside a structure (common pattern for dependency injection). Before the rule was not running for this rules.

The only thing is the function scope is tied to the structure name of the object, not to the object name.

This MR tries to fix this. A test was added to the string-format to evaluate this.

- [x] Did you add tests?
- [x] Does the Travis build passes?

Closes #839

### DIFF
--- a/rule/string-format.go
+++ b/rule/string-format.go
@@ -185,11 +185,13 @@ func (w lintStringFormatRule) Visit(node ast.Node) ast.Visitor {
 	if !ok {
 		return w
 	}
+
 	// Get the name of the call expression to check against rule scope
 	callName, ok := w.getCallName(call)
 	if !ok {
 		return w
 	}
+
 	for _, rule := range w.rules {
 		if rule.scope.funcName == callName {
 			rule.Apply(call)
@@ -217,7 +219,6 @@ func (lintStringFormatRule) getCallName(call *ast.CallExpr) (callName string, ok
 		if ok {
 			return recv.Sel.Name + "." + selector.Sel.Name, true
 		}
-		return "", false
 	}
 
 	return "", false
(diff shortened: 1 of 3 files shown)


chavacava:
Hi @rmarku thanks for the PR!

(abrazo de un santiagueño a un tucumano)

PR #799, opened 2023-03-03
Update default rule set to match recommendations

walles:
Before this change, Revive's defaults did not match the recommended configuration:

https://github.com/mgechev/revive#recommended-configuration

I found that to be confusing.

With this change in place, Revive now defaults to following its own recommendations.

Closes #800 

And as a (side?) note, I just [fixed actual bugs in my code by enabling unused-parameters warnings](https://github.com/walles/moar/commit/7bd5d3e0990e975ae9aa342dd237a3f8b9ebc58c#diff-f5d6cd5ff2d53fa27c0d76a18c3613153a8b43b492bc07e1ae12671c6e94ddba). Woho for unused parameters warnings!

Issue #537, opened 2021-06-29
if-return incorrectly treated as golint rule


rliebz:
**Describe the bug**
The README describes `if-return` as a `golint` rule, and therefore includes it in the set of defaults here and in `golangci-lint`. This is no longer true as of 2018.

- Related issue: https://github.com/golang/lint/issues/363
- Related PR: https://github.com/golang/lint/pull/388

Because of this discrepancy, there is an increased barrier to adoption of `revive` over `golint`.

**To Reproduce**
1. I updated revive `go install github.com/mgechev/revive@latest`
2. I ran revive on the following file with no `config.toml`:

```go
package example

import "errors"

// Foo ...
func Foo() error {
	if err := returnErr(); err != nil {
		return err
	}

	return nil
}

func returnErr() error {
	return errors.New("oh no")
}
```

And got the following results:

```shell
$ revive example.go
example.go:7:2: redundant if ...; err != nil check, just return error instead. 
```

With `golint`:

```shell
$ golint example.go
# no output
```

**Expected behavior**
I would expect the output from `revive` to be the same as `golint`, since this is what the README claims.

**Logs**
N/A

**Desktop (please complete the following information):**
 - OS: macOS 11.4
 - Version of Go: `go version go1.16.5 darwin/amd64`

**Additional context**
The same problem exists with `golangci-lint`:
```
$ golangci-lint run -E revive .
example.go:7:2: if-return: redundant if ...; err != nil check, just return error instead. (revive)
        if err := returnErr(); err != nil {
                return err
        }

$ golangci-lint run -E golint .
WARN [runner] The linter 'golint' is deprecated (since v1.41.0) due to: The repository of the linter has been archived by the owner.  Replaced by revive.

```

chavacava:
Hi @rliebz, thanks for filling  the issue. I expect to work on it in the coming days.

PR #843, opened 2023-06-23
Drop if-return from default ruleset

rliebz:
The `if-return` rule was originally a golint rule which was removed from their ruleset for being out of scope. Similarly, it was dropped from revive intentionally as a result of #537. More recently, it was reintroduced into the default ruleset as a result of #799 due to a discrepancy in documentation without a discussion of whether this rule in particular belonged as a part of that default rule set.

While it is no longer a goal of this project to align 100% with the golint defaults, I believe that this rule gives bad advice often enough that it should not be enabled by default.

For example, consider the following code:

```go
if err := func1(); err != nil {
	return err
}

if err := func2(); err != nil {
	return err
}

if err := func3(); err != nil {
	return err
}

return nil
```

The `if-return` rule considers this a violation of style, and instead suggests the following:

```go
if err := func1(); err != nil {
	return err
}

if err := func2(); err != nil {
	return err
}

return func3()
```

While this is more terse, it has a few shortcomings compared to the original. In particular, it means extending the size of the diff if changing the order of checks, adding logic after the call that currently happens to be last, or choosing to wrap the error. And in that last case, it can make it less obvious that there is an unwrapped error being propagated up the call stack.

This in practice has a very similar effect to disabling trailing commas; while it is not necessarily wrong as a style choice, I don't believe it warrants a position as part of the default ruleset here.

See-also: https://github.com/golang/lint/issues/363

### DIFF
--- a/README.md
+++ b/README.md
@@ -407,7 +407,6 @@ warningCode = 0
 [rule.error-strings]
 [rule.error-naming]
 [rule.exported]
-[rule.if-return]
 [rule.increment-decrement]
 [rule.var-naming]
 [rule.var-declaration]
--- a/config/config.go
+++ b/config/config.go
@@ -31,7 +31,6 @@ var defaultRules = []lint.Rule{
 	&rule.TimeNamingRule{},
 	&rule.ContextKeysType{},
 	&rule.ContextAsArgumentRule{},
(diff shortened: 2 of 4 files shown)


mgechev (approved):
Thanks for the PR!

Issue #845, opened 2023-07-01
another opensource project, the famous vector db milvus,who use Revive


kerneltravel:
refer to 
[milvus's Makefile](https://github.com/milvus-io/milvus/blob/4130b6dfc82a289b2899418f8a7c7dec9d589215/Makefile#L55C2-L55C2)
the famous vector db milvus is also using Revive for build.

chavacava:
Hi @kerneltravel thanks for pointing milvus as `revive` user.
Feel free to open a PR to add it to the README file.

PR #847, opened 2023-07-05
doc: add milvus to README usage (#845)

kerneltravel:
doc: add milvus to README usage (#845)


chavacava:
Thanks @kerneltravel !

Issue #851, opened 2023-07-17
Add `upperCaseConst` to `var-naming`


comdiv:
In our team naming policy we allow `SUCH_NAMES` for `constants`, (but not vars or functions), but now we cannot express it in revive

chavacava:
closed by #852

PR #852, opened 2023-07-17
imporve `var-naming`  - add upperCaseConst  option to allow UPPER_CASED constants #851

comdiv:
it's improvement for  #851 task

### DIFF
--- a/rule/var-naming.go
+++ b/rule/var-naming.go
@@ -13,11 +13,15 @@ import (
 
 var anyCapsRE = regexp.MustCompile(`[A-Z]`)
 
+// regexp for constant names like `SOME_CONST`, `SOME_CONST_2`, `X123_3` (#851)
+var upperCaseConstRE = regexp.MustCompile(`^[A-Z][A-Z\d]*(_[A-Z\d]+)*$`)
+
 // VarNamingRule lints given else constructs.
 type VarNamingRule struct {
-	configured bool
-	whitelist  []string
-	blacklist  []string
+	configured     bool
+	whitelist      []string
+	blacklist      []string
+	upperCaseConst bool // if true - allows to use UPPER_SOME_NAMES for constants
 	sync.Mutex
 }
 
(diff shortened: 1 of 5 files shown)


rule/var-naming.go
```diff
@@ -82,25 +103,31 @@ func (*VarNamingRule) Name() string {
...
 	}
 	for _, f := range fl.List {
 		for _, id := range f.Names {
-			check(id, thing, w)
+			w.check(id, thing)
 		}
 	}
 }
 
-func check(id *ast.Ident, thing string, w *lintNames) {
```
comdiv, line 96: It was too gready to "reserve" name `check` in package rule just for var-naming, moved lintNames as reviever to minimize exported named

chavacava:
Hi @comdiv, thanks for the PR! (and sorry for the late response)
I've made some little modifications to the code. 
Could you please provide a fix for the failing test case and update the description of the rule in RULES_DESCRIPTIONS.md ?

comdiv:
> Hi @comdiv, thanks for the PR! (and sorry for the late response) I've made some little modifications to the code. Could you please provide a fix for the failing test case and update the description of the rule in RULES_DESCRIPTIONS.md ?

Just not understand failing tests for now. Locally all work well.... Will try to fix

chavacava:
> Just not understand failing tests for now. Locally all work well.... Will try to fix

Test fails because I' ve added à test case for constant names without `_` (e.g. `VER`,  `MAX`, ...) 
You might need to pull locally the modifications I've made on your branch.

comdiv:
Ok, pulled.
But it's not after my changes, in legacy code we see:

```
if len(id.Name) >= 5 && allCapsRE.MatchString(id.Name) && strings.Contains(id.Name, "_") 
```
so your example of `VER` failed in old version of rule too.

What is better:

1. Change your VER example to VERSION one
2. Remove len limit for uppercase totally
3. Move len limit to parameter
4. Make len 1 or 2?

chavacava:
I see. The >= 5 condition comes from `golint` code base.
Please remove the test case I' ve added and keep te condition as is.
Thanks

comdiv:
Fix MD file too.

chavacava:
Thansk @comdiv !

Issue #853, opened 2023-07-29
rule for redundant import aliases


damif94:
**Is your feature request related to a problem? Please describe.**
In Goland IDE, I got errors for redundant import aliases, i.e., when the imported package name matches the alias names.
My coworkes at [draftea](https://www.draftea.com/) using VSCode don't get this errors, so it would be nice to make this compliant at the precommit/CI stage.

**Describe the solution you'd like**
A new rule `redundant-import-alias` that would warn upon this condition.

**Additional context**
The idea is to get this warning internalized into revive.

<img width="628" alt="Screen Shot 2023-07-29 at 17 40 02" src="https://github.com/mgechev/revive/assets/29461526/bd18a841-2300-4b81-9d84-8966bc6acd2d">

damif94:
@chavacava now that the PR is merged, we need a new release right?

chavacava:
The approach in `revive` is to use the master branch as the "latest version" of the tool.
Anyway, if for some reason a release is needed (for example if you are using `revive` through `golangci-lint`) @mgechev triggers a release  every ~4 months (or when critical bugs are fixed)

damif94:
I think a release would be convenient so we can use it with github actions
https://github.com/morphy2k/revive-action/blob/master/Dockerfile#L4

chavacava:
closed by #854

PR #854, opened 2023-07-29
feat: add rule for redundant import alias

damif94:
PR for issue https://github.com/mgechev/revive/issues/853

A new rule `redundant-import-alias` that would warn upon redundant import aliases, i.e., when the imported package name matches the alias names.


README.md
```diff
@@ -501,6 +501,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | [`use-any`](./RULES_DESCRIPTIONS.md#use-any)          |  n/a   |  Proposes to replace `interface{}` with its alias `any` |    no    |  no   |
 | [`datarace`](./RULES_DESCRIPTIONS.md#datarace)          |  n/a   |  Spots potential dataraces |    no    |  no   |
 | [`comment-spacings`](./RULES_DESCRIPTIONS.md#comment-spacings)          |  []string   |  Warns on malformed comments |    no    |  no   |
+| [`unused-import-alias`](./RULES_DESCRIPTIONS.md#unused-import-alias)          |  []string   |  Warns on import aliases matching the imported package name |    no    |  no   |
```
chavacava, line 504: the rule does no have arguments thus `[]string` should be replaced by n/a

rule/unused-import-alias.go
```diff
@@ -0,0 +1,52 @@
...
+
+	for _, imp := range file.AST.Imports {
+		if imp.Name == nil {
+			continue
+		}
+
+		if getImportPackageName(imp) == imp.Name.Name {
+			failures = append(failures, lint.Failure{
+				Confidence: 1,
+				Failure:    fmt.Sprintf("Import alias \"%s\" is not used", imp.Name.Name),
```
chavacava, line 26:
```suggestion
				Failure:    fmt.Sprintf("Import alias \"%s\" is redundant", imp.Name.Name),
```

rule/unused-import-alias.go
```diff
@@ -0,0 +1,52 @@
...
+			})
+		}
+	}
+
+	return failures
+}
+
+// Name returns the rule name.
+func (*UnusedImportAlias) Name() string {
+	return "unused-import-alias"
```
chavacava, line 38: What about `redundant-import-alias` ?

chavacava (commented):
Hi, thanks for the rule idea and its PR! 
I've left some comments, please let me know what do you think about them.

rule/unused-import-alias.go
damif94: Yes, i think its more accurate to call it "redundant"

damif94:
> Hi, thanks for the rule idea and its PR! I've left some comments, please let me know what do you think about them.

I agree with your suggestions, "redundant" seems like more appropriate. I already did the changes.

What's your view on making this rule as default? It seems like a pretty unquestionable linting rule, but on the other hand I got fear of retrocompatibility on upgrading versions and not sure on what's the project policies on upgrades.

chavacava:
> > Hi, thanks for the rule idea and its PR! I've left some comments, please let me know what do you think about them.
> 
> I agree with your suggestions, "redundant" seems like more appropriate. I already did the changes.
> 
> What's your view on making this rule as default? It seems like a pretty unquestionable linting rule, but on the other hand I got fear of retrocompatibility on upgrading versions and not sure on what's the project policies on upgrades.

Thanks for the modifications.
About making this rule default: historically, `revive` default rules were those of `golint`. I propose to let the new rule out of the defaults and actvite it through the config.

chavacava:
Thanks @damif94 !

PR #857, opened 2023-08-05
per-rule file exclude filters (#850)

### LINKED ISSUE #850: Glob/regex file excludes at rule level
Where are different policies for some files.
For example we never treat `add-constant` as warning in `*_test.go` files.
For now - we cannot express excludes at rule level - just at global level.

Suggested to add `exclude : []string` for configuration and `exclude: []*regexp.Regex` inside.

To distinguish between regexes and strings we could use `~` prefix for regexes:

```
# revive toml
# bettter to treat Exclude as common thing for all rules
[rule.add-constant]
    Exclude = [ "pkg/x/some.go" , "~_test.go$" ] # first is direct file name, second is regex
    Arguments = [
          {
             allowInts="-1,0,1,2",
             allowStrs="\"\"",
             allowFloats="0.0,1.0",
             ignoreFuncs="os\\.*,fmt\\.*"
          }
   ]
```

comdiv:
Rule-level file exclude filter feature implementation #850

Tryed to keep project at max compatibility level.

But where are some changes in core structures.

Filters are prechecked and initialized after reading from TOML, so it could cause error.
So `Initialize() error` method added to RuleConfig

Logically it should be applyed in `normalize...`  part of GetConfig, but just parse can return error, not normalize, so
added call of initialization to `parse` part of GetConfig

All tested.

Closes #850

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ vendor
 *.swp
 dist/
 *.log
+.vscode/settings.json
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ vendor
 *.swp
 dist/
 *.log
-.vscode/settings.json
--- a/README.md
+++ b/README.md
@@ -439,17 +439,20 @@ errorCode = 0
 warningCode = 0
 
 [rule.blank-imports]
-   exclude=["**/*.pb.go"]
+   Exclude=["**/*.pb.go"]
 [rule.context-as-argument]
-   exclude=["src/somepkg/*.go", "TEST"]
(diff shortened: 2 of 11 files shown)


chavacava:
Hi @comdiv, thanks for the PR !
(I've modified the semantics of the `""`exclusion pattern to _match nothing_ )

comdiv:
> Hi @comdiv, thanks for the PR ! (I've modified the semantics of the `""`exclusion pattern to _match nothing_ )

OK, it's well semantic too.

Issue #613, opened 2021-11-24
Allow use of `_xxx` with `unused-parameter` rule


Gusted:
**Is your feature request related to a problem? Please describe.**
[We're](https://github.com/go-gitea/gitea/pull/17803) enabling the `unused-parameter` rule on our repo to detect if there are possible bugs. A lot of parameters can be changed to `_`, but in [certain cases](https://github.com/go-gitea/gitea/pull/17803#discussion_r756222147) it will make the code more unreadable

**Describe the solution you'd like**
Add a option(or by default) that certain use-case are allowed, this can be limited to just the `_` prefix or a generic regex option. Using the `_` prefix is common in JS to still have the variable name but show that the variable is unused.

**Describe alternatives you've considered**
Adding the `//revive:disable...` comment, but that's better to avoided for each function that needs this.

**Additional context**
None

chavacava:
Hi @Gusted thanks for filling the issue.

>Using the _ prefix is common in JS to still have the variable name but show that the variable is unused.

It is common in JS but not in GO, to the point that `golint` (thus `revive`) will warn you with a: 
`don't use underscores in Go names`

Gusted:
I see, fair point.

Thank you for your response.

silverwind:
Is it out of question to have a rule configuration option for this? For example allow the user to define a regex pattern? The default pattern could be `^_$` and users would be free to define `^_` for example.

chavacava:
>Is it out of question to have a rule configuration option for this? 
Not at all. My intention with the previous comment was to highlight that prefixing identifiers with `_` is not idiomatic in GO

>It could allow the user to define a regex pattern
Yes, it is something that could be studied

twmb:
Prefixing with `_` is common for fields in structs when you want to define a struct but not have it exported. This is commonly used for forcing users of the struct to specify all fields, and to specifically pad the struct. This can be seen within the Go stdlib itself:
```
$ rg '^\s+_[a-zA-Z0-9]+\s+[a-zA-Z0-9]'
syscall/syscall_freebsd.go
144:		_p0          unsafe.Pointer

runtime/defs_plan9_amd64.go
31:	_type uint64

runtime/stack.go
1373:	_ptrdata int32 // ptrdata, or -ptrdata is GC prog is used

runtime/defs_linux_ppc64.go
184:	_pad0       int32

runtime/defs1_netbsd_amd64.go
97:	_signo  int32
98:	_code   int32
99:	_errno  int32
[...]

PR #858, opened 2023-08-07
extended [allowRegex] parameter for `unused-parameter` and `unused-receiver` rules

comdiv:
# Impl for #613 feature-request

Now can provide custom regexes for `unused-parameter` and `unused-receiver` to allow names matching some regex, no just `_` itself (does'nt break default behavior)

```toml
[rule.unused-parameter]
     Arguments = [ { allowRegex = "^_" } ]

[rule.unused-receiver]
     Arguments = [ { allowRegex = "^_" } ]
```

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -735,13 +735,36 @@ _Configuration_: N/A
 
 _Description_: This rule warns on unused parameters. Functions or methods with unused parameters can be a symptom of an unfinished refactoring or a bug.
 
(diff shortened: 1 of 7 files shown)


chavacava:
Hi @comdiv, thanks for the PR. Please open an issue before opening a PR.
Opening an issue will let us discuss on the subject and find the right way to cope with the request.

comdiv:
Sorry, i thought it was open early - it's fix for old issue #613 - what exatly i have to do?

PR #862, opened 2023-08-09
Allow import-blacklist to run against go test files

### LINKED ISSUE #860: Make import-blacklist configurable for test files
I'd like to enforce some import restrictions for our test files, but this is not possible due to the following:

https://github.com/mgechev/revive/blob/df39256ea77c2cfa014172d0c421bf300a112384/rule/imports-blacklist.go#L55

I'd like to either remove the if condition (which is probably undesirable because it might bring issues for the existing projects) or make it configurable (preserving the current behavior as default).

I could have just done that by catching regex values in my code, but that's not precise and I'd like to avoid it.

**Additional context**
I can provide a PR, if it's something that the project can accept.

denisvmedia:
This PR removes a condition that prevents import-blacklist from checking go test files (closes #860).

### DIFF
--- a/rule/imports-blacklist.go
+++ b/rule/imports-blacklist.go
@@ -52,10 +52,6 @@ func (r *ImportsBlacklistRule) Apply(file *lint.File, arguments lint.Arguments) 
 
 	var failures []lint.Failure
 
-	if file.IsTest() {
-		return failures // skip, test file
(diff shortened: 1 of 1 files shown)


chavacava:
@denisvmedia thanks for the PR!

PR #866, opened 2023-08-10
(var-naming) support private uppercase constants #865

### LINKED ISSUE #865: support private `_UPPER_CASED_CONST` with `var-naming` and `upperCaseConst=true`
I see new feature added last week #851 with allowing to use uppercase constants.

I think there should be also a **private** uppercase constant support for this feature.

```
const _SOME_PRIVATE_CONST = "hello, world"
```

fregin:
`var-naming` rule with `upperCaseConst=true` now allows to use private uppercase constants

Closes #865


chavacava:
Thanks for the PR @fregin !

Issue #872, opened 2023-08-18
use-any wont skip comments


damif94:
**Describe the bug**
It's unnecessary that use-any alerts on situations like:

```go
type Foo interface {
  // Baz()
}
```

chavacava:
Hi @damif94, thanks for creating the issue (and the corresponding PR)
Could you please elaborate on why the rule should not alert on those situations?
IMO the example you provide should be handled as what it is: an empty interface; therefore, the rule should trigger. (Moreover, the commented text is actual code and some linters will warn on it because it's not a good practice to keep commented code in the code base)

mgechev:
Hey @damif94 could you respond to @chavacava? His questions make sense, I just saw the discussion.

damif94:
Hey @chavacava .

For me the key thing is _intention_. 
It's difficult for me to think of a scenario where someone writing down
```go
type Foo interface {
  // Baz()
}
```
would want to actually mean `interface{}`.

> (Moreover, the commented text is actual code and some linters will warn on it because it's not a good practice to keep commented code in the code base)

This is a valid catch though; for me this argument means that the scenario being described is too rare and the fix is not worth the complexity being introduced into the rule implementation.

chavacava:
Hi @damif94 thanks for your comments. I agree on the use case (only commented code inside an interface declaration) being very exceptional and handling it does not justify increasing rule's runtime complexity.
I will revert the merge.
Thanks again (abrazo rioplatense)

damif94:
> Hi @damif94 thanks for your comments. I agree on the use case (only commented code inside an interface declaration) being very exceptional and handling it does not justify increasing rule's runtime complexity. I will revert the merge. Thanks again (abrazo rioplatense)

Un abrazo @chavacava 🧉

PR #873, opened 2023-08-18
fix: unnecesary alert for use-any when comments inside interface{}

damif94:
See https://github.com/mgechev/revive/issues/872


mgechev:
Thank you for the fix 🙌

PR #889, opened 2023-09-06
Add unchecked-type-assertion

### LINKED ISSUE #888: New Rule for dangerous type casts
A dynamic type cast like `x := something(*somethingElse[])` can cause a panic.  Because this can be extremely dangerous there should be a lint rule for it.

This rule should trigger if 
- a type ist dynamicly casted to a single value
- a type is dynamicly casted and the second response value (ok) is ignored
  - the first result can be checked for nil - so this might be nice to configure
- a dynamic type cast happens in a return statement

The alternative would be to always `panic.recover()` in bigger projects and just live with something as unsafe like
```golang
return ctx.Value("myCoolErrorChan").(chan err)
```

Dirk007:
Closes #888 

Unchecked type assertions results can (and will) result in an unexpected panic and should be a lint.
This new rule covers this with configurable non-default option to ignore `foo, _ := bar.(Baz)` - cases where the `ok` value is ignored. 

See tests.

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -62,6 +62,7 @@ List of all available rules.
   - [string-format](#string-format)
   - [superfluous-else](#superfluous-else)
   - [time-equal](#time-equal)
+  - [unchecked-type-assertion](#unchecked-type-assertion)
   - [time-naming](#time-naming)
   - [var-naming](#var-naming)
   - [var-declaration](#var-declaration)
@@ -170,8 +171,8 @@ Example:
 [rule.cognitive-complexity]
   arguments =[7]
 ```
-
 ## comment-spacings
+
 _Description_: Spots comments of the form:
 ```go
 //This is a malformed comment: no space between // and the start of the sentence
@@ -682,6 +683,26 @@ _Configuration_: N/A
 _Description_: Using unit-specific suffix like "Secs", "Mins", ... when naming variables of type `time.Duration` can be misleading, this rule highlights those cases.
 
 _Configuration_: N/A
+
+## unchecked-type-assertion
+
+_Description_: This rule checks whether a type assertion result is checked (the `ok` value), preventing unexpected `panic`s.
+
+_Configuration_: list of key-value-pair-map (`[]map[string]any`).
+
+- `acceptIgnoredAssertionResult` : (bool) default `false`, set it to `true` to accept ignored type assertion results like this:
+
+```go
+foo, _ := bar(.*Baz).
+//   ^
+```
+
+Example:
+
+```yaml
+[rule.unchecked-type-assertion]
+arguments = [{acceptIgnoredAssertionResult=true}]
+```
 
 ## var-naming
 
--- a/rule/unchecked-type-assertion.go
+++ b/rule/unchecked-type-assertion.go
@@ -96,8 +96,34 @@ func (w *lintUnchekedTypeAssertion) requireNoTypeAssert(expr ast.Expr) {
 	}
 }
 
+func (w *lintUnchekedTypeAssertion) handleIfStmt(n *ast.IfStmt) {
+	ifCondition, ok := n.Cond.(*ast.BinaryExpr)
+	if !ok {
+		return
+	}
+
+	w.requireNoTypeAssert(ifCondition.X)
+	w.requireNoTypeAssert(ifCondition.Y)
+}
+
+func (w *lintUnchekedTypeAssertion) requireBinaryExpressionWithoutTypeAssertion(expr ast.Expr) {
+	binaryExpr, ok := expr.(*ast.BinaryExpr)
+	if ok {
+		w.requireNoTypeAssert(binaryExpr.X)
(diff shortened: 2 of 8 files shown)


testdata/unchecked-type-assertion.go
```diff
@@ -0,0 +1,58 @@
...
+	switch n := foo.(type) {
+	case string:
+	case int:
+		//
+	}
+}
+
+func handleTypeSwitchReturn() {
+	// Should not be a lint
+	return foo.(type)
```
denisvmedia, line 57: Is it a valid construction?

testdata/unchecked-type-assertion.go
```diff
@@ -0,0 +1,58 @@
+package fixtures
```
denisvmedia, line 1: Can we have a test that will check this:

if foo.(int) == 1 {
    return
}

Is it supported?

testdata/unchecked-type-assertion.go
```diff
@@ -0,0 +1,58 @@
...
+func handleSkipped() {
+	r := foo.(int) // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
+}
+
+func handleReturn() int {
+	return foo.(int) // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
+}
+
+func handleSwitch() {
+	switch foo.(int) { // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
```
denisvmedia, line 23: Is this construction supported:

```go
switch foo.(int) == 1 {
    case true:
    case false:
}

?

Comment (denisvmedia) [inline testdata/unchecked-type-assertion.go:23]: Is this construction supported:

```go
switch {
    case foo.(int) == 1:
    case bar.(int) == 1:
}
```

testdata/unchecked-type-assertion.go
denisvmedia: Also, consider this example:

```go
	var value any
	value = 42 // int

	if v := value.(int); v == 42 {
		fmt.Printf("Value is an integer: %d\n", v)
	}
```

Will the linter check and fail here?

denisvmedia:
@chavacava I can, but I think this should be a part of the tests ;)

Dirk007:
> @chavacava I can, but I think this should be a part of the tests ;)

I totally agree! Unfortunately I am a bit short of time this week but will work on the suggested ideas ASAP.

testdata/unchecked-type-assertion.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+func handleTypeComparison() {
+	if foo.(int) == 1 { // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
+		return
+	}
+}
+
+func handleTypeComparisonReverse() {
+	if foo.(int) == 1 { // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
```
denisvmedia, line 62: I think, it's a small mistake here, it had to be:

```suggestion
	if 1 == foo.(int) { // MATCH /type cast result is unchecked in foo.(int) - type assertion will panic if not matched/
```

denisvmedia (approved):
It's all good from my POV (just a small issue in the test case). Good job 👍

denisvmedia:
Just one thing came to my mind now, probably this is still left out (sorry, can't check it now): `channel <- v.(int)`.

Dirk007:
> Just one thing came to my mind now, probably this is still left out (sorry, can't check it now): `channel <- v.(int)`.

good idea.. let me write a test 👍

RULES_DESCRIPTIONS.md
```diff
@@ -683,6 +684,26 @@ _Description_: Using unit-specific suffix like "Secs", "Mins", ... when naming v
...
+
+_Description_: This rule checks whether a type assertion result is checked (the `ok` value), preventing unexpected `panic`s.
+
+_Configuration_: list of key-value-pair-map (`[]map[string]any`).
+
+- `acceptIgnoredAssertionResult` : (bool) default `false`, set it to `true` to accept ignored type assertion results like this:
+
+```go
+foo, _ := bar(.*Baz).
+//   ^
```
chavacava, line 697: It could be interesting to list all the cases the rule is capable to detect (switch, channel, ...)

rule/unchecked-type-assertion.go
```diff
@@ -0,0 +1,196 @@
...
+type UncheckedTypeAssertionRule struct {
+	sync.Mutex
+	acceptIgnoredAssertionResult bool
+}
+
+func (u *UncheckedTypeAssertionRule) configure(arguments lint.Arguments) {
+	u.Lock()
+	defer u.Unlock()
+
+	if len(arguments) == 0 {
```
chavacava, line 26: We could add a `configure` field to the rule's struct to trace the rule's configuration was already processed and then quit ASAP once the rule was configured a first time.

```suggestion
	if len(arguments) == 0 || u.configured {
```

PR #895, opened 2023-09-16
feat: add support for enforce-map-style rule 

### LINKED ISSUE #892: Linter rule to check the consistency of make(map[type]type) vs empty map[type]type{}
There are two ways of declaring a map: either via `make` or via literal. Some projects might want to require to be consistent by choosing this or that way of map declaration.

I'd like to let the user choose between the styles (either `make` or via literal). `make(map[type]type, size)` will always be allowed, as well as `map[type]type{k1: v1, k2: v2}`. I'm also considering an option to check or ignore map descendant types (e.g. `type Map map[string]string`), on by default.

I didn't find any solution.

**Additional context**
IMO, consistent code, be it opinionated or not, is better than inconsistent. Also, I'm in favor for `make(map[type]type)` because it's then consistent with `make(map[type]type, size)`. Also, when a descendant type is declared and then used, it's not clear for the reader if it's a struct or a map, when we call it as `Map{}`.

P.S. I can implement such a linter rule, if it makes sense to the others.

denisvmedia:
closes #892

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -332,6 +332,22 @@ _Description_: Sometimes `gofmt` is not enough to enforce a common formatting of
 
 _Configuration_: N/A
 
+## enforce-map-style
+
+_Description_: This rule enforces consistent usage of `make(map[type]type)` or `map[type]type{}` for map initialization. It does not affect `make(map[type]type, size)` constructions.
+
+_Configuration_: (string) Specifies the enforced style for map initialization. The options are:
+- "any": No enforcement (default).
+- "make": Enforces the usage of `make(map[type]type)`.
+- "literal": Enforces the usage of `map[type]type{}`.
+
+Example:
+
+```toml
+[rule.enforce-map-style]
+  arguments = ["make"]
+```
+
 ## error-naming
 
 _Description_: By convention, for the sake of readability, variables of type `error` must be named with the prefix `err`.
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -334,7 +334,7 @@ _Configuration_: N/A
 
 ## enforce-map-style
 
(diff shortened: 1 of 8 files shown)


rule/enforce-map-style.go
```diff
@@ -0,0 +1,154 @@
...
+			[]enforceMapStyleType{
+				enforceMapStyleTypeAny,
+				enforceMapStyleTypeMake,
+				enforceMapStyleTypeLiteral,
+			},
+		)
+	}
+}
+
+// EnforceMapStyleRule implements a rule to enforce `make(map[type]type)` over `map[type]type{}`
```
denisvmedia, line 40:
```suggestion
// EnforceMapStyleRule implements a rule to enforce `make(map[type]type)` over `map[type]type{}`.
```

rule/enforce-map-style.go
```diff
@@ -0,0 +1,159 @@
...
+		}
+		return true
+	})
+
+	return failures
+}
+
+// Name returns the rule name.
+func (r *EnforceMapStyleRule) Name() string {
+	return "enforce-make-style"
```
denisvmedia, line 143:
```suggestion
	return "enforce-map-style"
```

RULES_DESCRIPTIONS.md
```diff
@@ -332,6 +332,22 @@ _Description_: Sometimes `gofmt` is not enough to enforce a common formatting of
 
 _Configuration_: N/A
 
+## enforce-map-style
+
+_Description_: This rule enforces consistent usage of `make(map[type]type)` or `map[type]type{}` for map initialization. It does not affect `make(map[type]type, size)` constructions.
```
denisvmedia, line 337:
```suggestion
_Description_: This rule enforces consistent usage of `make(map[type]type)` or `map[type]type{}` for map initialization. It does not affect `make(map[type]type, size)` constructions as well as `map[type]type{k1: v1}`.
```

denisvmedia:
I checked using `kubernetes` repo. It seems there is no consistency - it is noisy for both make and literal settings. So, I decided to have `any` as a default (basically, the linter won't work if `any` is set). This should prevent issues when people run `revive` with all linters enabled.

PR #900, opened 2023-09-22
feat: add support for enforce-slice-style rule

### LINKED ISSUE #898: Linter rule to enforce empty slice declaration/initialization style
Styles:

1. slice := make([]type)
2. slice := []type{}

There is also a way of having a non-initialized slice (which should be preferred over the empty slice w/o len and cap in most cases), but this is beyond the scope of this linter

I'd like to introduce a new linter that will enforce a particular style of these 3 to be used.

Things to take into account:

- slices with elements (i.e. `[]type{el1, el2, elN}`) - they should be ignored by the linter.
- slices with explicit len and (optionally) capacity: i.e. `make([]type, len)`, `make([]type, len, cap)` - they should be ignored by the linter.

I wasn't able to find any working alternative.

**Additional context**
This can be considered as a follow up feature for #892.

denisvmedia:
Closes #898.

### DIFF
--- a/testdata/enforce-slice-style-make.go
+++ b/testdata/enforce-slice-style-make.go
@@ -4,10 +4,11 @@ func somefn() {
 	m0 := make([]string, 10)
 	m1 := make([]string, 0, 10)
 	m2 := make([]string, 0)
-	m3 := []string{} // MATCH /use make([]type) instead of []type{} (or declare nil slice)/
-	m4 := []string{"v1", "v2"}
-	m5 := [8]string{}
-	m6 := [...]string{}
+	m3 := make([]string, 0, 0)
+	m4 := []string{} // MATCH /use make([]type) instead of []type{} (or declare nil slice)/
+	m5 := []string{"v1", "v2"}
+	m6 := [8]string{}
+	m7 := [...]string{}
 
 	_ = m0
 	_ = m1
@@ -16,6 +17,7 @@ func somefn() {
 	_ = m4
 	_ = m5
 	_ = m6
+	_ = m7
 }
 
 type Slice []string
@@ -24,14 +26,16 @@ func somefn2() {
 	m0 := make(Slice, 10)
 	m1 := make(Slice, 0, 10)
 	m2 := make(Slice, 0)
-	m3 := Slice{} // MATCH /use make([]type) instead of []type{} (or declare nil slice)/
-	m4 := Slice{"v1", "v2"}
+	m3 := make(Slice, 0, 0)
+	m4 := Slice{} // MATCH /use make([]type) instead of []type{} (or declare nil slice)/
+	m5 := Slice{"v1", "v2"}
 
 	_ = m0
 	_ = m1
 	_ = m2
 	_ = m3
 	_ = m4
+	_ = m5
 }
 
 type SliceSlice Slice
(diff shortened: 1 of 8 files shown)


testdata/enforce-slice-style-make.go
```diff
@@ -0,0 +1,62 @@
+package fixtures
+
+func somefn() {
+	m0 := make([]string, 10)
+	m1 := make([]string, 0, 10)
+	m2 := make([]string)
+	m3 := []string{} // MATCH /use make([]type) instead of []type{}/
```
chavacava, line 7: the match string must have the `(or declare nil slice)` suffix to pass tests

testdata/enforce-slice-style-make.go
```diff
@@ -0,0 +1,62 @@
...
+	_ = m6
+}
+
+type Slice []string
+
+func somefn2() {
+	m0 := make(Slice, 10)
+	m1 := make(Slice, 0, 10)
+	m2 := make(Slice)
+	m3 := Slice{} // MATCH /use make([]type) instead of []type{}/
```
chavacava, line 27: the match string must have the `(or declare nil slice)` suffix to pass tests

testdata/enforce-slice-style-make.go
```diff
@@ -0,0 +1,62 @@
...
+	_ = m4
+}
+
+type SliceSlice Slice
+
+func somefn3() {
+	m0 := make(SliceSlice, 10)
+	m1 := make(SliceSlice, 0, 10)
+	m2 := make(SliceSlice)
+	m3 := SliceSlice{} // MATCH /use make([]type) instead of []type{}/
```
chavacava, line 43: the match string must have the `(or declare nil slice)` suffix to pass tests

testdata/enforce-slice-style-make.go
```diff
@@ -0,0 +1,62 @@
...
+
+	_ = m0
+	_ = m1
+	_ = m2
+	_ = m3
+	_ = m4
+}
+
+func somefn4() {
+	m1 := [][]string{} // MATCH /use make([]type) instead of []type{}/
```
chavacava, line 54: the match string must have the `(or declare nil slice)` suffix to pass tests

testdata/enforce-slice-style-make.go
```diff
@@ -0,0 +1,62 @@
...
+	m1["el0"] = make([]string, 10)
+	m1["el1"] = make([]string, 0, 10)
+	m1["el2"] = make([]string)
+	m1["el3"] = []string{} // MATCH /use make([]type) instead of []type{}/
```
chavacava, line 58: the match string must have the `(or declare nil slice)` suffix to pass tests

chavacava (changes requested):
Hi @denisvmedia, thanks for the PR.
Test are not passing because one testdata file has expected issue messages not matching with actual messages.

One think you could do in testdata is to use the same set of slice declarations for all test cases (any, make, and literal) and at each testdata file just modify accordingly the `// MATCH /... /` annotations

denisvmedia:
@chavacava yeah, you a right, thanks. I added this suffix but forgot to fix one of the test cases. Now fixed.

chavacava:
Merged. Thank you @denisvmedia.

Issue #902, opened 2023-09-23
False positive in `unconditional-recursion` in presence of func literals


chavacava:
**Describe the bug**
The rule `unconditional-recursion` produces false positives when it finds calls to `foo` in function literals defined inside `foo`.
For example:

```go
func (c *connector) SpanStats(ctx context.Context, req *roachpb.SpanStatsRequest) (*roachpb.SpanStatsResponse, error) {
        // ... 
        err := c.withClient(ctx, func(ctx context.Context, c *client) error {
                stats, err := c.SpanStats(ctx, req)  // <- this call is wrongly flagged as unconditional recusive
               // ...
                return nil
        })
        return response, err
}
```

**To Reproduce**
1. I updated revive `go install github.com/mgechev/revive@latest`
2. Include the following code in `testdata/unconditional-recursion.go`
```go
func falsePositiveFuncLiteral() {
        _ = foo(func() {
                falsePositiveFuncLiteral()
        })
}
```
3. run `make test` (test will fail for `unconditional-recursion` rule)

**Expected behavior**
The new test pass.

mfederowicz:
@chavacava to be clear what is to fix content of unconditional-recursion.go, or writing new test which asserts other testcases than actual one, or maybe both?

chavacava:
The goal is to add the new test *and* fix the rule implementation to make tests pass.

mfederowicz:
@chavacava we waiting for something more in this issue, other tests or something?

chavacava:
Closed by #909

PR #909, opened 2023-09-26
fix: literal calls case

mfederowicz:
related with issue #902

fix is simple, we should add `case *ast.FuncLit` in rule: rule/unconditional-recursion.go , and return nil, then test pass. 

but at the end I dont know if I should add some additional code as in previous cases for ex: `w.updateFuncStatus(n.Body)` but if test pass it should be ok :P

### DIFF
--- a/rule/unconditional-recursion.go
+++ b/rule/unconditional-recursion.go
@@ -62,13 +62,11 @@ func (w lintUnconditionalRecursionRule) Visit(node ast.Node) ast.Visitor {
 		var rec *ast.Ident
 		switch {
 		case n.Recv == nil:
-			rec = nil
 		case n.Recv.NumFields() < 1 || len(n.Recv.List[0].Names) < 1:
-			rec = &ast.Ident{Name: "_"}
+			rec = nil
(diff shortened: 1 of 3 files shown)


chavacava:
Hi @mfederowicz, thanks for the PR.
I added a new test case:
```go
func nr902() {
	go func() {
		nr902() // MATCH /unconditional recursive call/
	})()
}
```

In this case, the function literal contains an actual unconditional recursive call and the rule should spot it.
To cope with this case, and still avoid false positives as the initial case of this issue, the rule should skip analyzing func literals used as *function call arguments*.

mfederowicz:
ok @chavacava with that last one test was a problem with checking line number in: `if p.Position.Start.Line != in.Line {` statement and signature used to identifate current function:

`w.currentFunc = &funcStatus{&funcDesc{rec, n.Name}, false}`

after some tweaks, all tests passes

of course you can propose other solution, maye my changes were not optimal :P

FYI: I made cleanup in commits (rebase on branch), so please git pull on your copy

PR #917, opened 2023-10-07
Fix trailing whitespaces

mfederowicz:
cleanup in few files containing trailing whitespace and incorrect end line

no tests needed


Makefile
```diff
@@ -16,4 +16,3 @@ build:
 
 test:
 	@go test -v -race ./...
-
```
chavacava, line 19: https://unix.stackexchange.com/questions/18743/whats-the-point-in-adding-a-new-line-to-the-end-of-a-file

mfederowicz: hmm interesting because it is result of end-of-file-fixer hook from https://github.com/pre-commit/pre-commit-hooks :)

this hook have description:

```
id: end-of-file-fixer
    name: fix end of files
    description: ensures that a file is either empty, or ends with one newline.
```
other files *.go in commits have removed last line, so we have a proof that hook not works well :(

mfederowicz: ok @chavacava i removed commit related to end of files, and leave related with trailing white spaces :)

PR #918, opened 2023-10-13
refactor: remove refs to io/ioutil

ginglis13:
io/ioutil has been deprecated since Go 1.16
https://pkg.go.dev/io/ioutil

Ignore unused error in `os.WriteFile` call - already set to be ignored in https://github.com/mgechev/revive/blob/8dbd58748ae8bc7d95a884fd8012c0c56c4efdff/test/unhandled-error_test.go#L18

### DIFF
--- a/testdata/unhandled-error-w-ignorelist.go
+++ b/testdata/unhandled-error-w-ignorelist.go
@@ -3,7 +3,6 @@ package fixtures
 import (
 	b "bytes"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"os"
 )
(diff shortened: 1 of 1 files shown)


testdata/unhandled-error-w-ignorelist.go
```diff
@@ -38,7 +37,7 @@ func testCase2() {
 	os.Chmod("test_file", os.ModeAppend)                          // ignore
 	os.WriteFile("test_file", []byte("some data"), os.ModeAppend) // ignore
 
-	ioutil.WriteFile("test_file", []byte("some data"), os.ModeAppend) // MATCH /Unhandled error in call to function ioutil.WriteFile/
+	os.WriteFile("test_file", []byte("some data"), os.ModeAppend) // ignore
```
chavacava, line 40:
```suggestion
	os.WriteFile("test_file", []byte("some data"), os.ModeAppend) // MATCH /Unhandled error in call to function os.WriteFile/
```

PR #927, opened 2023-10-29
Add denyRegex support for import-alias-naming rule

### LINKED ISSUE #924: Supporting multiple expressions for import-alias-naming
In my initial PR about adding [`import-alias-naming`](#881 ) we agreed to have only one expression. However, I'm now not able to easily solve the following limitation:

- allow all alphanumeric aliases with optional digits and have a length at least 1 symbol (i.e. `^[a-z][a-z0-9]*`)
- disallow `^v\d+$` aliases (e.g. `v1`)
- disallow `^v\d+alpha\d+$` aliases (e.g. `v1alpha1`)

In my PR I assumed that golang supports negative lookahead. Unfortunately, my assumption was wrong (as well as there's no support for positive lookahead, also in regex there's no general support for logical `AND`). Do you think, we can revise our decision and allow multiple rules that will be combined by logical `AND`? @chavacava 

I can submit a PR with an update (it will be backwards compatible).

denisvmedia:
This PR provides an extension to the existing `import-alias-naming` rule. While it preserves the backwards compatible configuration, with which it is possible to define an _allow_ regex pattern for import aliases, this PR brings a way to configure any or both pattern types: `allowRegex` and/or `denyRegex`. This brings two types of configuration:

1. The original one, allow pattern:
    ```toml
    # the original one, allow pattern
    [rule.import-alias-naming]
      arguments = ["^[a-z][a-z0-9]{0,}$"]
    ```
2. The new one, both allow and deny patterns:
    ```toml
    [rule.import-alias-naming]
      arguments = [ { allowRegex = "^[a-z][a-z0-9]{0,}$", denyRegex = '^v\d+$' } ]
    ```

Closes #924.

### DIFF
--- a/README.md
+++ b/README.md
@@ -534,7 +534,7 @@ List of all available rules. The rules ported from `golint` are left unchanged a
 | [`datarace`](./RULES_DESCRIPTIONS.md#datarace)          |  n/a   |  Spots potential dataraces |    no    |  no   |
 | [`comment-spacings`](./RULES_DESCRIPTIONS.md#comment-spacings)          |  []string   |  Warns on malformed comments |    no    |  no   |
 | [`redundant-import-alias`](./RULES_DESCRIPTIONS.md#redundant-import-alias)          |  n/a   |  Warns on import aliases matching the imported package name |    no    |  no   |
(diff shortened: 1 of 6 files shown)


chavacava:
Thank you @denisvmedia

Issue #424, opened 2020-05-20
extend testing machinery to allow more precise cheks


chavacava:
Rules are tested through a tiny mechanism where rules are executed on source files with annotations describing the expected failures.
For example:

```go
package fixtures

func foo(a, b, c, d int) {
	a = 1.0 // ignore
	b = "ignore"
	c = 2              // ignore
	println("lit", 12) // MATCH /avoid magic numbers like '12', create a named constant for it/
	if a == 12.50 {    // MATCH /avoid magic numbers like '12.50', create a named constant for it/
		if b == "lit" {
			c = "lit" // MATCH /string literal "lit" appears, at least, 3 times, create a named constant for it/
		}
		for i := 0; i < 1; i++ {
			println("lit")
		}
	}
}
``` 

The test machinery will check if the rule produced a failure at the line annotated with `// MATCH` and will also check that the message of the failure matches that of the annotation.

This approach works fine for almost all cases but it has its limitations: only the message attribute of the failure can be checked. We have no mean to check other attributes of the failure (for example, testing the fix for #416 needs to check that a failure has a certain confidence)

It could be interesting to extend the current testing mechanism to allow checking on any failure property

PR #929, opened 2023-11-01
extend parsing instructions to use json object instead of plain txt

mfederowicz:
related with #424 

now we can use instead of txt format:

`x = atomic.AddUint64(&x, 1)        // MATCH /direct assignment to atomic value/`

json format:

`x = atomic.AddUint64(&x, 1) // json:{"MATCH": "direct assignment to atomic value","Confidence": 1}`

of course this is only draft because I dont know how it should looks like logic to use all Failure properties, so I asume that for example we want to check if Confidence property have value equal to that setted up in json object, and if value is not equal then throw/log error otherwise everything is ok

this draft dont remove old MATH txt format, but add json feature, so we can use old format for some time :)

@chavacava what do you think about?

### DIFF
--- a/test/json-data-format_test.go
+++ b/test/json-data-format_test.go
@@ -1,0 +1,17 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/mgechev/revive/lint"
+	"github.com/mgechev/revive/rule"
+)
+
+func TestJsonDataFormat(t *testing.T) {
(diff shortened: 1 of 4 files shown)


chavacava:
Hi @mfederowicz, thank you for the PR (and sorry for the late response)
I'll will take a look and come back to you ASAP

chavacava:
Hi @mfederowicz, thanks for the feature. We will be able to test rules with more precision ;)

PR #939, opened 2023-11-16
[dot-imports] support allow list of packages

### LINKED ISSUE #923: Add configuration to the dot-import rule
The new release (included in golangci-lint v1.55.0) start checking the dot import also in test files. Many projects are using the [gomega](https://onsi.github.io/gomega) assertion package and the [ginkgo](https://onsi.github.io/ginkgo/) testing framework. The common pattern when using these packages is with dot import, because that makes (in this case) the code more readable.

Add an allow list of packages to the dot-import rule, so the rule will not apply for these packages.

e.g. in golangci-lint, it would be great to have a configuration similar to this:
```yaml
linters-settings:
  ...
  revive:
    ...
    rules:
      - name: dot-imports
        allow-packages:
        - "github.com/onsi/ginkgo/v2"
        - "github.com/onsi/gomega"
...
```

disable the rule.

nunnatsa:
Add the ability to allow list of packages to be dot imported.

Add a new don-imports configuration:
* `allowedPackages`: (list of strings) comma-separated list of allowed dot import packages

Example:

```toml
[rule.dot-imports]
  arguments = [{ allowedPackages = ["github.com/onsi/ginkgo/v2","github.com/onsi/gomega"] }]
```

Fixes #923

### DIFF
--- a/rule/dot-imports.go
+++ b/rule/dot-imports.go
@@ -1,16 +1,23 @@
 package rule
 
 import (
+	"fmt"
 	"go/ast"
+	"sync"
 
 	"github.com/mgechev/revive/lint"
 )
 
 // DotImportsRule lints given else constructs.
-type DotImportsRule struct{}
+type DotImportsRule struct {
+	sync.Mutex
+	allowedPackages allowPackages
+}
 
 // Apply applies the rule to given file.
-func (*DotImportsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+func (r *DotImportsRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
+	r.configure(arguments)
+
 	var failures []lint.Failure
 
 	fileAst := file.AST
@@ -20,6 +27,7 @@ func (*DotImportsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 		onFailure: func(failure lint.Failure) {
 			failures = append(failures, failure)
 		},
+		allowPackages: r.allowedPackages,
 	}
 
 	ast.Walk(walker, fileAst)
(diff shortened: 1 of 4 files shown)


rule/dot-imports.go
```diff
@@ -32,15 +40,47 @@ func (*DotImportsRule) Name() string {
 	return "dot-imports"
 }
 
+func (r *DotImportsRule) configure(arguments lint.Arguments) {
+	r.Lock()
+	defer r.Unlock()
+
+	if r.allowedPackages == nil {
+		r.allowedPackages = make(allowPackages)
+		if len(arguments) > 0 {
```
denisvmedia, line 49: To reduce cognitive complexity, I'd convert here and in other possible cases to guard conditions  (early return pattern).

testdata/import-dot.go
```diff
@@ -0,0 +1,18 @@
+// Test that dot imports are flagged.
+
+// Package pkg ...
+package pkg
+
+import (
+	. "context" // in allowedPackages (standard library => just the name without full path)
```
denisvmedia, line 7: Can we please have a non-aliased and aliased imports to make sure they are not affected.

test/dot_imports_test.go
```diff
@@ -0,0 +1,18 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/mgechev/revive/lint"
+	"github.com/mgechev/revive/rule"
+)
+
+func TestDotImports(t *testing.T) {
```
denisvmedia, line 10: Can we have a test case with a missing config? To make sure the change is backwards compatible.

testdata/import-dot.go
nunnatsa: Thanks for the review @denisvmedia. Fixed. PTAL.

chavacava:
Hi @nunnatsa thanks for the PR! (and sorry for the late response)

PR #941, opened 2023-11-24
variable to skip package name checks

gburanov:
It is a pity that two functionalities are merged together in the rule `rule.var-naming` - checking variable and package naming.

In our company we find variable naming check very useful, but we do not want to check for packages naming.

Therefore I propose to add option to skip package name checks

### DIFF
--- a/rule/var-naming.go
+++ b/rule/var-naming.go
@@ -18,10 +18,11 @@ var upperCaseConstRE = regexp.MustCompile(`^_?[A-Z][A-Z\d]*(_[A-Z\d]+)*$`)
 
 // VarNamingRule lints given else constructs.
 type VarNamingRule struct {
-	configured     bool
-	whitelist      []string
-	blacklist      []string
-	upperCaseConst bool // if true - allows to use UPPER_SOME_NAMES for constants
+	configured            bool
+	whitelist             []string
+	blacklist             []string
+	upperCaseConst        bool // if true - allows to use UPPER_SOME_NAMES for constants
+	skipPackageNameChecks bool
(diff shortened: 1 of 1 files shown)


rule/var-naming.go
```diff
@@ -78,22 +80,24 @@ func (r *VarNamingRule) Apply(file *lint.File, arguments lint.Arguments) []lint.
...
-	}
-	if anyCapsRE.MatchString(walker.fileAst.Name.Name) {
-		walker.onFailure(lint.Failure{
-			Failure:    fmt.Sprintf("don't use MixedCaps in package name; %s should be %s", walker.fileAst.Name.Name, strings.ToLower(walker.fileAst.Name.Name)),
-			Confidence: 1,
-			Node:       walker.fileAst.Name,
-			Category:   "naming",
-		})
+	if !r.skipPackageNameChecks {
+		// Package names need slightly different handling than other names.
```
chavacava, line 84: We could factor-out package checks into a method to get a simpler to read `Apply` method

chavacava:
HI @gburanov, thanks for the PR, I've left some comments in the code

chavacava:
@gburanov Please also:
1. update README.md and RULES_DESCRIPTIONS.md to document the new parameter.
2. Add testcases to check the parameter is working as expected

rule/var-naming.go
gburanov: done!

gburanov:
@chavacava  I made the changes you suggested

chavacava:
Thanks @gburanov !

PR #943, opened 2023-11-27
modifies-value-receiver: warn on slice or map

### LINKED ISSUE #942: modifies-value-receiver ignores modification to maps and slices
**Describe the bug**
The `modifies-value-receiver` rule is coded to ignore modifications where the object being modified is `map` or `[]foo`.
This is inappropriate: modifying those values is just as wrong as other types of values.
Assignments to a part of the map or slice (i.e. indexing operations) should still be ignored.

**To Reproduce**

2. I modified the test case `modifies-value-receiver.go`, assigning to `items`:
```
package fixtures

type data struct {
	num   int
	key   *string
	items map[string]bool
}

func (this data) vmethod() {
	this.items = make(map[string]bool) // MATCH /suspicious assignment to a by-value method receiver/
	this.items["vmethod"] = true
}
```

```shell
revive -config ./defaults.toml ./testdata/modifies-value-receiver.go 
```

```toml
# config file
```

**Expected behavior**
I expect a warning on the line assigning to `this.items`.

 - OS: Ubuntu 20.04
 - Version of Go 1.21.1

**Additional context**
The motivating case is here: https://github.com/prometheus/prometheus/pull/13040/files#diff-d9f058bd14a07ebec78d70aaa8d6e8e705bc09086e338b74b12666e7ac121605R1711

bboreham:
Assignments which index into the slice or map are already ignored on line 72, and those where the whole receiver is slice or map on line 50. We should not ignore assignments where the member variable is slice or map.

Closes #942

### DIFF
--- a/rule/modifies-value-receiver.go
+++ b/rule/modifies-value-receiver.go
@@ -79,10 +79,6 @@ func (w lintModifiesValRecRule) Visit(node ast.Node) ast.Visitor {
 						continue
 					}
 
-					if w.skipType(ast.Expr(e.Sel)) {
-						continue
-					}
-
 				case *ast.Ident: // receiver := ...
 					if e.Name != receiverName {
 						continue
--- a/rule/modifies-value-receiver.go
+++ b/rule/modifies-value-receiver.go
@@ -78,7 +78,6 @@ func (w lintModifiesValRecRule) Visit(node ast.Node) ast.Visitor {
 					if name == "" || name != receiverName {
 						continue
 					}
-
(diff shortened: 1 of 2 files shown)


rule/modifies-value-receiver.go
```diff
@@ -79,10 +79,6 @@ func (w lintModifiesValRecRule) Visit(node ast.Node) ast.Visitor {
 						continue
 					}
```
denisvmedia, line 81:
```suggestion
```

Issue #690, opened 2022-05-21
Deprecate the use of words Whitelist and Blacklist


dschveninger:
# Overview
The words whitelist and blacklist been used for years, but with current events we are questioning if we could some more inclusive terms.

## Question
I was wonder if you would be opposed to deprecating the use of the words blacklist and whitelist in this repository?

## Suggestion
I would like to get feed back on either addressing the whole repo

Whitelist https://github.com/mgechev/revive/search?q=whitelist
Blacklist https://github.com/mgechev/revive/search?q=blacklist

Or just address it with the rules naming and config words. The approach is to mirror the old config with a new config. The old config would echo out a deprecation message for a future version.

Config changes https://github.com/mgechev/revive/blob/a67ecdd7ba96272124152b280054d16368e774ac/config/config.go

if you are not opposed to the idea, I may have someone that can address this. I just wanted to get your feedback first before we would submit a PR.

chavacava:
Hi @dschveninger, thanks for filling the issue.

Changing all internal declarations and references to `whiteList` and `blackList` seems OK and without risk.
On the other side, changing rule names (like `import-blacklist`) and rule configurations should be done very carefully. 
I propose to go ahead with your suggestion in three steps:
1. refactoring internal use of `whiteList` and `blackList`
2. **add alternatives** to external uses of `whiteList` and `blackList` and inform of the deprecation
3. completely remove `whiteList` and `blackList` from the code base

Please let me know if you are OK with the approach.

mgechev:
@dschveninger do you want to move forward opening a PR? I'd love to make this change happen.

mfederowicz:
Hi @dschveninger any suggestions/alternatives to thouse words, that we can prepare some commits in this topic?

mfederowicz:
hey @chavacava any next steps/plans/suggestions with this issue?

chavacava:
Hi @mfederowicz if you plan to go ahead with the renaming, your PR are welcome. That said, the next phase of refactoring might imply breaking changes. Thus, before putting some effort on coding, please open a thread to discuss.

mfederowicz:
ok i think that we can close this issue, latest changes removed usage of Whitelist and Blacklist

Whitelist https://github.com/mgechev/revive/search?q=whitelist
Blacklist https://github.com/mgechev/revive/search?q=blacklist

as you can see there is only one usage of prase "imports-blacklist" it is only for fix/resolve breaking changes

PR #946, opened 2023-11-28
refactor: rename blacklist to blocklist and whitelist to allowlist

mfederowicz:
related to #690 

I thougt that rename Whitelist to Allowlist and Blacklist to Blocklist (all variable of course respect lower case and upper case)

I dont have any idea how to implement deprecation for old `&rule.ImportsBlacklistRule{}`, one idea is that to leave empty ImportsBlacklistRule and throw info about deprecation what you think @chavacava 

ps. i checked other repos in github with `&rule.ImportsBlacklistRule{}` and there are 75 repos matching, so we dont have any simple solution :(


chavacava:
Hi @mfederowicz thanks for the PR.
I propose to do these changes in two steps:
1.  Remove white and black list names from the internal code
2. Remove white and black list names from the external interfaces (config)

Use this PR for the step 1 (the easy one) and discuss how to go on with step 2

mfederowicz:
ok @chavacava i cleaned up code from my changes related with deprecatedRules list, now there are only changes related with removing white and black lists (of course ImportsBlacklistRule was reverted to old state from master)

now iam listening sugestions/ideas how to  remove black list from the external interfaces (config), white list luckly was used only in lists internaly (not in config or dedicated rules)

README.md
chavacava, line 1: this file should not be modified

RULES_DESCRIPTIONS.md
chavacava, line 1: this file should not be modified

rule/add-constant.go
```diff
@@ -53,7 +53,7 @@ func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lin
 		onFailure:       onFailure,
 		strLits:         make(map[string]int),
 		strLitLimit:     r.strLitLimit,
-		whiteLst:        r.whiteList,
+		allowLst:        r.allowList,
```
chavacava, line 56:
```suggestion
		allowList:        r.allowList,
```

rule/imports-blocklist.go
chavacava, line 1: Changing the rule name is a breaking change for users.
Let the original name.

rule/var-naming.go
```diff
@@ -35,11 +35,11 @@ func (r *VarNamingRule) configure(arguments lint.Arguments) {
 
 	r.configured = true
 	if len(arguments) >= 1 {
-		r.whitelist = getList(arguments[0], "whitelist")
+		r.allowlist = getList(arguments[0], "allowlist")
```
chavacava, line 38: can not change the name of the configuration key, it is a breaking change for users

test/import-blocklist_test.go
chavacava, line 1: rollback modifications in this file, the rule can not be renamed because in doing that we introduce a breaking change for users

chavacava (commented):
@mfederowicz I've left some comments in the PR.
Please keep in mind that changing the name of a rule is a breaking change (users might need to modify configurations ) As I proposed, this PR should include only internal naming changes.

rule/var-naming.go
mfederowicz: as i understand we can change name r.allowlist  but not getList(arguments[0], "whitelist") because whitelist is used as parameter to var-naming rule?

rule/imports-blocklist.go
mfederowicz: this is copy of old imports-blacklist.go so dont change nothing for user :P

rule/var-naming.go
chavacava: Exactly.
`r.allowlist` is an internal field, on the contrary, requesting the argument to be `"allowlist"` is forcing the (external) configuration to be changed.

chavacava:
Side note: changing the failure messages could be also considered as a breaking change because users might rely on failure messages thus changing them will force users to update their code.
Do not worry, I'm okay with the new messages you pushed, my note is just to highlight that managing "breaking changes" is a big deal :)

denisvmedia:
> my note is just to highlight that managing "breaking changes" is a big deal :)

We'll have to put some release notes on that, I think.

rule/var-naming.go
mfederowicz: ok arguments names reverted

mfederowicz:
@chavacava what is the status, any new sugestions?

chavacava:
> @chavacava what is the status, any new sugestions?

As I commented before, the PR (still) has renamed rules (the `imports-blacklist` rule renamed as `imports-blocklist`) 
Changing the rule's name is a breaking change and my understanding was we agreed on not introducing breaking changes in this first phase of refactoring.

mfederowicz:
> > @chavacava what is the status, any new sugestions?
> 
> As I commented before, the PR (still) has renamed rules (the `imports-blacklist` rule renamed as `imports-blocklist`) Changing the rule's name is a breaking change and my understanding was we agreed on not introducing breaking changes in this first phase of refactoring.

yes but imports-blocklist is new file not rename old one. I can revert testdata for test/import-blacklist_test.go and testdata/imports-blacklist-original.go but this dont have affect to old code :)

chavacava:
> yes but imports-blocklist is new file not rename old one. I can revert testdata for test/import-blacklist_test.go and testdata/imports-blacklist-original.go but this dont have affect to old code :)

Does that mean we will have two rules for the very same check?

mfederowicz:
ok @chavacava i mean that I created copy of blacklist rule. Now I reverted tests/testdata for blacklist, and add separated test/testdata for blocklist. In near future we should remove Blacklist rule and add some notes about breaking change. If you (user) use development version of package you must know that changes are part of your live :P

Issue #936, opened 2023-11-10
bug: `redundant-import-alias` rule check for import path instead of package name


candiduslynx:
**Describe the bug**

When checking `redundant-import-alias` the code will only look at the import path, whereas the proper way would've been to check the imported package name.
I've seen this with `import apiv2 "cloud.google.com/go/run/apiv2"` (the code is under `apiv2` directory, but the package name is `run`: https://pkg.go.dev/cloud.google.com/go/run/apiv2)

**To Reproduce**
1. Create a package in a dir named `redundant` with the following content (note that the package name must be different):
   ```go
   package relevant
   
   var Lovely = int(123)
   ```
3. Import it via `import redundant "path/to/redundant"`
4. Run check

**Expected behavior**
No issue is presented, as the alias is used to override the package name, not the path.

**Logs**
```
redundant-import-alias: Import alias "apiv2" is redundant (revive)
```
This is printed when importing:
```go
import apiv2 "cloud.google.com/go/run/apiv2"
```

chavacava:
Hi @candiduslynx, thank you for reporting the issue.

denisvmedia:
The only viable solution for that seems to be using

```
	import "golang.org/x/tools/go/packages"

	// ...

	// Load the package information
	pkgs, err := packages.Load(cfg, importPath)
	if err != nil {
		return "", fmt.Errorf("Error loading package: %w", err)
	}

	// Ensure that the package information is available
	if len(pkgs) == 0 || pkgs[0].Name == "" {
		return "", fmt.Errorf("Package information not found for import path: %s", importPath)
	}

	return pkgs[0].Name, nil // <-- here we'll return `run` for https://pkg.go.dev/cloud.google.com/go/run/apiv2
```

The problem of this solution is that package.Load uses `exec.Command` internally. And thus it's very slow on big code bases. For such cases I'd use `importas` linter, where you can strictly define an alias (or an absence of such). I'm not sure if `revive` has a built-in alternative.

candiduslynx:
Another possible solution would be to load only package name when traversing the vendored imports, but this will also require an additional effort:
* How the vendor directory is defined based on the place where linter was executed from?
* Are the dependencies required to be vendored now? If now, this will be a major change to simple code scanning

denisvmedia:
[...]

PR #950, opened 2023-12-21
fix redunant-import-alias rule - unused alias is also redundant

mfederowicz:
related with: #936 

I think we should change definition of redundant alias, because import like this:

import apiv2 "cloud.google.com/go/run/apiv2" is ok i one condition: when alias apiv2 is used min one time in code below import, otherwise is redundant because is not used

of course this rule checks only imports with aliases, imports without aliases are ignored in this rule :)

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -672,7 +672,8 @@ _Configuration_: N/A
 
 ## redundant-import-alias
 
-_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name. 
+_Description_: This rule warns on redundant import aliases. This happens when the alias 
+used on the import statement matches the imported package name.
 
 _Configuration_:
 
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -672,9 +672,18 @@ _Configuration_: N/A
 
 ## redundant-import-alias
 
-_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name.
+_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name. 
 
-_Configuration_: N/A
+_Configuration_:
+
+* `ignoreUsed` : (bool) ignore aliases used in code (useful when using popular sdk packages).
+
+Example:
+
+```toml
+[rule.redundant-import-alias]
+  arguments = [{ ignoreUsed = true}]
+```
 
 ## string-format
 
--- a/rule/redundant-import-alias.go
+++ b/rule/redundant-import-alias.go
@@ -3,6 +3,7 @@ package rule
 import (
 	"fmt"
 	"go/ast"
+	"strings"
 	"sync"
 
 	"github.com/mgechev/revive/lint"
@@ -59,35 +60,39 @@ func (r *RedundantImportAlias) checkRedundantAliases(node ast.Node) map[string]s
 			return true
 		}
 
-		if imp.Name != nil && imp.Path != nil {
-			aliasedPackages[imp.Name.Name] = "redundant"
+		if imp.Name != nil && imp.Path != nil && imp.Name.Name != "_" && getImportPackageName(imp) == imp.Name.Name {
+			aliasedPackages[imp.Name.Name] = "redundant"	
 		}
+
 		return true
 	})
 
-	if r.ignoreUsed {
-		// Second pass: remove one time used aliases
-		ast.Inspect(node, func(n ast.Node) bool {
-			sel, ok := n.(*ast.SelectorExpr)
-			if !ok {
-				return true
-			}
-
-			x, ok := sel.X.(*ast.Ident)
-			if !ok {
-				return true
-			}
-
-			// This alias is being used; it's not redundant
-			if _, exists := aliasedPackages[x.Name]; exists {
-				delete(aliasedPackages, x.Name)
-			}
-
-			return true
-		})
+	if !r.ignoreUsed {
+		return aliasedPackages
 	}
 
+	// Second pass: remove one-time used aliases
+	ast.Inspect(node, func(n ast.Node) bool {
+		sel, ok := n.(*ast.SelectorExpr)
+		if !ok {
+			return true
+		}
+
+		x, ok := sel.X.(*ast.Ident)
+		if !ok {
+			return true
+		}
+
+		// This alias is being used; it's not redundant
+		if _, exists := aliasedPackages[x.Name]; exists {
+			delete(aliasedPackages, x.Name)
+		}
+
+		return true
+	})
+
 	return aliasedPackages
+
 }
 
 func (r *RedundantImportAlias) configure(arguments lint.Arguments) {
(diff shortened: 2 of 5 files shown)


RULES_DESCRIPTIONS.md
```diff
@@ -672,9 +672,18 @@ _Configuration_: N/A
 
 ## redundant-import-alias
 
-_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name.
+_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name.
```
denisvmedia, line 675:
```suggestion
_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name.
```

rule/redundant-import-alias.go
```diff
@@ -38,15 +48,69 @@ func (*RedundantImportAlias) Name() string {
...
+			return true
+		}
+
+		if imp.Name != nil && imp.Path != nil {
+			aliasedPackages[imp.Name.Name] = "redundant"
+		}
+		return true
+	})
+
+	if r.ignoreUsed {
```
denisvmedia, line 68: I'd invert the condition to reduce the indentation.

rule/redundant-import-alias.go
```diff
@@ -38,15 +48,69 @@ func (*RedundantImportAlias) Name() string {
...
-	return strings.Trim(path[i+1:], strDelim)
+	return aliasedPackages
+}
+
+func (r *RedundantImportAlias) configure(arguments lint.Arguments) {
+	r.Lock()
+	defer r.Unlock()
+
+	r.ignoreUsed = defaultIgnoreUsed
+	if len(arguments) > 0 {
```
denisvmedia, line 98: nit: I'd invert the condition to apply early return and reduce the indentation.

testdata/redundant-import-alias.go
```diff
@@ -1,12 +1,24 @@
...
+	strings "strings" // MATCH /Import alias "strings" is redundant/
+
 	"crypto/md5"
-		"strings"
-		_ "crypto/md5"
-		str "strings"
-		strings "strings"  // MATCH /Import alias "strings" is redundant/
-		crypto "crypto/md5"
-		md5 "crypto/md5"  // MATCH /Import alias "md5" is redundant/
+	_ "crypto/md5" // MATCH /Import alias "_" is redundant/
```
denisvmedia, line 11: This is wrong. `_` is a special alias type for side effect loads. If your intention was to avoid duplicate imports, then this is a wrong rule to apply it here.

denisvmedia (changes requested):

testdata/redundant-import-alias.go
mfederowicz: I know this is wrong, but it is testdata file, in normal apps not used :)

RULES_DESCRIPTIONS.md
mfederowicz: done

testdata/redundant-import-alias.go
mfederowicz: ok removed wrong example :)

testdata/redundant-import-alias.go
```diff
@@ -1,12 +1,16 @@
 package fixtures
 
-import(
+import (
 	"crypto/md5"
-		"strings"
-		_ "crypto/md5"
```
denisvmedia, line 6: Can we please have this back in the test case?

RULES_DESCRIPTIONS.md
```diff
@@ -672,9 +672,19 @@ _Configuration_: N/A
...
 ## redundant-import-alias
 
-_Description_: This rule warns on redundant import aliases. This happens when the alias used on the import statement matches the imported package name.
+_Description_: This rule warns on redundant import aliases. This happens when the alias 
+used on the import statement matches the imported package name.
 
-_Configuration_: N/A
+_Configuration_:
+
+* `ignoreUsed` : (bool) ignore aliases used in code (useful when using popular sdk packages).
```
denisvmedia, line 680: Can you improve the wording. It's not really clear for me what exactly is to be ignored.

mfederowicz: import apiv2 "cloud.google.com/go/run/apiv2"
import v1 "https://pkg.go.dev/k8s.io/api/core/v1"

technically above aliases are correct if they are used min one time later in code. What is your sugestion to describe that flag?

rule/redundant-import-alias.go
mfederowicz: done

testdata/redundant-import-alias.go
mfederowicz: old testdata restored

RULES_DESCRIPTIONS.md
denisvmedia: The problem here is that if it's not used, the code won't compile.

```go
package main

import apiv2 "cloud.google.com/go/run/apiv2"
import v1 "k8s.io/api/core/v1"
import "fmt"

func main() {
    fmt.Println("dummy")
}
```

```bash
$ go build .
# poc
./main.go:3:8: "cloud.google.com/go/run/apiv2" imported and not used
./main.go:4:8: "k8s.io/api/core/v1" imported and not used
```

mfederowicz: no no no, your case is not that case, because this code is not used and make "noise" after compile

but in case mentioned in https://github.com/mgechev/revive/issues/936:

```
package main

import (
	"fmt"
	redundant "example.com/app/redundant"
)

func main() {
	fmt.Println(redundant.Lovely)
}

=$ ./app 
123
```

above code is working well, but linter shows warning:

```
=$ cat ~/revive-aliases.toml 
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.8
errorCode = 0
warningCode = 0

[rule.blank-imports]
[rule.dot-imports]
[rule.redundant-import-alias]
```

```
revive -config ~/revive-aliases.toml -formatter stylish ./...
main.go
  (5, 2)  https://revive.run/r#redundant-import-alias  Import alias "redundant" is redundant  


 ✖ 1 problem (0 errors) (1 warnings)
```

after set ignoreUsed = true

```
[rule.redundant-import-alias]
  arguments = [{ignoreUsed = true}]
```

linter dont produce warnings that is the case :)

denisvmedia: It is indeed redundant. This is the purpose of the linter.

mfederowicz: Yes i know, it is specific case, because in one way you should think how you use packages and aliases for them. On the other side we can leave this rule with that argument for more flexibility, but this is not my decision, what you think @chavacava

denisvmedia: What is so specific about this case? `ignoreUsed` seems to be essentially disabling the rule.

mfederowicz: @denisvmedia  ok, I think that issue #936 and this  pull request is to close, because in standard scenario this rule is used to detect redundant aliases, and results should used to fix/cleanup codebase of your project (change redundant aliases to proper form)

PR #952, opened 2023-12-24
cleanup rules functions declarations

mfederowicz:
cleanup rules functions declarations (add variable name next to struct name)

### DIFF
--- a/rule/add-constant.go
+++ b/rule/add-constant.go
@@ -63,7 +63,7 @@ func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lin
 }
 
 // Name returns the rule name.
-func (*AddConstantRule) Name() string {
+func (r *AddConstantRule) Name() string {
 	return "add-constant"
 }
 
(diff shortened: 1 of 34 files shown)


rule/add-constant.go
```diff
@@ -63,7 +63,7 @@ func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lin
 }
 
 // Name returns the rule name.
-func (*AddConstantRule) Name() string {
+func (r *AddConstantRule) Name() string {
```
denisvmedia, line 66: What is the reason of adding it? It's unused. Here and in most of the other places you did that.

mfederowicz: ok maybe for Name() was not a good idea, i reverted changes

rule/constant-logical-expr.go
```diff
@@ -11,7 +11,7 @@ import (
 type ConstantLogicalExprRule struct{}
 
 // Apply applies the rule to given file.
-func (*ConstantLogicalExprRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+func (r *ConstantLogicalExprRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
```
denisvmedia, line 14: `r` is unused here.

rule/constant-logical-expr.go
```diff
@@ -63,7 +63,7 @@ func (w *lintConstantLogicalExpr) Visit(node ast.Node) ast.Visitor {
 	return w
 }
 
-func (*lintConstantLogicalExpr) isOperatorWithLogicalResult(t token.Token) bool {
+func (l *lintConstantLogicalExpr) isOperatorWithLogicalResult(t token.Token) bool {
```
denisvmedia, line 66: `l` is unused here.

rule/constant-logical-expr.go
```diff
@@ -72,7 +72,7 @@ func (*lintConstantLogicalExpr) isOperatorWithLogicalResult(t token.Token) bool
 	return false
 }
 
-func (*lintConstantLogicalExpr) isEqualityOperator(t token.Token) bool {
+func (l *lintConstantLogicalExpr) isEqualityOperator(t token.Token) bool {
```
denisvmedia, line 75: `l` is unused here. Please check other places where it is unused. This will make the PR smaller.

mfederowicz:
sorry @denisvmedia for that PR but I thought that all rules files should have consistent Receiver Prefix/Suffix naming convention, but after reading all rules files I made decision to close this PR becase maybe not all rules function have the same Receiver Prefix but all of them have some (of course only if they are used).

PR #953, opened 2023-12-26
feat: add support for enforce-repeated-arg-type-style rule

### LINKED ISSUE #948: Add a linter rule to enforce omitting the type for the same type arguments in a function
Consider this function declaration:

```go
func Dummy(arg1 string, arg2 string)
```

It can be simplified to

```go
func Dummy(arg1, arg2 string)
```

essentially making the signature shorter.

This rule can be (and probably should be) also extended to the return values:

```go
func Dummy() (ret1 string, ret2 string)
```

to

```go
func Dummy() (ret1, ret2 string)
```

Note, on the opposite one might want to enforce to have the full signature all the time.

I'd like to have a linter rule to enforce the style. Proposed linter rule name is `enforce-repeated-arg-type-style` (I'm open to accept a better name though).

Simple configuration (applies to both: args and ret vals):

```
[rule.enforce-repeated-arg-type-style]
  arguments = [ "short" ] # can be any, short or full; default: any
```

Advanced configuration (individual configuration per type/local of val):

```
[rule.enforce-repeated-arg-type-style]
  # possible values for the style: any, short, full; default: any to both
  arguments = [ { funcArgStyle = "short", funcRetValStyle = "full" } ]
```

I'm not aware of any existing alternatives.

**Additional context**
This would bring a little bit more consistency to the code base. Suggested default value: "any" (this will make the rule essentially disabled by default even if all rules are enabled in the configuration, thus no impact on any existing system without explicit configuration).

I'm ready to provide a PR if there are no serious concerns (and no good alternatives that I might have overseen).

denisvmedia:
closes #948


chavacava:
Thanks for the proposal and the PR @denisvmedia

PR #954, opened 2023-12-27
fix: add-constant struct tags in anonymous struct literals false positive

### LINKED ISSUE #951: add-constant suggests invalid fix of replacing struct tags with string constants for anonymous structs
**Describe the bug**

revive incorrectly suggests replacing duplicated struct tags with string constants, for tags found in anonymous struct literals.

Example file to reproduce:

```
package test

func x() {
	a := struct {
		X int `json:"x"`
	}{}

	b := struct {
		X int `json:"x"`
	}{}

	c := struct {
		X int `json:"x"`
	}{}

	_, _, _ = a, b, c
}
```

**To Reproduce**

```shell
# flags

revive -config config.toml .
```

Output:

```
test.go:13:9: string literal `json:"x"` appears, at least, 3 times, create a named constant for it
```

```toml
[rule.add-constant]
```

**Expected behavior**
Struct tags in anonymous struct literals should be ignored by this rule, as they are for struct type definitions.

denisvmedia:
closes #951

### DIFF
--- a/rule/add-constant.go
+++ b/rule/add-constant.go
@@ -49,12 +49,13 @@ func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lin
 		failures = append(failures, failure)
 	}
 
-	w := lintAddConstantRule{
+	w := &lintAddConstantRule{
 		onFailure:       onFailure,
 		strLits:         make(map[string]int),
 		strLitLimit:     r.strLitLimit,
 		whiteLst:        r.whiteList,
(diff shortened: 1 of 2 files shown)


rule/add-constant.go
```diff
@@ -171,6 +187,12 @@ func (w lintAddConstantRule) checkNumLit(kind string, n *ast.BasicLit) {
 	})
 }
 
+// isStructTag checks if the given BasicLit is part of a struct tag.
```
denisvmedia, line 190:
```suggestion
```

PR #957, opened 2024-01-04
fix: remove errored replace directive, fix go install

### LINKED ISSUE #956: Cannot install 1.3.5
**Describe the bug**
I get the following error when installing 1.3.5. 1.3.4 works fine.

```shell
> go install github.com/mgechev/revive@latest  
go: downloading github.com/mgechev/revive v1.3.5
go: github.com/mgechev/revive@latest (in github.com/mgechev/revive@v1.3.5):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.
```

This happens on my local machine (Arch Linux) and on GitHub actions (Ubuntu 22.04). Both systems have go 1.21.

EXHades:
try fix #956


denisvmedia:
It seems like indeed, there is no need to have afero replace directive anymore.

jan-xyz:
Can we get a new patch release with this fix? The latest version 1.3.5 is currently broken and cannot install :(

PR #964, opened 2024-01-14
feat: capitalise var naming to IDs

### LINKED ISSUE #947: var-naming rule to check that plural vars with ID are capitalised
the var-naming rule only checks that vars ending in ID are capitalised (e.g. userId => userID). I think it would be nice that it also check that vars ending in IDs are checked. (e.g. userIds => userIDs).

Add a condition to the check that plurals vars such as userIDs are well capitalised.

VincentBaron:
`var-naming` does not check that ID is capitalised when it is plural such as IDs. For example if I have `uIds := []string{}` it will allow this var name where really it should return var `uIds should be uIDs`.

Closes #947

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ vendor
 *.swp
 dist/
 *.log
-.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ vendor
 *.swp
 dist/
 *.log
+.vscode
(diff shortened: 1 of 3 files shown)


chavacava:
Hi @VincentBaron, thanks for the PR!
Could you please add some test cases.

.gitignore
```diff
@@ -4,3 +4,4 @@ vendor
 *.swp
 dist/
 *.log
+.vscode
```
denisvmedia, line 7: nit: This seems to be unrelated to the PR.

VincentBaron:
@denisvmedia do you know when you are planning the next tag? As I am using revive via golangci-lint I will need to do a PR to them to update to your latest version :)

PR #966, opened 2024-01-22
Check function literals in `unused-param`

### LINKED ISSUE #965: Expand `unused-parameter` to check anonymous functions

The `unused-parameter` is phenomenally useful, however it doesn't seem to include anonymous functions. Had this linter included anonymous functions, we would have been able to catch a bug before it made its way into a recent release of ours.

Anonymous functions should be tracked just like normal functions for unused variable usages.

In [this example](https://go.dev/play/p/Kl3o91k9jl2), `a` is unused. There could be a suggestion to use `func(int)` (or `func(_ int)` rather than `func(a int)`.

Depending on how `unused-parameter` is implemented, it might make sense to have this be a separate rule entirely, like `unused-receiver`.

**Additional context**

Here is a [bug fix](https://github.com/ava-labs/coreth/pull/461/files#diff-4968a027180246a1a0b75376b875471a5d85755bac928434cea30b23b153240dL155-L158) that would have been prevented had this linter existed.

StephenButtolph:
Closes #965

See #965 for the motivation behind this PR.

- [X] Added tests
- [X] I believe this code follows the existing style, but if this could be improved I'm very open to any feedback
- [x] Passes CI

This does potentially significantly change the performance of this rule because we will now walk into the function bodies. If we are concerned about a performance regression here I think an alternative could be to either add a new rule or add a configuration to the rule to gate checking anonymous functions.

### DIFF
--- a/rule/unused-param.go
+++ b/rule/unused-param.go
@@ -87,54 +87,64 @@ type lintUnusedParamRule struct {
 }
 
 func (w lintUnusedParamRule) Visit(node ast.Node) ast.Visitor {
+	var (
+		funcType *ast.FuncType
+		funcBody *ast.BlockStmt
+	)
 	switch n := node.(type) {
+	case *ast.FuncLit:
+		funcType = n.Type
+		funcBody = n.Body
 	case *ast.FuncDecl:
-		params := retrieveNamedParams(n.Type.Params)
-		if len(params) < 1 {
-			return nil // skip, func without parameters
-		}
-
 		if n.Body == nil {
-			return nil // skip, is a function prototype
+			return w // skip, is a function prototype
 		}
 
-		// inspect the func body looking for references to parameters
-		fselect := func(n ast.Node) bool {
-			ident, isAnID := n.(*ast.Ident)
-
-			if !isAnID {
-				return false
-			}
-
-			_, isAParam := params[ident.Obj]
-			if isAParam {
-				params[ident.Obj] = false // mark as used
-			}
-
-			return false
-		}
-		_ = pick(n.Body, fselect)
-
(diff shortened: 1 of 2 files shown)


rule/unused-param.go
```diff
@@ -87,54 +87,64 @@ type lintUnusedParamRule struct {
...
+		funcBody = n.Body
 	case *ast.FuncDecl:
-		params := retrieveNamedParams(n.Type.Params)
-		if len(params) < 1 {
-			return nil // skip, func without parameters
-		}
-
 		if n.Body == nil {
-			return nil // skip, is a function prototype
+			return w // skip, is a function prototype
```
denisvmedia, line 100: Here as we as in some other places it was returning nil. Now it returns `w`. Why is that?

StephenButtolph: `Visit` is invoked in [ast.Walk](https://pkg.go.dev/go/ast#Walk). Previously the function returned `nil` to stop walking the AST after entering into a function definition. Because anonymous functions can be defined inside of any code block, the function now needs to walk through function bodies looking for `*ast.FuncLit` nodes.

StephenButtolph: I think this specific check could be `nil`, as there can't be any `FuncLit`s inside a function prototype... But it felt a bit more obviously correct to just return the visitor in all the cases rather than singling this case out.
```suggestion
			return nil // skip, is a function prototype
```

chavacava:
Hi @StephenButtolph, thanks for the PR and the proposal for improving the rule. I'll try to review the PR in the coming days.

chavacava:
merged, thanks @StephenButtolph

PR #967, opened 2024-01-27
Adds rule `max-control-nesting`

chavacava:
This PR adds rule `max-control-nesting`, the GO implementation of https://sonarsource.atlassian.net/browse/RSPEC-134

### DIFF
--- a/rule/max-control-nesting.go
+++ b/rule/max-control-nesting.go
@@ -109,18 +109,20 @@ func (w *lintMaxControlNesting) walkControlledBlock(b ast.Node) {
 func (r *MaxControlNestingRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
-	if r.max < 1 {
-		if len(arguments) < 1 {
-			r.max = defaultMaxControlNesting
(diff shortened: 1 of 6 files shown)


denisvmedia:
Thanks for the new rule @chavacava. 

What about select/case - should we (do we?) support it as well?

Another thing: can we have test cases for the code placed in closures?

rule/max-control-nesting.go
```diff
@@ -0,0 +1,108 @@
...
+	oldNestingLevel := w.nestingLevelAcc
+	w.nestingLevelAcc++
+	ast.Walk(w, b)
+	w.nestingLevelAcc = oldNestingLevel
+}
+
+func (r *MaxControlNestingRule) configure(arguments lint.Arguments) {
+	r.Lock()
+	defer r.Unlock()
+	if r.max < 1 {
```
denisvmedia, line 94: nit: I'd invert the condition to reduce the nesting level of the control flow.

denisvmedia (approved):
Thanks! LGTM. Feel free to merge :)

PR #970, opened 2024-02-07
fix: add enforce-repeated-arg-type-style in config

ldez:
Related to #953
Related to https://github.com/golangci/golangci-lint/issues/4353


denisvmedia (approved):

ldez:
@denisvmedia can you create a release?

PR #974, opened 2024-02-25
enforce-slice-style: Support nil declaration enforcement

### LINKED ISSUE #971: enforce-slice-style: Add declaration style

In https://github.com/mgechev/revive/issues/898, it was mentioned:
> There is also a way of having a non-initialized slice (which should be preferred over the empty slice w/o len and cap in most cases), but this is beyond the scope of this linter

This preferred way is the var declaration style:
```
var emptySlice []int
```

This is the preferred style by both the Golang and Uber style guides:
* https://google.github.io/styleguide/go/decisions#nil-slices
* https://github.com/uber-go/guide/blob/master/style.md#local-variable-declarations

It would be great if the `enforce-slice-style` rule could be configured to prefer the var declaration style. The tricky part may be that this style is only preferred when declaring a new variable, but other styles may be preferred when assigning an existing variable or passing a parameter.

The only alternative I have considered is disabling the rule.

**Additional context**

jalaziz:
Add support for enforcing nil slice declarations as recommended by both the Go Code Review and Uber style guides.

This initial version is quite strict in that it also prevents using empty literal and make-style slices in struct assignments and as function arguments.

Closes #971

### DIFF
--- a/testdata/enforce-slice-style-nil.go
+++ b/testdata/enforce-slice-style-nil.go
@@ -1,0 +1,92 @@
+package fixtures
+
+func somefn() {
+	m0 := make([]string, 10)
+	m1 := make([]string, 0, 10)
+	m2 := make([]string, 0)    // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
+	m3 := make([]string, 0, 0) // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
+	m4 := []string{}           // MATCH /use nil slice declaration (e.g. var args []type) instead of []type{}/
+	m5 := []string{"v1", "v2"}
+	m6 := [8]string{}
+	m7 := [...]string{}
+	var m8 []string
+
+	_ = m0
+	_ = m1
+	_ = m2
+	_ = m3
+	_ = m4
+	_ = m5
+	_ = m6
+	_ = m7
+	_ = m8
+}
+
+type Slice []string
+
+func somefn2() {
+	m0 := make(Slice, 10)
+	m1 := make(Slice, 0, 10)
+	m2 := make(Slice, 0)    // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
+	m3 := make(Slice, 0, 0) // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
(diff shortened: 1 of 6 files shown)


testdata/enforce-slice-style-nil.go
```diff
@@ -0,0 +1,92 @@
...
+func somefn() {
+	m0 := make([]string, 10)
+	m1 := make([]string, 0, 10)
+	m2 := make([]string, 0)    // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
+	m3 := make([]string, 0, 0) // MATCH /use nil slice declaration (e.g. var args []type) instead of make([]type, 0)/
+	m4 := []string{}           // MATCH /use nil slice declaration (e.g. var args []type) instead of []type{}/
+	m5 := []string{"v1", "v2"}
+	m6 := [8]string{}
+	m7 := [...]string{}
+	var m8 []string
```
denisvmedia, line 12: What about these cases:

```
var m9 []string = make([]string, 0)
var m10 = make([]string, 0)
m11 := []string(nil)
var m12 []string = []string{}
var m13 = []string{}
var m14 = []string(nil)
var m15 []string = nil
```

?

Also, how will this rule handle the following:

```
var m []string
m = make([]string, 0)
m = []string{}
```

? Is it meant to be an issue or on the opposite is not an issue?

And this (here it's artificial but instead of len there can be some meaningful function):

```
	if m := []string(nil); len(m) == 0 {
		fmt.Println("Hello, 世界")
	}
```

jalaziz: I've added more test cases to cover all these examples.

> ? Is it meant to be an issue or on the opposite is not an issue?

Currently, it is an issue and I think it's okay for it to be an issue.

Ideally, there would be a separate setting for declarations and assignments but this rule currently doesn't make a distinction between the two. Most style guides focus on slice declaration, but assignments are not well-defined so I think it makes sense to enforce different styles on assignment.

> And this (here it's artificial but instead of len there can be some meaningful function):

All of the rules today allow `nil` assignments, so that would be okay. I'm not sure what the preferred style should be in this case, but it seems like a pretty rare scenario (defining an empty slice in a conditional initializer).

jalaziz:
> Thank you for your PR @jalaziz , do you mind also updating the docs?

Done, let me know if you'd like me to add more detail 🙏🏼

chavacava:
LGTM

PR #979, opened 2024-04-13
adds `comments-density` rule

chavacava:
This PR adds a new rule `comments-density`.
The rule is inspired on the SonarQube rule on the [_comment lines density_ metric](https://docs.sonarsource.com/sonarqube/latest/user-guide/metric-definitions/)

The default argument value of the rule is `0`, i.e. the rule is de-facto deactivated, to avoid polluting analysis of those using `enableAllRules = true` in the configuration.

### DIFF
--- a/rule/comments-density.go
+++ b/rule/comments-density.go
@@ -22,18 +22,21 @@ func (r *CommentsDensityRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
 
-	if !r.configured {
-		r.configured = true
-		if len(arguments) < 1 {
-			r.minumumCommentsDensity = defaultMinimumCommentsPercentage
-			return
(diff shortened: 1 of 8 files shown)


denisvmedia:
I'd happily review it tomorrow evening (PRG time).

Thanks for the PR!

rule/comments-density.go
```diff
@@ -0,0 +1,92 @@
...
+	sync.Mutex
+}
+
+const defaultMinimumCommentsPercentage = 0
+
+func (r *CommentsDensityRule) configure(arguments lint.Arguments) {
+	r.Lock()
+	defer r.Unlock()
+
+	if !r.configured {
```
denisvmedia, line 25: nit: I'd invert the condition to outdent the inner code.

Issue #982, opened 2024-04-29
Possible false positive confusing-naming error on generics with pointer type receiver


emredalkiran:
**Describe the bug**
`confusing-naming` check possibly misfires on different named methods of generic types.

**To Reproduce**
1. I updated revive `go install github.com/mgechev/revive@latest`
2. I run it with the following flags & configuration file:
command:
```shell
revive -config revive.toml test.go
```
config:
```shell
[rule.confusing-naming]
```
test.go:
```golang
package test


type a[T any] struct{}

func (x * a[T]) method() {
}

type b[T any] struct{}

func (x *b[T]) method() {
}
```
**Expected behavior**
Since methods are on different types, I expect check to pass

**Logs**
```shell
 ✘  https://revive.run/r#confusing-naming  Method 'method' differs only by capitalization to function 'method' in the same source file
  test/test.go:11:16

✘ 1 problem (1 error, 0 warnings)

Errors:
  1  confusing-naming
  ```

**Desktop (please complete the following information):**
 - OS: [macOS]
 - Version of Go: 1.22

**Additional context**
The issue is previously mentioned in #864 but the error persists if the receiver of the method is of pointer type. Is this the expected behaviour?

I'm relatively new to `go`, sorry if this issue is due to my misunderstanding or lack of knowledge.

chavacava:
Hi @emredalkiran thanks for reporting the issue. I think the linter should be able to differentiate both methods and do not raise a warning in cases like yours.

Entuazism:
Hi! @chavacava can you assign it to me?

PR #985, opened 2024-05-08
Fix false positive confusing-naming error on generics with pointer type receiver

Entuazism:
Fix logic to handle expressions that include pointers to generic types.
Add test data

Closes https://github.com/mgechev/revive/issues/982

### DIFF
--- a/rule/confusing-naming.go
+++ b/rule/confusing-naming.go
@@ -138,16 +138,32 @@ func getStructName(r *ast.FieldList) string {
 
 	switch v := t.(type) {
 	case *ast.StarExpr:
-		t = v.X
+		return extractFromStarExpr(v)
 	case *ast.IndexExpr:
-		t = v.X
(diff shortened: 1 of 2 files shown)


rule/confusing-naming.go
```diff
@@ -136,11 +136,12 @@ func getStructName(r *ast.FieldList) string {
...
-	case *ast.StarExpr:
-		t = v.X
-	case *ast.IndexExpr:
-		t = v.X
+	if p, _ := t.(*ast.StarExpr); p != nil {
+		t = p.X
+	}
+
+	if p, _ := t.(*ast.IndexExpr); p != nil {
+		t = p.X
```
denisvmedia, line 144: Preserving the original switch would be more idiomatic and let us avoiding type assertions multiple times.

Entuazism: I tried to avoid a lot of code nesting. Returned the switch

PR #986, opened 2024-05-10
Fix #808

### LINKED ISSUE #808: comment-spacings false positives for nolint directives
**Describe the bug**
I'm using revive with golangci-lint and since the latest update I'm getting errors like this:
```
api/common/validate/validate.go:15:1: comment-spacings: no space between comment delimiter and comment text (revive)
//nolint:gochecknoglobals
```
I've updated the config to contain `nolint` in the `comment-spacings` arguments to make it go away but it's odd as this was not happening before.

**To Reproduce**
1. I updated golangci-lint:
```sh
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin
```

```shell
golangci-lint run
```

```yaml
linters:
  disable-all: true
  enable:
    - revive
linters-settings:
  revive:
    rules:
      - name: comment-spacings
        disabled: false
```

 - OS: Ubuntu 22.04.2 LTS
 - go version go1.20.2 linux/amd64

chavacava:
Closes #808 by ignoring comments matching the official pattern for directives

### DIFF
--- a/rule/utils.go
+++ b/rule/utils.go
@@ -165,3 +165,9 @@ func checkNumberOfArguments(expected int, args lint.Arguments, ruleName string) 
 		panic(fmt.Sprintf("not enough arguments for %s rule, expected %d, got %d. Please check the rule's documentation", ruleName, expected, len(args)))
 	}
 }
+
+var directiveCommentRE = regexp.MustCompile("//(line |extern |export |[a-z0-9]+:[a-z0-9])") // see https://go-review.googlesource.com/c/website/+/442516/1..2/_content/doc/comment.md#494
+
+func isDirectiveComment(line string) bool {
+	return directiveCommentRE.MatchString(line)
+}
--- a/rule/comment-spacings.go
+++ b/rule/comment-spacings.go
@@ -22,7 +22,7 @@ func (r *CommentSpacingsRule) configure(arguments lint.Arguments) {
 	defer r.Unlock()
 
 	if r.allowRe == nil {
-		r.allowRe = regexp.MustCompile("//(line|extern|export|[a-z0-9]+:[a-z0-9])") // see https://go-review.googlesource.com/c/website/+/442516/1..2/_content/doc/comment.md#494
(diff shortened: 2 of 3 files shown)


denisvmedia (approved):
LGTM

ccoVeille:
For reference, here is what is implemented in gofumpt

https://github.com/mvdan/gofumpt/blob/master/format%2Fformat.go#L334

They use a slightly different regexp

You can look at historical changes, it gives context

rule/utils.go
```diff
@@ -165,3 +165,9 @@ func checkNumberOfArguments(expected int, args lint.Arguments, ruleName string)
 		panic(fmt.Sprintf("not enough arguments for %s rule, expected %d, got %d. Please check the rule's documentation", ruleName, expected, len(args)))
 	}
 }
+
+var directiveCommentRE = regexp.MustCompile("//(line |extern |export |[a-z0-9]+:[a-z0-9])") // see https://go-review.googlesource.com/c/website/+/442516/1..2/_content/doc/comment.md#494
```
thaJeztah, line 169: I wonder if it needs to be anchored (`^`); I did a quick try using the comment below, and it didn't flag that as invalid;

```go
//this is a regular command that's incorrectly formatted //nolint:foobar // because one two three
```


(disclaimer: I only used the function directly, not in context of the whole code, so perhaps there's already some other code handling this)

ccoVeille: I think you are right, that's why I raised the point

https://github.com/mgechev/revive/pull/986#issuecomment-2104473372

A comment like this may cause issues or being wrongly reported

    // commenting a nolint //nolint:whatever // whatever is a false positive here

ccoVeille: It was fixed with #988

thaJeztah: Thanks!

ccoVeille:
Please note this change had a unexpected (but normal) consequence for me

Previously the following syntax were allowed

```go
//nolint: somelinter // what ever the reason
```

This one was somehow invalid, and excluded by the code before this PR

not it reports a normal error, because it expects `no space between comment delimiter and comment text`

```go
//nolint:somelinter // what ever the reason
```

PR #987, opened 2024-05-11
fix regexp of directive comments

chavacava:
Fix the regular expression for identifying directive comments and handle cases like

`//this is a regular command that's incorrectly formatted //nolint:foobar // because one two three`


chavacava:
deprecated

PR #988, opened 2024-05-11
more precise regular expression for directive comments

chavacava:
Improves regular expression for handling cases like 

`//this is a regular command that's incorrectly formatted //nolint:foobar // because one two three`


ccoVeille:
This PR supersedes #986 and is related to #808

PR #996, opened 2024-06-03
docs: clean rule descriptions

### LINKED ISSUE #991: out-of-order rules in documentation
https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md

error-naming, error-return, error-strings, and errorf are currently listed between empty-lines and enforce-map-style.

They should instead appear between enforce-slice-style and exported. (They are correct in the table of contents, but not the actual content.)

ldez:
- add missing lines
- add missing `[rule.` prefix
- change rule order

Fixes #991


ccoVeille:
LGTM thanks

Issue #1002, opened 2024-06-28
rule `exported` add `checkPublicInterface` option to check documentation in public interface definition


comdiv:
For now this code is valid:

```go
// Some - some interface
type Some interface {
     // Correct - should do all correct
     Correct()
     // not valid format of comment
     SemiCorrect()
     NonCorrect() // no documentation
}
```

suggest to add `checkPublicInterface` to `exported` rule to check documentation on interface method declaration,
so here `SemiCorrect()` has not valid comment, and `NonCorrect()` doesn't have it at all

chavacava:
Hi @comdiv, thank you for reporting the issue and opening a PR (I'll check it in the next days)

arxeiss:
I see the #1002 was merged, however no release was made since then. Can you please release new version so we can use this feature with GolangCI-Lint?

Thank you

zignd:
@mgechev, @chavacava, or other collaborators. Do you have any plans for when to release a new version so that we can use this new flag?

arxeiss:
Based on https://github.com/mgechev/revive/releases/tag/v1.4.0 it should be released. 
I guess this issue can be closed.

PR #1003, opened 2024-06-28
#1002 - add "checkPublicInterface" option to "exported" rule to allow check documentation on public methods on public interfaces

comdiv:
Feature implementation for #1002


rule/exported.go
```diff
@@ -70,7 +77,12 @@ func (*ExportedRule) Name() string {
 	return "exported"
 }
 
-func (r *ExportedRule) getConf(args lint.Arguments) (checkPrivateReceivers, disableStutteringCheck, sayRepetitiveInsteadOfStutters bool) {
+func (r *ExportedRule) getConf(args lint.Arguments) (
+	checkPrivateReceivers,
+	disableStutteringCheck,
+	sayRepetitiveInsteadOfStutters bool,
+	checkPublicInterface bool,
+) {
```
denisvmedia, line 85: This seems to be a good candidate for refactoring. Instead of adding another bool value, I'd refactor it to a struct.

chavacava:
Hi @comdiv. Are you still working in the PR? (some tests are not passing)

rule/exported.go
```diff
@@ -330,11 +319,54 @@ func (w *lintExported) Visit(n ast.Node) ast.Visitor {
...
+		w.onFailure(lint.Failure{
+			Node:       m,
+			Confidence: 1,
+			Category:   "comments",
+			Failure:    fmt.Sprintf("public interface method %s.%s should be commented", typeName, name),
+		})
+		return
+	}
+	s := normalizeText(m.Doc.Text())
+	prefix := m.Names[0].Name + " "
```
ccoVeille, line 363: This line needs a comment to explain why you are doing this.

I get it, but it would be better.

But this pattern is common in the current code, you are only reusing it.

So maybe a func hasComment could help.

But it's somehow out of the scope of this PR

ccoVeille (commented):
I found a few things. Maybe you could consider them

comdiv:
> I found a few things. Maybe you could consider them

yes, reading notes... and what suggestions about fixing test issues (renaming donuts and others) it's not my own code, what have i do by project policy?

rule/exported.go
```diff
@@ -330,11 +319,54 @@ func (w *lintExported) Visit(n ast.Node) ast.Visitor {
...
+			continue
+		}
+		if ast.IsExported(m.Names[0].Name) {
+			w.lintInterfaceMethod(typeName, m)
+		}
+	}
+}
+
+func (w *lintExported) lintInterfaceMethod(typeName string, m *ast.Field) {
+	name := m.Names[0].Name
```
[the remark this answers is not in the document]
comdiv: ok, but at this place m.Names always non empty, it will fix it due to more readable, but if someone will chaeck coverage it will be dgraded

rule/exported.go
comdiv: yeah, i have not apply much refactoring because of "tradition"

comdiv:
Cannot test type comment linting. Issue appears at 36 line of exported go, but if i place `//MATCH` at same line as comment - it doesnt work as expected...

ccoVeille (approved):

mgechev:
@chavacava I'm ready to merge when you also approve!

rule/exported.go
```diff
@@ -18,23 +18,36 @@ type ExportedRule struct {
...
-		var sayRepetitiveInsteadOfStutters bool
-		r.checkPrivateReceivers, r.disableStutteringCheck, sayRepetitiveInsteadOfStutters = r.getConf(arguments)
 		r.stuttersMsg = "stutters"
-		if sayRepetitiveInsteadOfStutters {
-			r.stuttersMsg = "is repetitive"
+		for _, flag := range arguments {
+			flagStr, ok := flag.(string)
+			if !ok {
+				panic(fmt.Sprintf("Invalid argument for the %s rule: expecting a string, got %T", r.Name(), flag))
+			}
```
denisvmedia, line 35: nit: the switch/case below can work without this assertion (it can compare `flag` to case values directly).

testdata/exported-issue-1002.go
```diff
@@ -0,0 +1,24 @@
...
+}
+
+// Other - just to check names compatibility
+type Other interface {}
+
+// for private interfaces it doesn't check docs anyway
+
+type somePrivate interface {
+	AllGood()
+}
```
denisvmedia, line 24: nit

```suggestion
}

```

denisvmedia (approved):
LGTM (few nits though)

PR #1005, opened 2024-07-01
Rule optimize-operands-order: do not consider built-in len as a caller

### LINKED ISSUE #1004: Rule `optimize-operands-order` reports potentially dangerous change when `len` of slice is used before slice accesing.
**Describe the bug**
In this case:

```golang
if len(s) > 0 || s[0] == "" { // should not match, not safe
    // ...
}
```

Rule `optimize-operands-order` reports 

```plain
for better performance 'len(s) == 0 || s[0] == ""' might be rewritten as 's[0] == "" || len(s) == 0'
```

Which could lead to panic if implemented. 

**To Reproduce**

```shell
# no extra flags
revive -config revive.toml .
```

```toml
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.3
errorCode = 0
warningCode = 0

[rule.optimize-operands-order]
```

**Expected behavior**
I think in this case, where the `len` check is essential to happen before slice accessing, the rule should not match. 

**Additional context**
My first thought about fixing this was to not consider the built-in `len` function as a caller. 
I tried to change the `isCaller` function in the [optimize-operands-order.go](https://github.com/mgechev/revive/blob/0df1bb0860f7b25091b9227cccd45f69a28ab37a/rule/optimize-operands-order.go#L51) file so that it does not consider the built-in `len` function as a caller.

I opened a pr with this approach.

moukoublen:
Fixes: https://github.com/mgechev/revive/issues/1004

Update rule `optimize-operands-order`, `isCaller` logic to not consider built-in function `len` as a caller. 

Closes #1004

### DIFF
--- a/rule/optimize-operands-order.go
+++ b/rule/optimize-operands-order.go
@@ -49,8 +49,17 @@ func (w lintOptimizeOperandsOrderlExpr) Visit(node ast.Node) ast.Visitor {
 	}
 
 	isCaller := func(n ast.Node) bool {
-		_, ok := n.(*ast.CallExpr)
-		return ok
+		ce, ok := n.(*ast.CallExpr)
+		if !ok {
+			return false
+		}
+
+		ident, isIdent := ce.Fun.(*ast.Ident)
+		if !isIdent {
+			return true
+		}
+
+		return ident.Name != "len" || ident.Obj != nil
 	}
 
 	// check if the left sub-expression contains a function call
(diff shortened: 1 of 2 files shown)


chavacava:
Hi @moukoublen, thanks for the PR.
(I'm sure in the future somebody will ask for a configurable list of _functions to ignore_ but let things simple for the moment)

PR #1009, opened 2024-07-10
Add multiple scopes support to string-format rule

### LINKED ISSUE #1008: Multiple scopes to string-format rule
I stated to use **string-format** rule to lint some function arguments and found problem that I can not use one regex to validate some scopes. I should repeat same regex multiple times for different scopes.

I suggest separate one scope from another by `,` in one rule to make config more readable.

Now I should configure rule something like this:
```toml
[rule.string-format]
  arguments = [
    ["core.WriteError[1].Message", "/^([^A-Z]|$)/", "must not start with a capital letter"],
    ["fmt.Errorf[0]", "/^([^A-Z]|$)/", "must not start with a capital letter"],
    ["panic", "/^([^A-Z]|$)/", "must not start with a capital letter"],
  ]
```

I want to configure it like this:
```toml
[rule.string-format]
  arguments = [
    ["core.WriteError[1].Message,mt.Errorf[0],panic", "/^([^A-Z]|$)/", "must not start with a capital letter"],
  ]
```

KirillSluchaev:
Hi, I added new feature.

I describe my motivation in issue.

I added some tests and refactor old one little bit, becouse change error message.
My code almost follow the coding style of the rest of the repository.
Build passed.

Closes #1008


rule/string-format.go
```diff
@@ -125,26 +128,33 @@ func (w lintStringFormatRule) parseArgument(argument any, ruleNum int) (scope st
...
-		w.parseError(fmt.Sprintf("unexpected number of submatches when parsing scope: %d, expected 4", len(matches)), ruleNum, 0)
-	}
-	scope.funcName = matches[1]
-	if len(matches[2]) > 0 {
-		var err error
-		scope.argument, err = strconv.Atoi(matches[2])
-		if err != nil {
-			w.parseError("unable to parse argument number in rule scope", ruleNum, 0)
+	// Parse rule scopes
+	rawScopes := strings.Split(rule[0], ",")
```
ccoVeille, line 132: I'm a bit worried about the fact people may use `, `, or ` ,`, or ` , ` as delimiter.

So maybe, you could use strings.TrimSpace on the rawScope in the for loop

Also, xhecking that the rawScope is not empty could avoid surprises and bugs

I'm pretty sure someone would face a 

`foo,,bar`, or `foo , , bar`

No matter the implementation choice to either clean them, or forbid them. It should be covered by a test

ccoVeille:
I like the proposal and the way you implemented it, I found a few things that might be improved maybe

KirillSluchaev:
Add behavior from comment

rule/string-format.go
```diff
@@ -66,12 +67,14 @@ type lintStringFormatRule struct {
...
 type stringFormatSubrule struct {
 	parent       *lintStringFormatRule
-	scope        stringFormatSubruleScope
+	scopes       stringFormatSubruleScopes
 	regexp       *regexp.Regexp
 	negated      bool
 	errorMessage string
 }
 
+type stringFormatSubruleScopes []*stringFormatSubruleScope
```
denisvmedia, line 76: out of curiosity: why is that a slice of pointers? (not a slice of values)

KirillSluchaev: I saw one video on youtube where this two approaches were compared and slice of pointers use less memory little bit)

KirillSluchaev:
@chavacava Hi, could you review my PR, please)

RULES_DESCRIPTIONS.md
```diff
@@ -775,6 +775,7 @@ Example:
     ["core.WriteError[1].Message", "/^([^A-Z]|$)/", "must not start with a capital letter"],
     ["fmt.Errorf[0]", "/(^|[^\\.!?])$/", "must not end in punctuation"],
     ["panic", "/^[^\\n]*$/", "must not contain line breaks"],
+    ["fmt.Errorf[0],core.WriteError[1].Message", "/^.*%w.*$/", "must not contains '%w'"],
```
chavacava, line 778:
```suggestion
    ["fmt.Errorf[0],core.WriteError[1].Message", "/^.*%w.*$/", "must not contain '%w'"],
```
Using this conf the rule (wrongly) matches things like 
```go
return fmt.Errorf("not implemented: AlterableDoltTable.DropPrimaryKey()"
```

chavacava:
Thanks for the PR!
I've tested the rule using the example configuration in the documentation and against real code bases (https://github.com/dolthub/dolt for example) and the rule spots, what I think, false positives (see my previous comment)

KirillSluchaev:
I made mistake in rule example. I fixed it, you can check again. Regexp rule should be inversed to lint only strings which contains '%w'

PR #1010, opened 2024-07-11
Update README.md & Typo

hatamiarash7:
- Fix the Docker image version. The `v1.1.2-next` tag is not available, so we can use the latest released version: `v1.3.7`
- Fix some problems with syntax and grammar
- Fix typo for `minimum` word


chavacava:
Thanks @hatamiarash7

Issue #1007, opened 2024-07-08
negative add-constant report when expr is not **ast.Ident**


lsytj0413:
**Describe the bug**

**To Reproduce**
1. I updated revive `go install github.com/mgechev/revive@latest`
2. I run it with the following flags & configuration file:

```go
package main

import (
	"context"
	"log/slog"
)

func main() {
	slog.Default().InfoContext(context.Background(), "key1")
	slog.Default().InfoContext(context.Background(), "key1")
	slog.Default().InfoContext(context.Background(), "key1")
	slog.Default().InfoContext(context.Background(), "key1")

	slog.InfoContext(context.Background(), "key2")
	slog.InfoContext(context.Background(), "key2")
	slog.InfoContext(context.Background(), "key2")
	slog.InfoContext(context.Background(), "key2")
}
```

```shell
# flags
golangci-lint run
```

```toml
linters:
  enable:
    - revive

linters-settings:
  revive:
    enable-all-rules: true
    rules:
      - name: add-constant
        severity: warning
        disabled: false
        arguments:
          - maxLitCount: "3"
            ignoreFuncs: 'slog\.*,fmt\.*,InfoContext\.*'
```

**Expected behavior**

It didn't report "key1" for add-constant.

**Logs**

```
main.go:12:51: add-constant: string literal "key1" appears, at least, 4 times, create a named constant for it (revive)
        slog.Default().InfoContext(context.Background(), "key1")
```

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - Version of Go
go version go1.22.2 darwin/arm64
golangci-lint has version v1.59.0 built with go1.22.2 from (unknown, modified: ?, mod sum: "h1:st69YDnAH/v2QXDcgUaZ0seQajHScPALBVkyitYLXEk=") on (unknown)

**Additional context**

PR #1012, opened 2024-07-15
fix(1007): return Sel.Name as FuncName when selector is an CallExpr

lsytj0413:
Fix: https://github.com/mgechev/revive/issues/1007

When the func selector is an callexpr, for example:

`getLogger().Info(context.Background(), "test1007")`

it will return empty string as function name, so ingoreFuncs will not take affect, this fixed change it to return `Sel.Name`


denisvmedia:
Thank you for your efforts!

PR #1024, opened 2024-08-14
check clear/max/min in redefines-builtin-id rule on go1.21+

### LINKED ISSUE #1021: Request: treat min, max and clear as built-in functions

go 1.21 has introduced new functions `min`, `max` and `clear`.
https://go.dev/ref/spec#Go_1.21

Currently revive does not treat `min`, `max` and `clear` as built-in functions,
https://github.com/mgechev/revive/blob/master/rule/redefines-builtin-id.go#L18-L34
so it does not report `redefinition of the built-in function` error.

```golang
// Reproduce issue
package main

import "fmt"

func main() {
	min := 0
	fmt.Println(min)
}
```

```
❯ revive -set_exit_status ./...

❯ echo $?
0
```

revive treats `min`, `max` and `clear` as built-in functions on go 1.21+,
and reports the following error:
```
main.go: redefinition of the built-in function min
```

There may be an option not to make changes in order to maintain backward compatibility.

**Additional context**

NA

skaji:
Close #1021 

go1.21 has introduced new functions `clear`, `max` and `min`.
It would be nice revive checks these functions in redefines-builtin-id rule on go1.21+.

### DIFF
--- a/lint/package.go
+++ b/lint/package.go
@@ -33,6 +33,7 @@ var (
 	falseValue = 2
 	notSet     = 3
 
+	go121 = goversion.Must(goversion.NewVersion("1.21"))
 	go122 = goversion.Must(goversion.NewVersion("1.22"))
 )
 
@@ -192,6 +193,11 @@ func (p *Package) lint(rules []Rule, config Config, failures chan Failure) {
 		})(file)
 	}
 	wg.Wait()
+}
+
+// IsAtLeastGo121 returns true if the Go version for this package is 1.21 or higher, false otherwise
+func (p *Package) IsAtLeastGo121() bool {
+	return p.goVersion.GreaterThanOrEqual(go121)
 }
 
 // IsAtLeastGo122 returns true if the Go version for this package is 1.22 or higher, false otherwise
--- a/rule/redefines-builtin-id.go
+++ b/rule/redefines-builtin-id.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/token"
+	"maps"
 
 	"github.com/mgechev/revive/lint"
 )
@@ -31,6 +32,12 @@ var builtFunctions = map[string]bool{
 	"println": true,
 	"real":    true,
 	"recover": true,
+}
+
+var builtFunctionsAfterGo121 = map[string]bool{
+	"clear": true,
+	"max":   true,
+	"min":   true,
 }
 
 var builtInTypes = map[string]bool{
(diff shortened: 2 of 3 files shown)


testdata/redefines-builtin-id.go
```diff
@@ -32,3 +32,12 @@ var i, copy int // MATCH /redefinition of the built-in function copy/
 
 // issue #792
 type ()
+
+func foo() {
+	clear := 0 // MATCH /redefinition of the built-in function clear/
```
skaji, line 37: revive itself requires go1.21. So I don't add any condition for go version here.
https://github.com/mgechev/revive/blob/master/go.mod#L3


I manually confirmed that the following code does not report any errors:

go.mod:
```go.mod
module playground

go 1.20
```
main.go
```go
// package comment
package main

func main() {
	clear := 0
	_ = clear
}
```

```
❯ revive -set_exit_status

❯ echo $?
0
```
```

File: rule/redefines-builtin-id.go [hunk 2]
```diff
@@ -162,16 +182,16 @@ func (w lintRedefinesBuiltinID) addFailure(node ast.Node, msg string) {
 	})
 }
 
-func (lintRedefinesBuiltinID) isBuiltIn(id string) (r bool, builtInKind string) {
-	if builtFunctions[id] {
+func (w lintRedefinesBuiltinID) isBuiltIn(id string) (r bool, builtInKind string) {
```
denisvmedia, line 185:
```suggestion
func (w *lintRedefinesBuiltinID) isBuiltIn(id string) (r bool, builtInKind string) {
```

skaji: @denisvmedia Thank you for your review.

Applied your suggested change 67857ce.

lint/package.go
```diff
@@ -33,6 +33,7 @@ var (
 	falseValue = 2
 	notSet     = 3
 
+	go121 = goversion.Must(goversion.NewVersion("1.21"))
```
denisvmedia, line 36: I'm wondering: isn't go 1.21 already EOL? Because if it is, we should better use the existing go122 constant.

skaji: > isn't go 1.21 already EOL?

Yes, go 1.21 is EOL.

> Because if it is, we should better use the existing go122 constant.

Does this mean "revive does not care about EOL go"?
Then, I think we should just add clear/max/min to `builtFunctions` unconditionaly.

However, I'm not unwilling to use `go122` / `IsAtLeastGo122` here.

ccoVeille: > Then, I think we should just add clear/max/min to builtFunctions unconditionaly.

👍

Suggesting to avoid using function named max/min is not really a problem, even for version under go 1.21. They could be renamed even if max/min is not available

chavacava: revive itself requires Go 1.21 and it will eventually require 1.22 because 1.21 is EOL but it should be possible to use revive to analyze Go code bases of any version.

skaji:
@denisvmedia @ccoVeille @chavacava 
Thank you all!

PR #1028, opened 2024-08-18
use a default Go version (1.0) in absence of go.mod

### LINKED ISSUE #1027: Error when analyzing packages without go.mod
if the package under analysis hasn't a `go.mod` file, `revive` stops with the error 
```
linting - retrieving failures channel: "<package full path>" doesn't seem to be part of a Go module
```

**To Reproduce**
1. I updated `revive` `go install github.com/mgechev/revive@latest`
1. Check out [vim-go](https://github.com/fatih/vim-go)
1. Remove or rename the file `vim-go/autoload/go/test-fixtures/lint/src/errcheck/go.mod`
3. From the `go-vimI` directory, run `revive`:

```shell
revive ./...
```

**Expected behavior**
`revive` should analyze the go sources without reporting the above error

chavacava:
Closes #1027 by using a default Go version as a fallback in case of absence of a `go.mod` file for the package under analysis.

### DIFF
--- a/lint/linter.go
+++ b/lint/linter.go
@@ -55,8 +55,9 @@ func (l Linter) readFile(path string) (result []byte, err error) {
 }
 
 var (
-	genHdr = []byte("// Code generated ")
-	genFtr = []byte(" DO NOT EDIT.")
+	genHdr           = []byte("// Code generated ")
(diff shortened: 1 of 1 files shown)


bhcleek:
Confirmed that this fixes the problems that I've reported. Thank you 🙇

PR #1046, opened 2024-09-21
fix: enforce-repeated-arg-type-style rule finds false positives in case of invalid types

### LINKED ISSUE #1032: The `enforce-repeated-arg-type-style` rule finds some false positives.
**Describe the bug**
The `enforce-repeated-arg-type-style` rule finds some false positives. I will demonstrate this using the repository https://github.com/influxdata/telegraf (over 650k lines) as an example.

**To Reproduce**
2. I cloned repo and switched to particular branch
```
git clone https://github.com/influxdata/telegraf.git
cd telegraf
git checkout v1.31.3
```
```shell
# flags
revive -config revive.toml ./... | sort
```
```toml
#revive.toml
[rule.enforce-repeated-arg-type-style]
  arguments = ["short"]
```
4. Some of the findings are false positives.

**Expected behavior**
I would like only the findings that are not false positives to be shown.

**Logs**
Some examples:

`plugins/inputs/webhooks/artifactory/artifactory_webhook.go:23:87: repeated argument type can be omitted`:
image

`plugins/outputs/bigquery/bigquery.go:244:45: repeated argument type can be omitted`:
image

`plugins/outputs/mqtt/mqtt_test.go:803:59: repeated argument type can be omitted`:
image

 - OS: Ubuntu 23:10
 - Version of Go:
```shell
$ go version 
go version go1.23.0 linux/amd64
```

denisvmedia:
closes #1032

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -389,6 +389,9 @@ declaration. The 'short' style encourages omitting repeated types for concisenes
 whereas the 'full' style mandates explicitly stating the type for each argument
 and return value, even if they are repeated, promoting clarity.
 
+_IMPORTANT_: When `short` style is used, the rule will not flag the arguments that use
+imported types. This is because the rule cannot efficiently determine the imported type.
+
 _Configuration (1)_: (string) as a single string, it configures both argument
 and return value styles. Accepts 'any', 'short', or 'full' (default: 'any').
 
--- a/rule/enforce-repeated-arg-type-style.go
+++ b/rule/enforce-repeated-arg-type-style.go
@@ -135,12 +135,8 @@ func (r *EnforceRepeatedArgTypeStyleRule) Apply(file *lint.File, arguments lint.
 				var prevType ast.Expr
 				if fn.Type.Params != nil {
 					for _, field := range fn.Type.Params.List {
-						if r.isInvalidType(typesInfo.Types[field.Type].Type) {
-							// TODO: In theory, we could have compared raw import names (import package alias + selector), but will it work properly in all the cases?
-							continue
-						}
-
-						if types.Identical(typesInfo.Types[field.Type].Type, typesInfo.Types[prevType].Type) {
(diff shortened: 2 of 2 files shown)


zak-pawel:
@denisvmedia Thanks! I tested code from this PR and it fixes most of the cases.

However, it still sees a problem for such cases (for `op`):
```
func printConfig(name string, p telegraf.PluginDescriber, op string, commented bool, di telegraf.DeprecationInfo, outputBuffer io.Writer) {
```
But when I change the second argument from `telegraf.PluginDescriber` to `int`, the function stops being problematic:
```
func printConfig(name string, p int, op string, commented bool, di telegraf.DeprecationInfo, outputBuffer io.Writer) {
```

There is a similar problem with return types.

denisvmedia:
Yeah, that was my oversight. I will fix it. Thanks!

denisvmedia:
@zak-pawel does it work for you now?

zak-pawel (approved):
Now it works very well, thanks!

chavacava:
Hi @denisvmedia thanks for the PR, it LGTM.
Could you please add some test cases to the rule tests?

denisvmedia:
@chavacava it's actually non-trivial to test it. We'll have to provide the means to pass fake imports in our test engine. Otherwise, any package we import won't be found and the tests will start failing. I played around with a custom (fake) importer, but I didn't have enough time to simulate the invalid package situation. I can try checking it later. Do you think we can merge this PR now without having the tests at the moment?

chavacava:
IMO relying on actual type analysis is over-killing for such a simple rule.
I wonder if just comparing the string representation of parameters types is not enough.
A canonical string representation of parameters' types can be obtained by calling `gofmt` on each `*Field.Type` of  `FuncDecl.Type.Params.List`.

denisvmedia:
That's what my TODO comment actually says (maybe I wasn't clear enough in it ;)). But what I wanted is to have a quick fix to make sure that this rule at least doesn't have false positives (it seems to me to be better than having false negatives), and then invest some more time to have a full fix. But if you think it makes no sense to merge it as is, I can revise the PR. I'm just not sure when I will have time to play with it.

Issue #1045, opened 2024-09-20
How to disable `rule.exported` for consts?


Potherca:
## TL;DR:

I am aware that the goal of this project is no longer to be a 100% drop-in replacement for golint (which I think is fair) but is there a way for me to disable the `exported` rule but only for consts?

## Full Details

At [pipeline-components](https://pipeline-components.dev/) we have a component for [go-lint](https://gitlab.com/pipeline-components/go-lint). Although it is not one of our larger components<sup>\[1]</sup>, I'm thinking of prolonging support by replacing `golint` with revive.

However, the current claim that the default behaviour is the same as `golint` does not appear to be 100% accurate.

To be specific, I am seeing the following warnings from revive that are not there with golint:

1. "should have a package comment"
2. "this block is empty, you can remove it"
3. "comment on exported const [..] should be of the form [..]"

The first two I can resolve by tweaking the default config,<sup>\[2]</sup>, which is fine for our purposes (as golint does not have a config, so we can just add one for revive).

But I have not yet been able to resolve the `comment on exported const [..] should be of the form [..]`. Please note, this is explicitly about exported "const", not other exported types.

How could this be achieved?

---

<sup>\[1]</sup> Currently it only has +/- 40K pulls on DockerHub ![][1]<br>
<details><summary><sup>[2]</sup> Click to see config</summary>

```toml
ignoreGeneratedHeader = false
severity = "warning"
confidence = 0.8
errorCode = 0
warningCode = 0

[rule.blank-imports]
[rule.context-as-argument]
[rule.context-keys-type]
[rule.dot-imports]
[rule.empty-block]
    Disabled = true
[rule.errorf]
[rule.error-naming]
[rule.error-return]
[rule.error-strings]
[rule.exported]
[rule.increment-decrement]
[rule.indent-error-flow]
[rule.package-comments]
     Disabled = true
[rule.range]
[rule.receiver-naming]
[rule.redefines-builtin-id]
[rule.superfluous-else]
[rule.time-naming]
[rule.unexported-return]
[rule.unreachable-code]
[rule.unused-parameter]
[rule.var-declaration]
[rule.var-naming]
```

</details>

[1]: https://img.shields.io/docker/pulls/pipelinecomponents/go-lint.png

chavacava:
Hi @Potherca, I'm not sure to understand your request on linting comments on public constants.
For the file `foo.go` 
```go
package main

// this is not ok
const AConstant = 10
```

[...]

PR #1047, opened 2024-09-22
rule.exported: add feature to disable selected warnings for rule (const,method,function)

mfederowicz:
related with #1045 

for rule.exported add feature to read array/list of types which we want to ignore in processing time

```
[rule.exported]                                                                                                                                                            
  ▎ arguments = [
"checkPrivateReceivers", 
["function","method","type","var"]
]
```
of course thouse are types discovered when rule.exported is enabled in config, if we set some other additional flags outside of that list it can results with panic/error


test/exported_test.go
```diff
@@ -30,3 +30,10 @@ func TestCheckPublicInterfaceOption(t *testing.T) {
 
 	testRule(t, "exported-issue-1002", &rule.ExportedRule{}, &lint.RuleConfig{Arguments: args})
 }
+
+func TestExportedDisableKindWarnings(t *testing.T) {
+
+	args := []any{[]interface{}{"const", "function", "method", "type"}}
```
denisvmedia, line 36: nit:

```suggestion
	args := []any{[]any{"const", "function", "method", "type"}}
```

zak-pawel:
It would be very nice to see this feature merged. I would especially need turning off warnings for types.

mfederowicz:
hi @denisvmedia I resolved conflicts from master, and if you have and sugestions  (changes) let me know :)

chavacava:
Hi @zak-pawel could you please elaborate on the reasons for turning off warnings for types? (I've quickly read the thread on telegraph repo)

RULES_DESCRIPTIONS.md
```diff
@@ -475,12 +475,13 @@ Available flags are:
...
 * _sayRepetitiveInsteadOfStutters_ replaces the use of the term _stutters_ by _repetitive_ in failure messages
 * _checkPublicInterface_ enabled checking public method definitions in public interface types
+* []string list of code types to ignore warnings, avaliable options: function, method, type and var
 
 Example:
 
 ```toml
 [rule.exported]
-  arguments = ["checkPrivateReceivers", "disableStutteringCheck", "checkPublicInterface"]
+  arguments = ["checkPrivateReceivers", "disableStutteringCheck", "checkPublicInterface", ["function","method","type","var"]]
```
chavacava, line 484: the configuration entry gives no clue about the "ignore" nature of the list.

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
 	"github.com/mgechev/revive/lint"
 )
 
+// IgnoredWarnings store ignored warnings types
+type IgnoredWarnings struct {
```
chavacava, line 18: No need to be public

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
+	Const    bool
+	Function bool
+	Method   bool
+	Type     bool
+	Var      bool
+}
+
+// Name returns the rule name.
```
chavacava, line 26: Comment is not related with this function

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
+func (i *IgnoredWarnings) isDisabled(t string) bool {
+
```
chavacava, line 28:
```suggestion
```

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
+type IgnoredWarnings struct {
+	Const    bool
+	Function bool
+	Method   bool
+	Type     bool
+	Var      bool
+}
+
+// Name returns the rule name.
+func (i *IgnoredWarnings) isDisabled(t string) bool {
```
chavacava, line 27: what is `t`?

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
 	checkPublicInterface   bool
 	stuttersMsg            string
+	disabledWarningsTypes  IgnoredWarnings
 	sync.Mutex
 }
 
 func (r *ExportedRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
 	if !r.configured {
```
chavacava, line 60:
```suggestion
	if r.configured {
		return // already configured
	}
	// ...
```

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
+					}
+				}
+
 			default:
-				panic(fmt.Sprintf("Unknown configuration flag %s for %s rule", flagStr, r.Name()))
+				panic(fmt.Sprintf("Unknown configuration flag type %s for %s rule", reflect.TypeOf(flag), r.Name()))
 			}
+
 		}
 		r.configured = true
```
chavacava, line 108: move to be just after the `if r.configured {` block

test/exported_test.go
```diff
@@ -30,3 +30,10 @@ func TestCheckPublicInterfaceOption(t *testing.T) {
 
 	testRule(t, "exported-issue-1002", &rule.ExportedRule{}, &lint.RuleConfig{Arguments: args})
 }
+
+func TestExportedDisableKindWarnings(t *testing.T) {
+
```
chavacava, line 35:
```suggestion
```

test/exported_test.go
```diff
@@ -30,3 +30,10 @@ func TestCheckPublicInterfaceOption(t *testing.T) {
...
+	args := []any{[]any{"const", "function", "method", "type"}}
+	testRule(t, "exported-issue-1045", &rule.ExportedRule{}, &lint.RuleConfig{Arguments: args})
+
```
chavacava, line 38:
```suggestion
```

rule/exported.go
```diff
@@ -13,38 +14,96 @@ import (
...
+	case "function":
+		return i.Function
+	case "method":
+		return i.Method
+	case "type":
+		return i.Type
+	default:
+		panic(fmt.Sprintf("Unknown ignore warning flag %s", t))
+	}
+
```
chavacava, line 43:
```suggestion
```

chavacava (changes requested):
I've left some comments on the implementation details but, in general, IMO the semantics of the new configuration is not clear or, not righlty implemented.
For example, from the configuration documentation we understand that setting "type" will avoid generating *any* warning on types at all. Then, by reading the implementation we can see that some warning on types will be generated (those of the form `"exported type %v should have comment or be unexported"`)

RULES_DESCRIPTIONS.md
mfederowicz: ok i try struct with dedicated fields but in toml there is problem with notation like:

``
IgnoredWarnings{field:true,field2:true}
``

you can only use like this:

``
{field:true,field2:true} or list ["val","val2","val3"]
``

but there is no option to name/prefix that (or maybe I dont know the magic yet :()

rule/exported.go
mfederowicz: changed comment

rule/exported.go
mfederowicz: ok changed var name to kind

test/exported_test.go
mfederowicz: ok test removed

rule/exported.go
mfederowicz: removed default, but add return false after switch (by default selected kind/type is not disabled so warning can be printed )

RULES_DESCRIPTIONS.md
srebhan: I think what @chavacava is asking is to have a better naming of the setting, e.g. `disableFunctionDocumentationCheck` instead of `function`...

srebhan:
@chavacava let me elaborate a bit on
> Hi @zak-pawel could you please elaborate on the reasons for turning off warnings for types? (I've quickly read the thread on telegraph repo)

Telegraf has a "plugin-like" structure, where each "plugin" fulfills a set of interfaces and by now we exported the plugin struct as well as the interface functions by convention. We now want to use the checker to make sure each plugin _only_ exports the plugin type and the interface functions! However, each of the plugin types has to implement certain functions like `Gather()` for inputs and we now do have 200+ plugins all implementing `Gather()`. Guess what the comment for this function is in 99.9999% of all cases? `// Gather` or `Gather gathers data`. There is no value in adding these kind of comments. Similar for the plugin struct... Currently, we can only keep the `exported` rule disabled or add churn to our 300+ plugins... Therefore we would prefer to disable those checks. :-)

chavacava:
Hi @srebhan thank you for the details on your use case. 
As I understand, the "zero-value added" comments problem concerns the code related with plugins. If my understanding is right you ask for the ability to disabling certain warnings of the rule. By doing that you will mute unnecessary warnings on plugins code but... it will also disable warnings on all the code base (and, for example, it will not warn on [malformed comments](https://github.com/influxdata/telegraf/blob/f7d0dc7b8785edd3e7730d0df5330903adc06984/internal/snmp/field.go#L96-L97) or undocumented [public methods](https://github.com/influxdata/telegraf/blob/f7d0dc7b8785edd3e7730d0df5330903adc06984/agent/accumulator.go#L122)). Is that what you need?
Is disabling the rule as per (plugin) file not an option?

RULES_DESCRIPTIONS.md
mfederowicz: > I think what @chavacava is asking is to have a better naming of the setting, e.g. `disableFunctionDocumentationCheck` instead of `function`...

ok so each element of ["function","method","type","var"] as dedicated flag 

arguments = ["disableFunctionCheck", "disableMethodCheck", "disableTypeCheck", "disableVarCheck"]

that was my first idea, but thinks using array/slice will be simpler (one place to modify) :)

mfederowicz: ok maybe we can walk that way:

arguments = ["disableFunctionCheck", "disableMethodCheck", "disableTypeCheck", "disableVarCheck"]

and of course modify my commited code to use that flags insted of additional array in config what you think @chavacava ?

srebhan:
@chavacava we are currently _solely_ checking the plugins with this rule. Disabling the whole rule defeats the purpose as we actually _want_ the exported rule to warn about exported structs and functions only whitelisting the _known_ interface functions. This way the rule will warn about any unexpected exports which is exactly what we want. We only need to shut-up the comment warnings for the plugins...

chavacava:
@zak-pawel  @srebhan 
I've made some refactoring in the PR, please check if the rule+configuration behave as you expect for your use case

cc: @mfederowicz

zak-pawel:
@chavacava I tested it on the Telegraf code, and it looks like the individual switches correctly disable the respective checks. Thanks!

chavacava (approved):

chavacava:
Thanks for the PR @mfederowicz

mfederowicz:
@chavacava  no problem, but forgot to add var testdata, created quick update: https://github.com/mgechev/revive/pull/1055 :P

PR #1048, opened 2024-09-22
adds check for receiver names length

### LINKED ISSUE #1040: Add the receiver name length check to `rule.receiver-naming`
I would like to check the length of receiver variable names.

Add the receiver variable name length check to the `receiver-naming` rule.

I created a [linter](https://github.com/ytnsym/receivernamelen), but it seemed like it would be better to improve the `receiver-naming` rule.

**Additional context**
https://github.com/golangci/golangci-lint/pull/5007#issuecomment-2350457699

chavacava:
Closes #1040 
The rule receiver-naming now accepts an optional argument indicating the maximum length of receiver names. If no argument is provided the rule does not check for names length.

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -737,7 +737,12 @@ _Configuration_: N/A
 
 _Description_: By convention, receiver names in a method should reflect their identity. For example, if the receiver is of type `Parts`, `p` is an adequate name for it. Contrary to other languages, it is not idiomatic to name receivers as `this` or `self`.
 
-_Configuration_: N/A
+_Configuration_: (optional) a string indicating the maximum length of the receiver name. If not set, name length is not checked.
+
+```toml
+[rule.receiver-naming]
+    arguments=["max-length=2"]
+```
 
 ## redefines-builtin-id
 
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -737,11 +737,13 @@ _Configuration_: N/A
 
 _Description_: By convention, receiver names in a method should reflect their identity. For example, if the receiver is of type `Parts`, `p` is an adequate name for it. Contrary to other languages, it is not idiomatic to name receivers as `this` or `self`.
 
-_Configuration_: (optional) a string indicating the maximum length of the receiver name. If not set, name length is not checked.
+_Configuration_: (optional) list of key-value-pair-map (`[]map[string]any`).
+
+- `maxLength` : (int) max length of receiver name
 
 ```toml
 [rule.receiver-naming]
-    arguments=["max-length=2"]
+    arguments = [{maxLength=2}]
 ```
 
 ## redefines-builtin-id
--- a/rule/receiver-naming.go
+++ b/rule/receiver-naming.go
@@ -3,8 +3,6 @@ package rule
 import (
 	"fmt"
 	"go/ast"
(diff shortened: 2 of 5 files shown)


rule/receiver-naming.go
```diff
@@ -3,24 +3,69 @@ package rule
...
+			var err error
+			r.receiverNameMaxLength, err = strconv.Atoi(value)
+			if err != nil {
+				panic(fmt.Sprintf("Invalid value %s for the configuration key max-length, expected integer value: %v", value, err))
+			}
+		default:
+			panic(fmt.Sprintf("Unknown configuration key %s for %s rule.", key, r.Name()))
+		}
+	}
+}
```
ccoVeille, line 54: That's a function wide `if` statement

I would suggest this instead

```suggestion
func (r *ReceiverNamingRule) configure(arguments lint.Arguments) {
	r.Lock()
	defer r.Unlock()
	if r.receiverNameMaxLength > 0 {
		return
	}

	if len(arguments) < 1 {
		r.receiverNameMaxLength = defaultReceiverNameMaxLength
		return
	}
	arg := arguments[0]
	argStr, ok := arg.(string)
	if !ok {
		panic(fmt.Sprintf("Invalid argument for %s rule. Expecting a string, got %T", r.Name(), arg))
	}

	parts := strings.Split(argStr, "=")
	if len(parts) != 2 {
		panic(fmt.Sprintf("Invalid argument for %s rule. Expecting a string of the form 'key=value', got %s", r.Name(), argStr))
	}

	key := strings.TrimSpace(parts[0])
	value := strings.TrimSpace(parts[1])
	switch key {
	case "max-length":
		var err error
		r.receiverNameMaxLength, err = strconv.Atoi(value)
		if err != nil {
			panic(fmt.Sprintf("Invalid value %s for the configuration key max-length, expected integer value: %v", value, err))
		}
	default:
		panic(fmt.Sprintf("Unknown configuration key %s for %s rule.", key, r.Name()))
	}
}
```

Also I fixed the typo `an string` twice

RULES_DESCRIPTIONS.md
```diff
@@ -737,7 +737,12 @@ _Configuration_: N/A
 
 _Description_: By convention, receiver names in a method should reflect their identity. For example, if the receiver is of type `Parts`, `p` is an adequate name for it. Contrary to other languages, it is not idiomatic to name receivers as `this` or `self`.
 
-_Configuration_: N/A
+_Configuration_: (optional) a string indicating the maximum length of the receiver name. If not set, name length is not checked.
+
+```toml
+[rule.receiver-naming]
+    arguments=["max-length=2"]
```
denisvmedia, line 744: Why did you decide to have this way of configuration? This seems to be different comparing g to what we have in the other rules. We either have ints directly or we have named args as in unchecked-type-assertion.

denisvmedia (approved):
LGTM

rule/receiver-naming.go
```diff
@@ -66,6 +107,17 @@ func (w lintReceiverName) Visit(n ast.Node) ast.Visitor {
 		})
 		return w
 	}
+
+	if w.receiverNameMaxLength > 0 && len(name) > w.receiverNameMaxLength {
```
ytnsym, line 111: I thought it would be better to use rune.
```suggestion
	if w.receiverNameMaxLength > 0 && len([]rune(name)) > w.receiverNameMaxLength {
```

PR #1053, opened 2024-09-29
code cleanup

chavacava:
Little code refactor with the goal of making code more go idiomatic and easy to understand.

### DIFF
--- a/rule/max-public-structs.go
+++ b/rule/max-public-structs.go
@@ -19,20 +19,23 @@ const defaultMaxPublicStructs = 5
 func (r *MaxPublicStructsRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
-	if r.max < 1 {
-		if len(arguments) < 1 {
-			r.max = defaultMaxPublicStructs
-			return
-		}
-
-		checkNumberOfArguments(1, arguments, r.Name())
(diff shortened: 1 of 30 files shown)


rule/max-public-structs.go
```diff
@@ -19,20 +19,22 @@ const defaultMaxPublicStructs = 5
...
+	}
 
-		max, ok := arguments[0].(int64) // Alt. non panicking version
-		if !ok {
-			panic(`invalid value passed as argument number to the "max-public-structs" rule`)
-		}
-		r.max = max
+	checkNumberOfArguments(1, arguments, r.Name())
+
+	max, ok := arguments[0].(int64) // Alt. non panicking version
```
denisvmedia, line 33: Note, max here will override a built-in.

chavacava: right, I've created this branch before the last merged PR #1052, I will rebase once I've refactored all rules

denisvmedia:
There is some issue with 1.21 tests, but other than that I really like the way how you refactored the code.

chavacava:
> There is some issue with 1.21 tests, but other than that I really like the way how you refactored the code.

Testing step fails before actually running the test (some problem at the preparation of the environment) I'll check what is wrong

Issue #1056, opened 2024-10-04
Exclude vendor/... by default


mcandre:
**Is your feature request related to a problem? Please describe.**

revive's default behavior with `revive ./...` generates linter warnings for third party code in the `vendor` directory, when using the `go mod vendor` system to manage dependencies.

Linter warnings for third party code tends to be less actionable compared to linter warnings for first party code.

**Describe the solution you'd like**

Exclude `vendor/...` by default.

**Describe alternatives you've considered**

Baking `-exclude vendor/...` into mcand/remage-extras's Revive task.

mcandre:
Yes, I see a configuration file and command line flag to do this.

I'm asking for better default behavior. The vendor directory is a Go programming language standard.

mfederowicz:
@chavacava my #1058 adds vendor/... to ignored only if we don set -exclude from cli or exclude = [] in revive.toml what you think?

chavacava:
@mfederowicz thanks gor the PR. My concern with by defaylt excluding vendor is that revive will not behave as before thus vendor will be *silently* excluded.

About the PR: What if I actually want to analyze vendor? How I need to run revive?

mfederowicz:
ofcourse @chavacava iam open to other options (maybe dedicated flag in config set by default to true) or something else :)

chavacava:
@mfederowicz IMO the hack of excluding a non existent pattern rather than a solution is a smell on the inconvenience of excluding vendor by default. The same on adding a flag to actually (un)do what another already existing flag does.

@mcandre 
As you already know, there is a flag to exclude vendor (or any other arbitrary directory). Use it if you are not interested in analyzing vendored code.

I'll close this issue for now, if in the future there is more people interested in changes on the treatment of vendor we can re-open the discussion.

mcandre:
This is the first Go linter in recent memory to not exclude the vendor directory properly by default.

denisvmedia:
#1058 is merged.

mcandre:
Amazing!

When can we expect a new release to include this change?

mfederowicz:
@mcandre if you dont want to wait for new release you can install dev version of revive by:

```go install github.com/mgechev/revive@master```

mcandre:
Fair point. @master or @latest would suffice if I need to rush.

However, I prefer to target stable, tagged version releases as a safeguard against potentially breaking changes.

PR #1058, opened 2024-10-06
feat: ignore vendor if no exclude set

mfederowicz:
related with #1056 

by default if we dont have any excludes, we can ignore vendor/...


mgechev (approved):

mgechev:
Thanks for the PR!

PR #1080, opened 2024-10-26
chore: remove unnecessary empty lines

alexandear:
The PR removes empty lines that don't add to readability.


chavacava:
Thanks @alexandear

PR #1091, opened 2024-10-30
adds check on filenames to banned-characters rule

### LINKED ISSUE #1090: Warn on filenames with non ASCII characters
As described [here](https://www.arp242.net/jia-tan-go.html), source files with names containing non-ASCII characters can be used to compromise a code base. 
Enforcing ASCII-only filenames helps preventing these kind of attacks. `revive` is the perfect place to implement such enforcement. 

There is already a `banned-characters` rule that helps mitigating some security risks, it seems a good candidate where to implement the checks on filenames.

chavacava:
Closes #1090 by adding filename checks to the `banned-characters` rule.

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -126,7 +126,7 @@ _Configuration_: N/A
 
 ## banned-characters
 
-_Description_: Checks given banned characters in identifiers(func, var, const). Comments are not checked.
+_Description_: Checks given banned characters in identifiers (func, var, const). Comments are not checked. The rule also warns on filenames with non-ASCII characters; this helps preventing attacks like the one described [here](https://www.arp242.net/jia-tan-go.html).
 
 _Configuration_: This rule requires a slice of strings, the characters to ban.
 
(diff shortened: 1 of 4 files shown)


denisvmedia (approved):
Lgtm 👍

ccoVeille (changes requested):
For me, it should be a separate rule

The banned characters list won't be used for the file names

People might want to add other rules by extending the rules such as no space, no hypen, lowercase only, to respect the Go convention

Linting files is a good thing, but it should be separate.

chavacava:
@ccoVeille I've created the PR #1092 that adds the rule.

The rule, of course, accepts a regular expression. Like that, it anticipates people clamor for regular expressions everywhere. 

> to respect the Go convention

there is not such a convention. Check https://github.com/golang/go/issues/36060

ccoVeille:
There is indeed no convention, but a common practice inherited from package name convention.

Anyway it's great you opened the other PR

But then, shouldn't this one be simply closed?

chavacava:
replaced by #1092

PR #1114, opened 2024-11-12
fix: fixes #1103 by fixing the processing of max-public-structs arguments

### LINKED ISSUE #1103: The `max-public-structs` rule starts finding many false positives in version `1.5.0`
**Describe the bug**
The `max-public-structs` rule starts finding many false positives in version `1.5.0`.
Version `1.4.0`, with the same configuration, doesn't find any findings.
I will demonstrate this using the repository https://github.com/influxdata/telegraf (over 650k lines) as an example.

**To Reproduce**
2. I cloned repo and switched to particular branch
```
git clone https://github.com/influxdata/telegraf.git
cd telegraf
git checkout v1.32.2
```
```shell
# flags
revive -config revive.toml ./... | sort
```
```toml
#revive.toml
[rule.max-public-structs]
  arguments = [14]
```
4. There are a `632` false positives.

**Expected behavior**
I would like there to be no findings for the given repo, branch and configuration - just as it was for the previous version of `revive` (`1.4.0`)."

**Logs**
All findings attached: 
[revive_1.5.0_findings_for_telegraf.txt](https://github.com/user-attachments/files/17686531/revive_1.5.0_findings_for_telegraf.txt)

 - OS: Ubuntu 23:10
 - Version of Go:
```shell
$ go version 
go version go1.23.3 linux/amd64
```

chavacava:
Closes #1103

### DIFF
--- a/rule/max_public_structs.go
+++ b/rule/max_public_structs.go
@@ -44,6 +44,10 @@ func (r *MaxPublicStructsRule) Apply(file *lint.File, arguments lint.Arguments) 
 
 	var failures []lint.Failure
 
+	if r.max < 1 {
+		return failures
+	}
+
 	fileAst := file.AST
 
 	walker := &lintMaxPublicStructs{
--- a/rule/max_public_structs.go
+++ b/rule/max_public_structs.go
@@ -1,6 +1,7 @@
 package rule
 
 import (
+	"fmt"
 	"go/ast"
 	"strings"
 	"sync"
@@ -19,7 +20,7 @@ const defaultMaxPublicStructs = 5
 func (r *MaxPublicStructsRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
-	if r.max == 0 {
+	if r.max != 0 {
 		return // already configured
 	}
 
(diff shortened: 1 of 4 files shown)


ccoVeille (changes requested):
Typo in commit message and PR title

plublic => public

rule/max_public_structs.go
```diff
@@ -19,7 +20,7 @@ const defaultMaxPublicStructs = 5
 func (r *MaxPublicStructsRule) configure(arguments lint.Arguments) {
 	r.Lock()
 	defer r.Unlock()
-	if r.max == 0 {
+	if r.max != 0 {
```
ccoVeille, line 23: I'm worried by what the code will do with maxstruct set to 0

configure will be called again and again,no?

I mean someone who want to disable the rule, like what have been done a few days ago with file-max-lines

rule/max_public_structs.go
```diff
@@ -56,7 +57,7 @@ func (r *MaxPublicStructsRule) Apply(file *lint.File, arguments lint.Arguments)
 
 	if walker.current > r.max {
 		walker.onFailure(lint.Failure{
-			Failure:    "you have exceeded the maximum number of public struct declarations",
+			Failure:    fmt.Sprintf("you have exceeded the maximum number (%d) of public struct declarations", r.max),
```
ccoVeille, line 60: Also, what happen if max < 0, like -1

Number of struct will always be positive, do bigger than -1

Also, what is the point of reporting an error about something bigger than -1 in the error message

Your changes are not exactly about that, but I thought about it by reading the code

PR #1126, opened 2024-11-15
refactor: replace panic with error in rules

mfederowicz:
related with  PR #1107 

changes in areas: 
- refactor rules to return error in r.configure func
- make changes in almost all rules
- add return error for Apply func to handle error in upper level (bacause there were panic outisde configure)
- and other places where panic was used
- add comment for package rule :)

I am open to suggestion what to change in next steps @chavacava @denisvmedia


lint/file.go
```diff
@@ -96,7 +96,7 @@ func (f *File) isMain() bool {
 
 const directiveSpecifyDisableReason = "specify-disable-reason"
 
-func (f *File) lint(rules []Rule, config Config, failures chan Failure) {
+func (f *File) lint(rules []Rule, config Config, failures chan Failure) error {
```
chavacava, line 99: Once this method returns an error, what to do with it? (as is, the caller just ignores the error)

mfederowicz: that is good point what next, we should handle that here:

```pkg.lint(ruleSet, config, failures)``` and return error and it should finish in:

```
if err := l.lintPackage(pkg, gover, ruleSet, config, failures); err != nil {                                                                                   
  fmt.Fprintln(os.Stderr, err)                                                                                                                               
  os.Exit(1)                                                                                                                                                 
 }
 ```

chavacava: Do we exit at `pkg.lint` or should we propagate the error up to the main function and exit there?

chavacava:
As is, the implementation of rules in this branch does not handles errors but put them "under the carpet"
Almost the single situation where rules might err is when reading arguments, therefore `rule.Apply`  implementations  shall handle errors returned by calls to `r.configure`

lint/file.go
mfederowicz: we should handle pkg.ling err and return because level up is:

```
if err := l.lintPackage(pkg, gover, ruleSet, config, failures); err != nil {                                                                                   
  fmt.Fprintln(os.Stderr, err)                                                                                                                               
  os.Exit(1)                                                                                                                                                 
 }
 ```

Comment (ccoVeille) [reply lint/file.go:99]: I agree with this change. Panic should be avoided

File: lint/package.go [hunk 2]
```diff
@@ -182,17 +183,41 @@ func (p *Package) scanSortable() {
...
 }
 
-func (p *Package) lint(rules []Rule, config Config, failures chan Failure) {
+func (p *Package) lint(rules []Rule, config Config, failures chan Failure) error {
 	p.scanSortable()
 	var wg sync.WaitGroup
+	errCh := make(chan error, 1) // Buffered channel to signal the first error
+    _, cancel := context.WithCancel(context.Background())
+
+    defer cancel() // Ensure the context is canceled at the end
```
ccoVeille, line 192: Using a cancelable context only make sense if you use a context.

So you should remove this
```suggestion
```
And [this](https://github.com/mgechev/revive/pull/1126#discussion_r1844984529)

lint/package.go
```diff
@@ -182,17 +183,41 @@ func (p *Package) scanSortable() {
...
 	for _, file := range p.files {
 		wg.Add(1)
 		go (func(file *File) {
-			file.lint(rules, config, failures)
-			wg.Done()
+			defer wg.Done()
+			if err := file.lint(rules, config, failures); err != nil {
+                select {
+                case errCh <- err: // Send the error to the channel if it's empty
+                    cancel() // Signal other goroutines to stop
```
ccoVeille, line 201: See [previous comment](https://github.com/mgechev/revive/pull/1126#discussion_r1844984367)


```suggestion
```

rule/argument_limit.go
```diff
@@ -16,27 +17,28 @@ type ArgumentsLimitRule struct {
...
 	if len(arguments) < 1 {
 		r.max = defaultArgumentsLimit
-		return
+		return nil
 	}
 
 	maxArguments, ok := arguments[0].(int64) // Alt. non panicking version
 	if !ok {
-		panic(`invalid value passed as argument number to the "argument-limit" rule`)
+		return fmt.Errorf(`invalid value passed as argument number to the "argument-limit" rule`)
```
ccoVeille, line 34:
```suggestion
		return errors.New(`invalid value passed as argument number to the "argument-limit" rule`)
```

rule/enforce_map_style.go
```diff
@@ -44,39 +45,41 @@ type EnforceMapStyleRule struct {
...
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument '%v' for 'enforce-map-style' rule. Expecting string, got %T", arguments[0], arguments[0]))
+		return fmt.Errorf("Invalid argument '%v' for 'enforce-map-style' rule. Expecting string, got %T", arguments[0], arguments[0])
 	}
 
 	var err error
 	r.enforceMapStyle, err = mapStyleFromString(enforceMapStyle)
 	if err != nil {
-		panic(fmt.Sprintf("Invalid argument to the enforce-map-style rule: %v", err))
+		return fmt.Errorf("Invalid argument to the enforce-map-style rule: %v", err)
```
ccoVeille, line 70:
```suggestion
		return fmt.Errorf("Invalid argument to the enforce-map-style rule: %w", err)
```

rule/add_constant.go
```diff
@@ -229,35 +230,35 @@ func (r *AddConstantRule) configure(arguments lint.Arguments) {
...
 					if !ok {
-						panic(fmt.Sprintf("Invalid argument to the ignoreFuncs parameter of add-constant rule, string expected. Got '%v' (%T)", v, v))
+						return fmt.Errorf("Invalid argument to the ignoreFuncs parameter of add-constant rule, string expected. Got '%v' (%T)", v, v)
 					}
 
 					for _, exclude := range strings.Split(excludes, ",") {
 						exclude = strings.Trim(exclude, " ")
 						if exclude == "" {
-							panic("Invalid argument to the ignoreFuncs parameter of add-constant rule, expected regular expression must not be empty.")
+							return fmt.Errorf("Invalid argument to the ignoreFuncs parameter of add-constant rule, expected regular expression must not be empty")
```
ccoVeille, line 256:
```suggestion
							return errors.New("Invalid argument to the ignoreFuncs parameter of add-constant rule, expected regular expression must not be empty")
```

rule/add_constant.go
```diff
@@ -229,35 +230,35 @@ func (r *AddConstantRule) configure(arguments lint.Arguments) {
...
 						}
 
 						exp, err := regexp.Compile(exclude)
 						if err != nil {
-							panic(fmt.Sprintf("Invalid argument to the ignoreFuncs parameter of add-constant rule: regexp %q does not compile: %v", exclude, err))
+							return fmt.Errorf(fmt.Sprintf("Invalid argument to the ignoreFuncs parameter of add-constant rule: regexp %q does not compile: %v", exclude, err))
```
ccoVeille, line 261:
```suggestion
							return fmt.Errorf(fmt.Sprintf("Invalid argument to the ignoreFuncs parameter of add-constant rule: regexp %q does not compile: %w", exclude, err))
```

rule/enforce_repeated_arg_type_style.go
```diff
@@ -35,7 +36,7 @@ func repeatedArgTypeStyleFromString(s string) enforceRepeatedArgTypeStyleType {
 			},
 		)
 
-		panic(fmt.Sprintf("Invalid argument to the enforce-repeated-arg-type-style rule: %v", err))
+		return "", fmt.Errorf("Invalid argument to the enforce-repeated-arg-type-style rule: %v", err)
```
ccoVeille, line 39:
```suggestion
		return "", fmt.Errorf("Invalid argument to the enforce-repeated-arg-type-style rule: %w", err)
```

rule/enforce_slice_style.go
```diff
@@ -28,7 +29,7 @@ func sliceStyleFromString(s string) (enforceSliceStyleType, error) {
 	case string(enforceSliceStyleTypeNil):
 		return enforceSliceStyleTypeNil, nil
 	default:
-		return enforceSliceStyleTypeAny, fmt.Errorf(
+		return "", fmt.Errorf(
```
ccoVeille, line 32: Why this change?

rule/enforce_slice_style.go
```diff
@@ -48,39 +49,40 @@ type EnforceSliceStyleRule struct {
...
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument '%v' for 'enforce-slice-style' rule. Expecting string, got %T", arguments[0], arguments[0]))
+		return fmt.Errorf("Invalid argument '%v' for 'enforce-slice-style' rule. Expecting string, got %T", arguments[0], arguments[0])
 	}
 
 	var err error
 	r.enforceSliceStyle, err = sliceStyleFromString(enforceSliceStyle)
 	if err != nil {
-		panic(fmt.Sprintf("Invalid argument to the enforce-slice-style rule: %v", err))
+		return fmt.Errorf("Invalid argument to the enforce-slice-style rule: %v", err)
```
ccoVeille, line 74:
```suggestion
		return fmt.Errorf("Invalid argument to the enforce-slice-style rule: %w", err)
```

rule/file_header.go
```diff
@@ -19,30 +20,31 @@ var (
...
 	if len(arguments) < 1 {
-		return
+		return nil
 	}
 
 	var ok bool
 	r.header, ok = arguments[0].(string)
 	if !ok {
-		panic(fmt.Sprintf("invalid argument for \"file-header\" rule: argument should be a string, got %T", arguments[0]))
+		return fmt.Errorf("invalid argument for \"file-header\" rule: argument should be a string, got %T", arguments[0])
```
ccoVeille, line 37: This could be simplified

```suggestion
		return fmt.Errorf(`invalid argument for "file-header" rule: argument should be a string, got %T`, arguments[0])
```

rule/file_length_limit.go
```diff
@@ -43,7 +44,7 @@ func (r *FileLengthLimitRule) Apply(file *lint.File, arguments lint.Arguments) [
 	}
 
 	if err := scanner.Err(); err != nil {
-		panic(err.Error())
+		return nil, fmt.Errorf(err.Error())
```
ccoVeille, line 47: 😱

```suggestion
		return nil, err
```

rule/filename_format.go
```diff
@@ -54,34 +55,36 @@ func (*FilenameFormatRule) Name() string {
...
 	str, ok := arg.(string)
 	if !ok {
-		panic(fmt.Sprintf("rule %q expects a string argument, got %v of type %T", r.Name(), arg, arg))
+		return fmt.Errorf("rule %q expects a string argument, got %v of type %T", r.Name(), arg, arg)
 	}
 
 	format, err := regexp.Compile(str)
 	if err != nil {
-		panic(fmt.Sprintf("rule %q expects a valid regexp argument, got %v for %s", r.Name(), err, arg))
+		return fmt.Errorf("rule %q expects a valid regexp argument, got %v for %s", r.Name(), err, arg)
```
ccoVeille, line 84:
```suggestion
		return fmt.Errorf("rule %q expects a valid regexp argument, got %w for %s", r.Name(), err, arg)
```

rule/function_result_limit.go
```diff
@@ -16,31 +17,32 @@ type FunctionResultsLimitRule struct {
...
 	}
 
 	maxResults, ok := arguments[0].(int64) // Alt. non panicking version
 	if !ok {
-		panic(fmt.Sprintf(`invalid value passed as return results number to the "function-result-limit" rule; need int64 but got %T`, arguments[0]))
+		return fmt.Errorf(`invalid value passed as return results number to the "function-result-limit" rule; need int64 but got %T`, arguments[0])
 	}
 	if maxResults < 0 {
-		panic(`the value passed as return results number to the "function-result-limit" rule cannot be negative`)
+		return fmt.Errorf(`the value passed as return results number to the "function-result-limit" rule cannot be negative`)
```
ccoVeille, line 37:
```suggestion
		return errors.New(`the value passed as return results number to the "function-result-limit" rule cannot be negative`)
```

rule/imports_blocklist.go
```diff
@@ -26,15 +27,16 @@ func (r *ImportsBlocklistRule) configure(arguments lint.Arguments) {
 		for _, arg := range arguments {
 			argStr, ok := arg.(string)
 			if !ok {
-				panic(fmt.Sprintf("Invalid argument to the imports-blocklist rule. Expecting a string, got %T", arg))
+				return fmt.Errorf("Invalid argument to the imports-blocklist rule. Expecting a string, got %T", arg)
 			}
 			regStr, err := regexp.Compile(fmt.Sprintf(`(?m)"%s"$`, replaceImportRegexp.ReplaceAllString(argStr, `(\W|\w)*`)))
 			if err != nil {
-				panic(fmt.Sprintf("Invalid argument to the imports-blocklist rule. Expecting %q to be a valid regular expression, got: %v", argStr, err))
+				return fmt.Errorf("Invalid argument to the imports-blocklist rule. Expecting %q to be a valid regular expression, got: %v", argStr, err)
```
ccoVeille, line 34:
```suggestion
				return fmt.Errorf("Invalid argument to the imports-blocklist rule. Expecting %q to be a valid regular expression, got: %w", argStr, err)
```

rule/line_length_limit.go
```diff
@@ -20,28 +21,29 @@ type LineLengthLimitRule struct {
...
 	if len(arguments) < 1 {
 		r.max = defaultLineLengthLimit
-		return
+		return nil
 	}
 
 	maxLength, ok := arguments[0].(int64) // Alt. non panicking version
 	if !ok || maxLength < 0 {
-		panic(`invalid value passed as argument number to the "line-length-limit" rule`)
+		return fmt.Errorf(`invalid value passed as argument number to the "line-length-limit" rule`)
```
ccoVeille, line 38:
```suggestion
		return errors.New(`invalid value passed as argument number to the "line-length-limit" rule`)
```

rule/max_control_nesting.go
```diff
@@ -106,23 +107,24 @@ func (w *lintMaxControlNesting) walkControlledBlock(b ast.Node) {
...
-		return
+		return nil
 	}
 
 	checkNumberOfArguments(1, arguments, r.Name())
 
 	maxNesting, ok := arguments[0].(int64) // Alt. non panicking version
 	if !ok {
-		panic(`invalid value passed as argument number to the "max-control-nesting" rule`)
+		return fmt.Errorf(`invalid value passed as argument number to the "max-control-nesting" rule`)
```
ccoVeille, line 126:
```suggestion
		return errors.New(`invalid value passed as argument number to the "max-control-nesting" rule`)
```

rule/max_public_structs.go
```diff
@@ -16,29 +18,30 @@ type MaxPublicStructsRule struct {
...
-		return
+		return nil
 	}
 
 	checkNumberOfArguments(1, arguments, r.Name())
 
 	maxStructs, ok := arguments[0].(int64) // Alt. non panicking version
 	if !ok {
-		panic(`invalid value passed as argument number to the "max-public-structs" rule`)
+		return fmt.Errorf(`invalid value passed as argument number to the "max-public-structs" rule`)
```
ccoVeille, line 37:
```suggestion
		return errors.New(`invalid value passed as argument number to the "max-public-structs" rule`)
```

rule/unchecked_type_assertion.go
```diff
@@ -20,36 +21,37 @@ type UncheckedTypeAssertionRule struct {
...
-		return
+		return nil
 	}
 
 	u.configured = true
 
 	args, ok := arguments[0].(map[string]any)
 	if !ok {
-		panic("Unable to get arguments. Expected object of key-value-pairs.")
+		return fmt.Errorf("Unable to get arguments. Expected object of key-value-pairs")
```
ccoVeille, line 36:
```suggestion
		return errors.New("Unable to get arguments. Expected object of key-value-pairs")
```

rule/unhandled_error.go
```diff
@@ -17,36 +18,37 @@ type UnhandledErrorRule struct {
...
 		argStr, ok := arg.(string)
 		if !ok {
-			panic(fmt.Sprintf("Invalid argument to the unhandled-error rule. Expecting a string, got %T", arg))
+			return fmt.Errorf("Invalid argument to the unhandled-error rule. Expecting a string, got %T", arg)
 		}
 
 		argStr = strings.Trim(argStr, " ")
 		if argStr == "" {
-			panic("Invalid argument to the unhandled-error rule, expected regular expression must not be empty.")
+			return fmt.Errorf("Invalid argument to the unhandled-error rule, expected regular expression must not be empty")
```
ccoVeille, line 37:
```suggestion
			return errors.New("Invalid argument to the unhandled-error rule, expected regular expression must not be empty")
```

rule/unhandled_error.go
```diff
@@ -17,36 +18,37 @@ type UnhandledErrorRule struct {
...
 		}
 
 		exp, err := regexp.Compile(argStr)
 		if err != nil {
-			panic(fmt.Sprintf("Invalid argument to the unhandled-error rule: regexp %q does not compile: %v", argStr, err))
+			return fmt.Errorf("Invalid argument to the unhandled-error rule: regexp %q does not compile: %v", argStr, err)
```
ccoVeille, line 42:
```suggestion
			return fmt.Errorf("Invalid argument to the unhandled-error rule: regexp %q does not compile: %w", argStr, err)
```

rule/unused_param.go
```diff
@@ -41,23 +42,24 @@ func (r *UnusedParamRule) configure(args lint.Arguments) {
...
-				panic(fmt.Errorf("error configuring %s rule: allowedRegex is not string but [%T]", r.Name(), allowedRegexParam))
+				return fmt.Errorf("error configuring %s rule: allowedRegex is not string but [%T]", r.Name(), allowedRegexParam)
 			}
 		}
 	}
 	var err error
 	r.allowRegex, err = regexp.Compile(allowedRegexStr)
 	if err != nil {
-		panic(fmt.Errorf("error configuring %s rule: allowedRegex is not valid regex [%s]: %v", r.Name(), allowedRegexStr, err))
+		return fmt.Errorf("error configuring %s rule: allowedRegex is not valid regex [%s]: %v", r.Name(), allowedRegexStr, err)
```
ccoVeille, line 52:
```suggestion
		return fmt.Errorf("error configuring %s rule: allowedRegex is not valid regex [%s]: %w", r.Name(), allowedRegexStr, err)
```

rule/unused_receiver.go
```diff
@@ -41,22 +42,23 @@ func (r *UnusedReceiverRule) configure(args lint.Arguments) {
...
-				panic(fmt.Errorf("error configuring [unused-receiver] rule: allowedRegex is not string but [%T]", allowedRegexParam))
+				return fmt.Errorf("error configuring [unused-receiver] rule: allowedRegex is not string but [%T]", allowedRegexParam)
 			}
 		}
 	}
 	var err error
 	r.allowRegex, err = regexp.Compile(allowedRegexStr)
 	if err != nil {
-		panic(fmt.Errorf("error configuring [unused-receiver] rule: allowedRegex is not valid regex [%s]: %v", allowedRegexStr, err))
+		return fmt.Errorf("error configuring [unused-receiver] rule: allowedRegex is not valid regex [%s]: %v", allowedRegexStr, err)
```
ccoVeille, line 52:
```suggestion
		return fmt.Errorf("error configuring [unused-receiver] rule: allowedRegex is not valid regex [%s]: %w", allowedRegexStr, err)
```

ccoVeille (changes requested):
It wasn't fun to review 😅 27 minutes

lint/package.go
```diff
@@ -182,17 +183,41 @@ func (p *Package) scanSortable() {
...
+    defer cancel() // Ensure the context is canceled at the end
+
 	for _, file := range p.files {
 		wg.Add(1)
 		go (func(file *File) {
-			file.lint(rules, config, failures)
-			wg.Done()
+			defer wg.Done()
+			if err := file.lint(rules, config, failures); err != nil {
+                select {
```
chavacava, line 199: All this code block seems to be a c&p from an AI. Careful must be taken when integrating AI generated code, it usually has subtle bugs (in particular with concurrent code)

rule/add_constant.go
```diff
@@ -40,10 +41,12 @@ type AddConstantRule struct {
 }
 
 // Apply applies the rule to given file.
-func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configure(arguments)
-
+func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) ([]lint.Failure, error) {
 	var failures []lint.Failure
+	err := r.configure(arguments)
```
chavacava, line 46: A recently merged PR (merged after your fork) introduced the use of `sync.Once` as a replacement for `sync.Mutex` in configuration functions. Even nice per se, the use of `sync.Once` complicates the retrieval of configuration results (i.e. retrieving the error result is not as straightforward as `err := r.configure(arguments)`)

lint/package.go
mfederowicz: reverted to original version, any suggestion how to handle error for ```file.lint(rules, config, failures)``` in proper way?

especialy if that code runs in go routine?

rule/add_constant.go
mfederowicz: @chavacava any suggestion how to do that in proper way?

lint/package.go
mfederowicz: ok reverted that code, any idea how to handle error and stop processing other rules? If we have error in configure in any rule we should stop processing ?

rule/add_constant.go
mfederowicz: @chavacava aaah should I merge changes from https://github.com/mgechev/revive/pull/1118/files, before made changes with ```err := r.configure(arguments)```

lint/package.go
chavacava: > how to handle error and stop processing other rules?

Just exit the program at the first error

> If we have error in configure in any rule we should stop processing ?

Currently, `revive` stops at the first error it encounters, I propose to keep that behavior.

rule/add_constant.go
mfederowicz: ok @chavacava I added to each rule struct additional field ```configureErr    error``` and then update:

```r.configureOnce.Do(func() { r.configure(arguments) })```

to:

```
r.configureOnce.Do(func() { r.configureErr = r.configure(arguments) })                                                                                                 
                                                                                                                                                                            
if r.configureErr != nil {                                                                                                                                             
  return nil, r.configureErr                                                                                                                                         
}
```  
after that with some changes in config file (make some invalid arguments), revive stops and prints on stdout correct error (of course related to rule)

File: rule/add_constant.go [hunk 26]
```diff
@@ -36,13 +38,17 @@ type AddConstantRule struct {
...
 
 // Apply applies the rule to given file.
-func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) ([]lint.Failure, error) {
+	r.configureOnce.Do(func() { r.configureErr = r.configure(arguments) })
+
+	if r.configureErr != nil {
+		return nil, r.configureErr
+	}
```
ccoVeille, line 51: I would use this


to avoid adding a variable to a struct that you only use in the scope of one method


```suggestion
	var err error
	r.configureOnce.Do(func() { err = r.configure(arguments) })

	if err != nil {
		return nil, err
	}
```

But this would fail only on first run

So maybe what you are looking for is to use a sync.OnceFunc and not a syncOne

This way you can do this


```suggestion
	err := r.configureOnce.Do(func() error { return r.configure(arguments) })

	if err != nil {
		return nil, err
	}
```

ccoVeille (changes requested):

rule/add_constant.go
mfederowicz: @ccoVeille i choose first suggestion, because second one generates errors:

```
r.configureOnce.Do((func() error literal)) (no value) used as value compiler (TooManyValues)                                                                
cannot use (func() error literal) (value of type func() error) as func() value in argument to r.configureOnce.Do compiler (IncompatibleAssign) 
```

ccoVeille (approved):

rule/add_constant.go
```diff
@@ -36,13 +38,16 @@ type AddConstantRule struct {
 	allowList       allowList
 	ignoreFunctions []*regexp.Regexp
 	strLitLimit     int
-
-	configureOnce sync.Once
+	configureOnce   sync.Once
```
alexandear, line 41: Please do not remove an empty line. It's added by purpose to separate rule's arguments with internal fields.

rule/add_constant.go
```diff
@@ -36,13 +38,16 @@ type AddConstantRule struct {
...
 
 // Apply applies the rule to given file.
-func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) ([]lint.Failure, error) {
+	var configureErr error
+	r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
+	if configureErr != nil {
+		return nil, configureErr
+	}
```
alexandear, line 50: If `r.configure` fails, only the first call will receive the error. Any subsequent calls—second, third, and so on—won’t return that error.

See this article https://victoriametrics.com/blog/go-sync-once/index.html

Maybe we can fix this by adding the `configureErr error` field to the `AddConstantRule` struct.

alexandear (changes requested):

rule/early_return.go
```diff
@@ -1,3 +1,4 @@
+// Package rule implements revive's linting rules.
```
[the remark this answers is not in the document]
mfederowicz: yes because we have 83 files in rule directory, and I thought that is a good issue to update that :)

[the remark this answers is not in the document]
alexandear: This line is not needed to be duplicated. The only place where it should be is https://github.com/mgechev/revive/blob/655e6060b2ef3915215e0d795c79131f1a48d7f1/rule/doc.go#L1 See https://go.dev/doc/comment for details.

Because we will have the following in the [doc](https://pkg.go.dev/github.com/mgechev/revive/rule):
image

rule/add_constant.go
mfederowicz: Yes i know that, and we want to revive stops when first error exists, by the way  ```configureErr error``` field was replaced by local var few commits ago :P

ccoVeille: I suggested using sync.OnceFunc earlier

https://github.com/mgechev/revive/pull/1126#discussion_r1845293327

I'm unsure if it was understood, or if I'm the one who misundertood how it works

What about https://pkg.go.dev/sync#OnceValue then ?

mfederowicz: ok @ccoVeille, @alexandear  that code works:

```
check := sync.OnceValue(func() error { return r.configure(arguments) })                                                                                                
 if err := check(); err != nil {                                                                                                                                        
 return nil, err                                                                                                                                                    
 }
```
but what about with ```configureOnce sync.Once``` in rule struct, because if we decide to above code, then that var is useless, and if that solution is the final one because modifing 34+ files every secound commit is not optimal workflow :P

rule/early_return.go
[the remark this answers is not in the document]
mfederowicz: ok removed all package comments except doc.go :)

lint/rule.go
```diff
@@ -14,7 +14,7 @@ type DisabledInterval struct {
 // Rule defines an abstract rule interface
 type Rule interface {
 	Name() string
-	Apply(*File, Arguments) []Failure
+	Apply(*File, Arguments) ([]Failure, error)
```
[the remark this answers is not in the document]
mfederowicz: Ok @alexandear any idea how to handle errors without of course excuting panic(), we can use in each Apply:

```
if err != nil {                                                                                                                                                
   fmt.Fprintln(os.Stderr, err)                                                                         
   os.Exit(1)                                                                                          
   }
```
but it is not elegant way :(

lint/linter.go
```diff
@@ -152,7 +152,11 @@ func (l *Linter) lintPackage(filenames []string, gover *goversion.Version, ruleS
 		return nil
 	}
 
-	pkg.lint(ruleSet, config, failures)
+	err := pkg.lint(ruleSet, config, failures)
+
```
denisvmedia, line 156: nit: I'd remove this line (it will be more consistent stylistically).

lint/package.go
```diff
@@ -182,17 +184,22 @@ func (p *Package) scanSortable() {
...
+func (p *Package) lint(rules []Rule, config Config, failures chan Failure) error {
 	p.scanSortable()
 	var wg sync.WaitGroup
 	for _, file := range p.files {
 		wg.Add(1)
 		go (func(file *File) {
-			file.lint(rules, config, failures)
+			err := file.lint(rules, config, failures)
+			if err != nil {
+				fmt.Fprintln(os.Stderr, err)
```
denisvmedia, line 195: Maybe prepend with some explanation string:

```go
fmt.Fprintln(os.Stderr, "Error during linting:", err)
```

The error itself may not be that readable if dropped out without it.

lint/linter.go
mfederowicz: @denisvmedia maybe something like this: ```return pkg.lint(ruleSet, config, failures)```?

lint/rule.go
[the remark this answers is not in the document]
mfederowicz: ok got it @denisvmedia you want something like that: https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/revive/revive.go#L53 ? but is it possible to do that in separate PR, or it should be done in that PR?

lint/package.go
```diff
@@ -182,17 +184,22 @@ func (p *Package) scanSortable() {
...
 	var wg sync.WaitGroup
 	for _, file := range p.files {
 		wg.Add(1)
 		go (func(file *File) {
-			file.lint(rules, config, failures)
+			err := file.lint(rules, config, failures)
+			if err != nil {
+				fmt.Fprintln(os.Stderr, "Error during linting:", err)
+				os.Exit(1)
+			}
```
ccoVeille, line 197: I don't get this

OK it will exit, but then what's the purpose of having an error in the signature of `lint` method

mfederowicz: @ccoVeille good point, if last step in lint method is os.Exit(1) so error in the signature is useless :)

ccoVeille: I would have kept the signature and refactored to return an error

I would recommend using this

https://pkg.go.dev/golang.org/x/sync/errgroup#Go

https://pkg.go.dev/golang.org/x/sync/errgroup#Wait

It's what you need

lint/linter.go
```diff
@@ -153,7 +153,7 @@ func (l *Linter) lintPackage(filenames []string, gover *goversion.Version, ruleS
 	}
 
 	pkg.lint(ruleSet, config, failures)
-
+
```
alexandear, line 156:
```suggestion

```

rule/add_constant.go
```diff
@@ -36,13 +37,14 @@ type AddConstantRule struct {
...
 }
 
 // Apply applies the rule to given file.
-func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) ([]lint.Failure, error) {
+	check := sync.OnceValue(func() error {return r.configure(arguments)})
+	if err := check(); err != nil {
+		return nil, err
+	}
```
alexandear, line 47: This won't work as you expected. The `configure` will be called multiple times.

You can check this by adding the statement `fmt.Println("configure add-constant")` inside the `func() error`. After running `revive` with enabled `add-constant` on a real project with multiple packages, there are a lot of lines `configure add-constant`.

alexandear (changes requested):

lint/package.go
```diff
@@ -181,17 +183,22 @@ func (p *Package) scanSortable() {
...
 			wg.Done()
 		})(file)
 	}
 	wg.Wait()
+	return nil
 }
```
ccoVeille, line 202: Please refactor to use

https://pkg.go.dev/golang.org/x/sync/errgroup#Go

https://pkg.go.dev/golang.org/x/sync/errgroup#Wait

_Originally posted by @ccoVeille in https://github.com/mgechev/revive/pull/1126#discussion_r1850932089_

Something like this

```go
func (p *Package) lint(rules []Rule, config Config, failures chan Failure) error { 
          p.scanSortable() 
          var wg errgroup.WaitGroup 
          for _, file := range p.files { 
                  file := file // not sure if needed since Go 1.22, but I'm doubtful with a go routine
                  wg.Add(1) 
                  wg.Go(func() error { 
                         defer wg.Done()
                         file.lint(rules, config, failures) 
                         err := file.lint(rules, config, failures) 
                         if err != nil { 
                                 return  fmt.Errorf("Error during linting: %w", err)
                         }
                         return nil
                  })
          } 
          wg.Wait() 
         return nil 
  }
```

This is pseudo code written on a phone

ccoVeille (changes requested):

rule/add_constant.go
mfederowicz: ok iam back to concept with local error variable:

```
var configureErr error                                                                                                                                                 
 r.configureOnce.Do(func() { configureErr = r.configure(arguments) })                                                                                                   
                                                                                                                                                                           
    if configureErr != nil {                                                                                                                                               
      return nil, configureErr                                                                                                                                           
    }
```
but now executing ``` r.configureOnce.Do() ``` runs only onece (added ``` fmt.Println("configure add-constant")``` to configure() and run on large project), and on output we have only one message on stdout

lint/linter.go
```diff
@@ -106,7 +106,7 @@ func (l *Linter) Lint(packages [][]string, ruleSet []Rule, config Config) (<-cha
 		wg.Add(1)
 		go func(pkg []string, gover *goversion.Version) {
 			if err := l.lintPackage(pkg, gover, ruleSet, config, failures); err != nil {
-				fmt.Fprintln(os.Stderr, err)
+				fmt.Fprintln(os.Stderr, "Error during linting:"+err.Error())
 				os.Exit(1)
 			}
 			wg.Done()
```
ccoVeille, line 112: Thanks for the changes you just made on errgroup

I just found another one here that could be refactored with errgroup.

I noticed it because you somehow tried to wrathe error message at an higher level

lint/package.go
mfederowicz: ok @ccoVeille  i made changes similar to https://pkg.go.dev/golang.org/x/sync/errgroup#example-Group-JustErrors

lint/linter.go
ccoVeille: I reported it because fixing a deep os.Exit, is almost as the same scope of your PR. Panic and os.Exit should be removed.

I can of course open a PR if you prefer

ccoVeille: I reported it because fixing a deep os.Exit, is almost as the same scope of your PR. Panic and os.Exit should be removed.

I can of course open a PR if you prefer

mfederowicz: ok ill try to refactor func (l *Linter) Lint() with errgroup way :)

rule/banned_characters.go
```diff
@@ -12,22 +12,36 @@ import (
...
 const bannedCharsRuleName = "banned-characters"
 
-func (r *BannedCharsRule) configure(arguments lint.Arguments) {
+func (r *BannedCharsRule) configure(arguments lint.Arguments) error {
 	if len(arguments) > 0 {
-		checkNumberOfArguments(1, arguments, bannedCharsRuleName)
-		r.bannedCharList = r.getBannedCharsList(arguments)
+		check := checkNumberOfArguments(1, arguments, bannedCharsRuleName)
+		if check != nil {
+			return check
```
alexandear, line 25:
```suggestion
		err := checkNumberOfArguments(1, arguments, bannedCharsRuleName)
		if err != nil {
			return err
```

rule/banned_characters.go
```diff
@@ -12,22 +12,36 @@ import (
 // BannedCharsRule checks if a file contains banned characters.
 type BannedCharsRule struct {
 	bannedCharList []string
-
-	configureOnce sync.Once
+
```
alexandear, line 15:
```suggestion

```

rule/banned_characters.go
```diff
@@ -12,22 +12,36 @@ import (
 // BannedCharsRule checks if a file contains banned characters.
 type BannedCharsRule struct {
 	bannedCharList []string
-
-	configureOnce sync.Once
+	
+	configureOnce  sync.Once
```
alexandear, line 16:
```suggestion
	configureOnce sync.Once
```

rule/add_constant.go
```diff
@@ -228,39 +234,47 @@ func (r *AddConstantRule) configure(arguments lint.Arguments) {
...
 
 				r.ignoreFunctions = append(r.ignoreFunctions, exp)
 			}
 		}
 	}
+
+	return nil
+}
+
+func configOnce(r *AddConstantRule, arguments lint.Arguments) func() error {
```
alexandear, line 276: this function is not used?

lint/package.go
```diff
@@ -181,17 +182,23 @@ func (p *Package) scanSortable() {
...
-		go (func(file *File) {
-			file.lint(rules, config, failures)
-			wg.Done()
-		})(file)
+		eg.Go(func() error {
+			err := file.lint(rules, config, failures)
+			if err != nil {
+				return err
+			}
+			return nil
```
alexandear, line 194:
```suggestion
			return file.lint(rules, config, failures)
```

lint/package.go
```diff
@@ -181,17 +182,23 @@ func (p *Package) scanSortable() {
...
+		})
+	}
+	if err := eg.Wait(); err != nil {
```
alexandear, line 197: can be

```suggestion
	return eg.Wait()
```

rule/add_constant.go
mfederowicz: true forgot remove after tests, fixed :)

rule/add_constant.go
```diff
@@ -41,8 +42,13 @@ type AddConstantRule struct {
...
 // Apply applies the rule to given file.
-func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) ([]lint.Failure, error) {
+	var configureErr error
+	r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
+
+	if configureErr != nil {
+		return nil, configureErr
+	}
```
ccoVeille, line 51: I Iooked at it again, and there is something I don't get.

The code was

```go
        r.configureOnce.Do(func() { r.configure(arguments) })
```

When there were no problems, r.configure is called only once, even when Add is called multiple times.

Also, r.configure was raising panic
The panic was stopping everything.



Now, the behavior is 

```go
        var configureErr error 
         r.configureOnce.Do(func() { configureErr = r.configure(arguments) }) 
  
         if configureErr != nil { 
                 return nil, configureErr 
         }
```

We kept the behavior of having r.configure being called only once when there is no problem.

But, the behavior differs when an error occurs.

configureErr will be filled with the error only on the first call to Add. The same way the code was doing with a panic, somehow.

So consecutive calls to add won't return an error.

So we are relying on the fact the code calling Add, will stop everything.

It might be OK, but I would like to share my understanding.


So here, either:
- we add a comment that acknowledge,
 it's done on purpose
- we think about a better solution

ccoVeille: Please note, I'm not asking author to find a solution, but really we find one together

chavacava:
I've created the PR #1178 with a proposal on how to remove panics from rules. Please check it out and let me know if some ideas can be used in this PR

ccoVeille:
I love the idea, I totally forgot the message we are raising are "failures", so a kind of error.

It's a great idea that there is no need to update all rules

rule/add_constant.go
chavacava: The usual solution to this problem is memoization: store the result of the call to `configure` (error or nil) and initialize `configureError`  with that value. 

```go
configureError:= r.configureError
r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
```

ccoVeille: Yes, I also thought about it, as a possible solution, but it adds a another variable to the struct, but why not after all

lint/rule.go
```diff
@@ -14,7 +14,7 @@ type DisabledInterval struct {
 // Rule defines an abstract rule interface
 type Rule interface {
 	Name() string
-	Apply(*File, Arguments) []Failure
+	Apply(*File, Arguments) ([]Failure)
```
chavacava, line 17:
```suggestion
	Apply(*File, Arguments) []Failure
```

rule/add_constant.go
```diff
@@ -232,35 +233,35 @@ func (r *AddConstantRule) configure(arguments lint.Arguments) error {
 			}
 			list, ok := v.(string)
 			if !ok {
-				fmt.Errorf("invalid argument to the add-constant rule, string expected. Got '%v' (%T)", v, v)
+				return fmt.Errorf("Invalid argument to the add-constant rule, string expected. Got '%v' (%T)", v, v)
```
chavacava, line 236:
```suggestion
				return fmt.Errorf("invalid argument to the add-constant rule, string expected. Got '%v' (%T)", v, v)
```
Error messages should start in lowercase

chavacava:
@mfederowicz thanks for the updates. Please run
```
revive --config revive.toml ./...
```
On the branch, you will see that the problem of error messages in uppercase is present in almost all configuration error messages

nit: we could add an utility function (in rules/utils.go ?) to avoid repeating `[]lint.Failure{lint.NewInternalFailure(configureErr.Error())}` at each rule.

lint/file.go
```diff
@@ -107,6 +107,7 @@ func (f *File) lint(rules []Rule, config Config, failures chan Failure) error {
 			continue
 		}
 		currentFailures := currentRule.Apply(f, ruleConfig.Arguments)
+
```
alexandear, line 110: I suggest not modifying this file.

rule/dot_imports.go
```diff
@@ -41,30 +46,31 @@ func (*DotImportsRule) Name() string {
...
 	r.allowedPackages = allowPackages{}
 	if len(arguments) == 0 {
-		return
+		return nil
 	}
 
 	args, ok := arguments[0].(map[string]any)
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument to the dot-imports rule. Expecting a k,v map, got %T", arguments[0]))
+		return fmt.Errorf("Invalid argument to the dot-imports rule. Expecting a k,v map, got %T", arguments[0])
```
alexandear, line 57:
```suggestion
		return fmt.Errorf("invalid argument to the dot-imports rule. Expecting a k,v map, got %T", arguments[0])
```

rule/dot_imports.go
```diff
@@ -41,30 +46,31 @@ func (*DotImportsRule) Name() string {
...
 	}
 
 	if allowedPkgArg, ok := args["allowedPackages"]; ok {
 		pkgs, ok := allowedPkgArg.([]any)
 		if !ok {
-			panic(fmt.Sprintf("Invalid argument to the dot-imports rule, []string expected. Got '%v' (%T)", allowedPkgArg, allowedPkgArg))
+			return fmt.Errorf("Invalid argument to the dot-imports rule, []string expected. Got '%v' (%T)", allowedPkgArg, allowedPkgArg)
```
alexandear, line 63:
```suggestion
			return fmt.Errorf("invalid argument to the dot-imports rule, []string expected. Got '%v' (%T)", allowedPkgArg, allowedPkgArg)
```

rule/dot_imports.go
```diff
@@ -41,30 +46,31 @@ func (*DotImportsRule) Name() string {
...
 		}
 		for _, p := range pkgs {
 			pkg, ok := p.(string)
 			if !ok {
-				panic(fmt.Sprintf("Invalid argument to the dot-imports rule, string expected. Got '%v' (%T)", p, p))
+				return fmt.Errorf("Invalid argument to the dot-imports rule, string expected. Got '%v' (%T)", p, p)
```
alexandear, line 68:
```suggestion
				return fmt.Errorf("invalid argument to the dot-imports rule, string expected. Got '%v' (%T)", p, p)
```

rule/add_constant.go
mfederowicz: resolved with ``` return []lint.Failure{lint.NewInternalFailure(configureErr.Error())} ```

mfederowicz:
ok @chavacava i replaced almost all errors msg in rules, but there are some cases in testdata what about them?

rule/utils.go
```diff
@@ -121,3 +122,8 @@ func isDirectiveComment(line string) bool {
 func isCallToExitFunction(pkgName, functionName string) bool {
 	return exitFunctions[pkgName] != nil && exitFunctions[pkgName][functionName]
 }
+// newInternalFailureError returns single internal failure
```
chavacava, line 125:
```suggestion

// newInternalFailureError returns an slice of Failure with a single internal failure in it 
```

chavacava (approved):
Some details to adjust but overall OK.
Thanks @mfederowicz !

.gitignore
```diff
@@ -4,3 +4,4 @@ vendor
 *.swp
 dist/
 *.log
+.idea/
```
ccoVeille, line 7: Is it normal?

ccoVeille:
You did great, you managed to fix the panic issue.

I will try to open a PR with ideas I have.

The errgroup in linterlinter/lint.go (unsure it would work as I expect anyway) to avoid a deep-exit

Then I will be the one who will take feedbacks from you guys 😅

ccoVeille (approved):

.gitignore
mfederowicz: yes if somebody use goland as IDE from intelij :P

denisvmedia (approved):
LGTM

rule/add_constant.go
```diff
@@ -68,6 +69,7 @@ func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lin
 	return failures
 }
 
+
```
alexandear, line 72: please remove this redundant empty line

rule/banned_characters.go
```diff
@@ -18,16 +18,30 @@ type BannedCharsRule struct {
...
 
 // Apply applied the rule to the given file.
 func (r *BannedCharsRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+	var configureErr error
+	r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
+
+	if configureErr != nil {
+		return newInternalFailureError(configureErr)
+	}
```
alexandear, line 44: I propose adding a test to check a situation where `configureErr` is not `nil`. Not for all rules but at least for a one.

rule/enforce_slice_style.go
```diff
@@ -48,27 +48,33 @@ type EnforceSliceStyleRule struct {
...
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument '%v' for 'enforce-slice-style' rule. Expecting string, got %T", arguments[0], arguments[0]))
+		return fmt.Errorf("invalid argument '%v' for 'enforce-slice-style' rule. Expecting string, got %T", arguments[0], arguments[0])
 	}
 
 	var err error
 	r.enforceSliceStyle, err = sliceStyleFromString(enforceSliceStyle)
 	if err != nil {
-		panic(fmt.Sprintf("Invalid argument to the enforce-slice-style rule: %v", err))
+		return fmt.Errorf("invalid argument to the enforce-slice-style rule: %v", err)
```
alexandear, line 65:
```suggestion
		return fmt.Errorf("invalid argument to the enforce-slice-style rule: %w", err)
```

rule/file_header.go
```diff
@@ -69,7 +75,7 @@ func (r *FileHeaderRule) Apply(file *lint.File, arguments lint.Arguments) []lint
 
 	regex, err := regexp.Compile(r.header)
 	if err != nil {
-		panic(err.Error())
+		return []lint.Failure{lint.NewInternalFailure(err.Error())}
```
alexandear, line 78: Why not
```suggestion
		return newInternalFailureError(err)
```
?

rule/file_length_limit.go
```diff
@@ -44,7 +49,7 @@ func (r *FileLengthLimitRule) Apply(file *lint.File, arguments lint.Arguments) [
 	}
 
 	if err := scanner.Err(); err != nil {
-		panic(err.Error())
+		return []lint.Failure{lint.NewInternalFailure(err.Error())}
```
alexandear, line 52:
```suggestion
		return newInternalFailureError(err)
```

rule/filename_format.go
```diff
@@ -55,27 +60,29 @@ func (*FilenameFormatRule) Name() string {
...
 	str, ok := arg.(string)
 	if !ok {
-		panic(fmt.Sprintf("rule %q expects a string argument, got %v of type %T", r.Name(), arg, arg))
+		return fmt.Errorf("rule %q expects a string argument, got %v of type %T", r.Name(), arg, arg)
 	}
 
 	format, err := regexp.Compile(str)
 	if err != nil {
-		panic(fmt.Sprintf("rule %q expects a valid regexp argument, got %v for %s", r.Name(), err, arg))
+		return fmt.Errorf("rule %q expects a valid regexp argument, got %w for %s", r.Name(), err, arg)
```
alexandear, line 82: `err` usually the last:

```suggestion
		return fmt.Errorf("rule %q expects a valid regexp argument, got error for %s: %w", r.Name(), err)
```

rule/import_alias_naming.go
```diff
@@ -93,28 +109,30 @@ func (*ImportAliasNamingRule) Name() string {
...
 	namingRule, ok := value.(string)
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument '%v' for import-alias-naming allowRegexp rule. Expecting string, got %T", value, value))
+		return fmt.Errorf("invalid argument '%v' for import-alias-naming allowRegexp rule. Expecting string, got %T", value, value)
 	}
 
 	namingRuleRegexp, err := regexp.Compile(namingRule)
 	if err != nil {
-		panic(fmt.Sprintf("Invalid argument to the import-alias-naming allowRegexp rule. Expecting %q to be a valid regular expression, got: %v", namingRule, err))
+		return fmt.Errorf("invalid argument to the import-alias-naming allowRegexp rule. Expecting %q to be a valid regular expression, got: %v", namingRule, err)
```
alexandear, line 120:
```suggestion
		return fmt.Errorf("invalid argument to the import-alias-naming allowRegexp rule. Expecting %q to be a valid regular expression, got: %w", namingRule, err)
```

rule/import_alias_naming.go
```diff
@@ -93,28 +109,30 @@ func (*ImportAliasNamingRule) Name() string {
...
 	namingRule, ok := value.(string)
 	if !ok {
-		panic(fmt.Sprintf("Invalid argument '%v' for import-alias-naming denyRegexp rule. Expecting string, got %T", value, value))
+		return fmt.Errorf("invalid argument '%v' for import-alias-naming denyRegexp rule. Expecting string, got %T", value, value)
 	}
 
 	namingRuleRegexp, err := regexp.Compile(namingRule)
 	if err != nil {
-		panic(fmt.Sprintf("Invalid argument to the import-alias-naming denyRegexp rule. Expecting %q to be a valid regular expression, got: %v", namingRule, err))
+		return fmt.Errorf("invalid argument to the import-alias-naming denyRegexp rule. Expecting %q to be a valid regular expression, got: %v", namingRule, err)
```
alexandear, line 134:
```suggestion
		return fmt.Errorf("invalid argument to the import-alias-naming denyRegexp rule. Expecting %q to be a valid regular expression, got: %w", namingRule, err)
```

rule/string_format.go
```diff
@@ -43,10 +47,9 @@ func (StringFormatRule) ParseArgumentsTest(arguments lint.Arguments) *string {
 	// Parse the arguments in a goroutine, defer a recover() call, return the error encountered (or nil if there was no error)
 	go func() {
 		defer func() {
-			err := recover()
+			err := w.parseArguments(arguments)
 			c <- err
 		}()
```
alexandear, line 52: As we removed `recover` it's better to change also the explanation comment.

alexandear (changes requested):

rule/string_format.go
```diff
@@ -43,10 +47,9 @@ func (StringFormatRule) ParseArgumentsTest(arguments lint.Arguments) *string {
 	// Parse the arguments in a goroutine, defer a recover() call, return the error encountered (or nil if there was no error)
 	go func() {
 		defer func() {
-			err := recover()
+			err := w.parseArguments(arguments)
```
alexandear, line 50: Since we changed to return `error` instead of panicking, the `ParseArgumentsTest` became unnecessary. We can remove it.

.gitignore
ccoVeille: I disagree

This should be part of user's git ignore global settings

Otherwise, a .gitignore ends with excluding a lot things because everyone has a different IDE, so one may argue there is a need for VS Code, Sublime Text.

And it looks like a sneaky change out of scope added to the PR

That said I won't fight much than this.

mfederowicz: @ccoVeille yes you are 100% right, I forgott about global gitignore :P

denisvmedia: But in general I'd reiterate the discussion separately. The purpose of `.gitignore` (in my humble opinion) is to also prevent commits by mistake. Adding such folders for popular IDEs can help. I do have a global gitignore, but that's something that doesn't come out of the box and not everyone is aware of it. Not a big deal but can save some time for reviews if someone creates a PR with such files added by mistake.

rule/file_header.go
```diff
@@ -20,21 +20,27 @@ var (
...
 }
 
 // Apply applies the rule to given file.
 func (r *FileHeaderRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+	var configureErr error
+	r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
+
+	if configureErr != nil {
+		return []lint.Failure{lint.NewInternalFailure(configureErr.Error())}
```
alexandear, line 42:
```suggestion
		return newInternalFailureError(configureErr)
```

PR #1132, opened 2024-11-16
Filename-format rule: fix filename extension regex: ".go" to "\\.go".

denarced:
Just a quick fix for an obvious issue with the filename regex. Filename extension's dot wasn't escaped so, for example, "awesome$go" was considered a valid filename.

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -521,7 +521,7 @@ Example:
 ```
 
 ## filename-format
-_Description_: enforces conventions on source file names. By default, the rule enforces filenames of the form `^[_A-Za-z0-9][_A-Za-z0-9-]*.go$`: Optionally, the rule can be configured to enforce other forms.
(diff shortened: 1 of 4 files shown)


ccoVeille (approved):
Good catch, but a file naned `awesome$go`, won't be a go files, so something that revive would catch, no?

I mean the regex wasn't perfect, and you are fixing it. Thank you about that.

But it was a crime without victim, right ?

denarced:
True, very hard to find a victim. I've seen this type of slip-ups (not escaping filename extension dots in regex) many times before and in all cases, no one ever found the defect in production. I don't know the project so there was a chance that the fix was somehow useful, it was suppose to be quick (it actually wasn't), so I just fixed it. If I ended up breaking something, please just revert the commit :).

ccoVeille:
I'll let other people provide feedbacks.

But for me, you "bug fix" is legitimate

chavacava:
Hi @denarced, thanks for the fix.

PR #1142, opened 2024-11-20
adds rule use-errors-new

chavacava:
Implementing a new rule `use-errors-new` that proposes replacing `fmt.Errorf` by `errors.New` when possible.

As indicated in [this issue](https://github.com/googleapis/google-cloud-go/issues/9749), beyond style considerations, `errors.News` is ~4x faster that `fmt.Errorf`

(initial discussion on this PR can be retrieved at #1136)

### DIFF
--- a/rule/use_errors_new.go
+++ b/rule/use_errors_new.go
@@ -1,0 +1,60 @@
+package rule
+
+import (
+	"go/ast"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// UseErrorsNewRule spots calls to fmt.Errorf that can be replaced by errors.New.
+type UseErrorsNewRule struct{}
+
+// Apply applies the rule to given file.
+func (*UseErrorsNewRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	walker := lintFmtErrorf{
+		onFailure: func(failure lint.Failure) {
+			failures = append(failures, failure)
+		},
+	}
+
+	ast.Walk(walker, file.AST)
+
+	return failures
+}
+
+// Name returns the rule name.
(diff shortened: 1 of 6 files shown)


rule/use_errors_new.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	isFmtErrorf := isPkgDot(funcCall.Fun, "fmt", "Errorf")
+	if !isFmtErrorf {
+		return w // not a call to fmt.Errorf
+	}
+
+	if len(funcCall.Args) > 1 {
+		return w // the use of fmt.Errorf is legit
+	}
```
ccoVeille, line 49: Well, not exactly legit

We have crazy people out there

fmt.Errorf("%s, "whatever")
https://github.com/search?q=language%3Ago++%22fmt.Errorf%28%5C%22%25s%5C%22%2C+%5C%22%22&type=code

fmt.Errorf("%w", err)
https://github.com/search?q=language%3Ago++%22fmt.Errorf%28%5C%22%25w%5C%22%2C+%22&type=code

I created an issue on perfprint linter, as it's in their scope

- https://github.com/catenacyber/perfsprint/issues/36


I think you can merge like this, with current code.

My point can be addressed later


_Originally posted by @ccoVeille in https://github.com/mgechev/revive/pull/1136#discussion_r1850947674_

chavacava: It was never my original goal to detect misuses like those because it implies too much complexity (less performance) to spot very veeeeery rare cases (written by people deserving to live with their own crazy code :) )

ccoVeille: I'm fine with this.

Maybe the comment could be updated to explain what you just wrote

PR #1143, opened 2024-11-20
rename unused parameter

chavacava:
Renames `arguments` into `_` in redundant-build-tag rule


ccoVeille:
It's good to see revive code applying what revive suggests to do on my code

PR #1151, opened 2024-11-30
chore: Improve sortables detection

denisvmedia:
This PR resolves TODO comment that proposed to check sortable methods signatures.

* Ensures that only methods with the correct signature are identified as implementing required behavior (`sort.Interface`).
* Prevents accidental false positives from methods that happen to have the same name but differ in their functionality.

### DIFF
--- a/internal/astutils/ast_utils.go
+++ b/internal/astutils/ast_utils.go
@@ -1,0 +1,60 @@
+package astutils
+
+import "go/ast"
+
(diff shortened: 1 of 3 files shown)


chavacava:
Hi @denisvmedia thanks for the PR. I've did some refactoring, please check my commits

denisvmedia:
It's a new approach :) Better than mine. LGTM. Let's merge it.

PR #1154, opened 2024-12-01
chore: add missing tests for var-declarations, fix any/interface{} type issue

### LINKED ISSUE #1152: False positive with var-declaration
**Describe the bug**
The `var-declaration` check incorrectly warns if explicitly assigning a zero value in a `var` declaration, even if the zero value being assigned isn't the default for a variable of that type.

**To Reproduce**
2. I run it with the following flags:

```shell
revive main.go
```

main.go:

```go
// Package main is a package.
package main

func main() {
	var a any = 0
	_ = a

	var b any = ""
	_ = b
}
```

**Expected behavior**
I would expect no errors for `var-declaration`, because the zero value for a variable of type `any` is `nil`, not `0` or `""`.

**Logs**
```
main.go:5:14: should drop = 0 from declaration of var a; it is the zero value
main.go:8:14: should drop = "" from declaration of var b; it is the zero value
```

 - OS: macOS 15.1.1
 - Version of Go: `go version go1.23.3 darwin/arm64`

**Additional context**
N/A

denisvmedia:
Fixes #1152


testdata/var_declarations_type_inference.go
chavacava, line 1: I don"t see any added value of these tests with respect to those in `testdata/golint/var_declaration.go`

testdata/var_declarations_zero_value.go
chavacava, line 1: Add these tests to `testdata/golint/var_declaration.go` ?

testdata/var_declarations_zero_value.go
```diff
@@ -0,0 +1,9 @@
+package fixtures
+
+func zeroValueExample() {
+	var y int                // No warning, zero value is implicit
+	var z1 any = 0           // No warning, zero value for any is nil
```
chavacava, line 5: IMO, in cases of the form `var <id> any = <value>` where value is not `nil` the rule must warn with `should omit type any from declaration of var <id>; it will be inferred from the right-hand side` because `var <id> = <value>` is equivalent

testdata/var_declarations_type_inference.go
denisvmedia: Yeah, you are right. I missed that they were declared in `golint`.

testdata/var_declarations_zero_value.go
denisvmedia: I will revise the test cases in `testdata/golint/var_declaration.go` and maybe these ones are even not needed.

testdata/var_declarations_zero_value.go
chavacava: Yes

https://go.dev/play/p/TORLftc7jlR with any
https://go.dev/play/p/dYvM0OlUstA without any

chavacava: if an initial value is provided then declaring the variable of being type `any` has no utility

denisvmedia: Still, I think there is a difference. With `any` it is a dynamic type and it can be changed to any value later. Consider this:

```
package main

func main() {
	var a1 = 1
	// a1 = "string" // fail to compile
	var a2 any = 1
	a2 = "string" // this will compile
	_ = a1
	_ = a2
}
```

denisvmedia: > declaring the variable of being type `any` has no utility

I believe there can be a reason to declare an `any` type variable in case if you plan to have an arbitrary value for it later.

chavacava: Ooops; you are right... I  did not think in these kind of (ugly) cases

denisvmedia:
@chavacava I updated the tests. Are you fine with the current state?

ccoVeille:
I don't understand the last commit that provide a strange import order

denisvmedia:
It reverts what I committed by mistake.

PR #1165, opened 2024-12-04
 refactor (rule/modifies-value-receiver): replace AST walker by iteration over declarations

chavacava:
Simplifies modifies-value-receiver rule by replacing AST walker by an iteration over global declarations

### DIFF
--- a/rule/modifies_value_receiver.go
+++ b/rule/modifies_value_receiver.go
@@ -12,16 +12,41 @@ import (
 type ModifiesValRecRule struct{}
 
 // Apply applies the rule to given file.
-func (*ModifiesValRecRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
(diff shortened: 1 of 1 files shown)


rule/modifies_value_receiver.go
```diff
@@ -177,3 +118,60 @@ func (w lintModifiesValRecRule) findReturnReceiverStatements(receiverName string
...
+	}
+
+	if r.skipType(receiver.Type, pkg) {
+		return true // skip, receiver is a map or array
+	}
+
+	return false
+}
+
+func (r *ModifiesValRecRule) getAssignmentsToReceiver(receiverName string, funcBody *ast.BlockStmt) []ast.Node {
```
denisvmedia, line 143: ModifiesValRecRule seems to have a mixed receiver type, which is not recommended. I think they should all be pointers in this case.

PR #1166, opened 2024-12-04
refactor (rule/function-length): replace AST walker by iteration over declarations

chavacava:
Simplifies function-length rule by replacing AST walker by an iteration over global declarations

### DIFF
--- a/rule/function_length.go
+++ b/rule/function_length.go
@@ -41,7 +41,7 @@ func (r *FunctionLength) Apply(file *lint.File, arguments lint.Arguments) []lint
 		}
 
 		if r.maxStmt > 0 {
-			stmtCount := r.countStmts(funcDecl.Body.List)
+			stmtCount := r.countStmts(body.List)
 			if stmtCount > r.maxStmt {
 				failures = append(failures, lint.Failure{
 					Confidence: 1,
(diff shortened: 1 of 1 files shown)


rule/function_length.go
```diff
@@ -28,17 +28,40 @@ func (r *FunctionLength) Apply(file *lint.File, arguments lint.Arguments) []lint
...
-		},
-	}
+		body := funcDecl.Body
+		emptyBody := body == nil || len(body.List) == 0
+		if emptyBody {
+			return nil
+		}
+
+		if r.maxStmt > 0 {
+			stmtCount := r.countStmts(funcDecl.Body.List)
```
ccoVeille, line 44: Why not using `body` variable here?

rule/function_length.go
```diff
@@ -28,17 +28,40 @@ func (r *FunctionLength) Apply(file *lint.File, arguments lint.Arguments) []lint
...
+					Confidence: 1,
+					Failure:    fmt.Sprintf("maximum number of statements per function exceeded; max %d but got %d", r.maxStmt, stmtCount),
+					Node:       funcDecl,
+				})
+			}
+		}
 
-	ast.Walk(walker, file.AST)
+		if r.maxLines > 0 {
+			lineCount := r.countLines(funcDecl.Body, file)
```
ccoVeille, line 55: Here too

PR #1167, opened 2024-12-04
refactor (rule/function-result-limit): replace AST walker by iteration over declarations

chavacava:
Simplifies function-result-limit rule by replacing AST walker by an iteration over global declarations


rule/function_result_limit.go
```diff
@@ -15,39 +15,31 @@ type FunctionResultsLimitRule struct {
...
-		},
-	}
+		num := 0
+		hasResults := funcDecl.Type.Results != nil
+		if hasResults {
+			num = funcDecl.Type.Results.NumFields()
+		}
 
-	ast.Walk(walker, file.AST)
+		if num > r.max {
```
ccoVeille, line 35: What about simplifying to

    if num <= r.max {
       continue
    }

ccoVeille (approved):

PR #1168, opened 2024-12-04
refactor (rule/get-return): replace AST walker by iteration over declarations

chavacava:
Simplifies get-return rule by replacing AST walker by an iteration over global declarations

### DIFF
--- a/rule/get_return.go
+++ b/rule/get_return.go
@@ -15,22 +15,32 @@ type GetReturnRule struct{}
 func (*GetReturnRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 	var failures []lint.Failure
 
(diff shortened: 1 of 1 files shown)


rule/get_return.go
```diff
@@ -15,12 +15,26 @@ type GetReturnRule struct{}
...
+		fd, ok := decl.(*ast.FuncDecl)
+		if !ok {
+			continue
+		}
+
+		if !isGetter(fd.Name.Name) {
+			continue
+		}
+
+		if !hasResults(fd.Type.Results) {
```
ccoVeille, line 28:
```suggestion
		if hasResults(fd.Type.Results) {
			continue
		}
```

Also it's easier to read it continues if it has results, instead of the negation

PR #1169, opened 2024-12-04
refactor (rule/receiver-naming): replace AST walker by iteration over declarations

chavacava:
Simplifies receiver-naming rule by replacing AST walker by an iteration over global declarations

### DIFF
--- a/rule/receiver_naming.go
+++ b/rule/receiver_naming.go
@@ -47,18 +47,63 @@ func (r *ReceiverNamingRule) configure(arguments lint.Arguments) {
 func (r *ReceiverNamingRule) Apply(file *lint.File, args lint.Arguments) []lint.Failure {
 	r.configureOnce.Do(func() { r.configure(args) })
 
+	typeReceiver := map[string]string{}
 	var failures []lint.Failure
+	for _, decl := range file.AST.Decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Recv == nil || len(fn.Recv.List) == 0 {
+			continue
+		}
 
-	fileAst := file.AST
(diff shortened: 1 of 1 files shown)


rule/receiver_naming.go
```diff
@@ -47,18 +47,63 @@ func (r *ReceiverNamingRule) configure(arguments lint.Arguments) {
...
+
+		if name == "this" || name == "self" {
+			failures = append(failures, lint.Failure{
+				Node:       decl,
+				Confidence: 1,
+				Category:   "naming",
+				Failure:    `receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"`,
+			})
+			continue
+		}
```
ccoVeille, line 82: You could use a switch

You are  checking the same variable three times, also it would allow to remove the two continue, as they will be useless


```suggestion
		switch name {
			case "_":
				failures = append(failures, lint.Failure{
					Node:       decl,
					Confidence: 1,
					Category:   "naming",
					Failure:    "receiver name should not be an underscore, omit the name if it is unused",
				})
			case "this", "self":
				failures = append(failures, lint.Failure{
					Node:       decl,
					Confidence: 1,
					Category:   "naming",
					Failure:    `receiver name should be a reflection of its identity; don't use generic names such as "this" or "self"`,
				})
		}
```

chavacava: The switch on name will cope with the first two conditions for failure, the other conditions are not just on name thus IMO, adding a switch will not help in clarity (further, even with a switch `continue` is needed after adding a failure to do not evaluate the not-in-the-switch conditions)

ccoVeille (approved):

PR #1177, opened 2024-12-07
refactor: code cleanup

chavacava:
This PR:
1. removes unused code
2. replaces byte buffers with string builders when possible
3. removes if-then-else chains with switch and/or reduce nesting
4. renames some local variables


denisvmedia (approved):
LGTM

PR #1178, opened 2024-12-08
refactor: modifies linting machinery to use Failure as a mean to signal erros in rules application

chavacava:
This is PR adds the necessary modifications in the linting machinery to allow removing panics from rules.

It leverages the failure mechanism to communicate errors thus no need to change the rules interface.
It has the nice property of not requiring all rules to be modified at once.

### DIFF
--- a/lint/package.go
+++ b/lint/package.go
@@ -9,6 +9,7 @@ import (
 	"sync"
 
 	goversion "github.com/hashicorp/go-version"
+	"golang.org/x/sync/errgroup"
 
 	"github.com/mgechev/revive/internal/astutils"
 	"github.com/mgechev/revive/internal/typeparams"
@@ -183,32 +184,14 @@ func (p *Package) scanSortable() {
 
 func (p *Package) lint(rules []Rule, config Config, failures chan Failure) error {
 	p.scanSortable()
-	var wg sync.WaitGroup
-	errChan := make(chan error)
-	doneChan := make(chan struct{})
-
-	wg.Add(len(p.files))
-	go func() { // This goroutine will signal when all files where linted
-		wg.Wait()
-		doneChan <- struct{}{}
-	}()
-
+	var eg errgroup.Group
 	for _, file := range p.files {
-		go (func(file *File) {
-			err := file.lint(rules, config, failures)
-			if err != nil {
-				errChan <- err // signal the error
-			}
-			wg.Done()
-		})(file)
+		eg.Go(func() error {
+			return file.lint(rules, config, failures)
+		})
 	}
 
-	select { // We block until...
-	case <-doneChan: //...all files were linted
-		return nil
-	case err := <-errChan: //...or there is an error
-		return err
-	}
+	return eg.Wait()
 }
 
 // IsAtLeastGo121 returns true if the Go version for this package is 1.21 or higher, false otherwise
--- a/lint/package.go
+++ b/lint/package.go
@@ -181,17 +181,34 @@ func (p *Package) scanSortable() {
 	}
 }
 
-func (p *Package) lint(rules []Rule, config Config, failures chan Failure) {
+func (p *Package) lint(rules []Rule, config Config, failures chan Failure) error {
 	p.scanSortable()
 	var wg sync.WaitGroup
+	errChan := make(chan error)
+	doneChan := make(chan struct{})
+
+	wg.Add(len(p.files))
+	go func() { // This goroutine will signal when all files where linted
+		wg.Wait()
+		doneChan <- struct{}{}
+	}()
+
 	for _, file := range p.files {
-		wg.Add(1)
 		go (func(file *File) {
-			file.lint(rules, config, failures)
+			err := file.lint(rules, config, failures)
+			if err != nil {
+				errChan <- err // signal the error
+			}
 			wg.Done()
 		})(file)
 	}
-	wg.Wait()
+
+	select { // We block until...
+	case <-doneChan: //...all files were linted
+		return nil
+	case err := <-errChan: //...or there is an error
+		return err
+	}
 }
 
 // IsAtLeastGo121 returns true if the Go version for this package is 1.21 or higher, false otherwise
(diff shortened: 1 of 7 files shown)


lint/package.go
```diff
@@ -181,17 +181,34 @@ func (p *Package) scanSortable() {
...
 		})(file)
 	}
-	wg.Wait()
+
+	select { // We block until...
+	case <-doneChan: //...all files were linted
+		return nil
+	case err := <-errChan: //...or there is an error
+		return err
+	}
```
ccoVeille, line 211: What about using

https://pkg.go.dev/golang.org/x/sync/errgroup

The problems I see with your code are:
- the channels make it complicated
- errgroup exists
- all linters will be configured, no matter if one fail
- but you will return on first error
- so the go routines keeps being called for nothing

I would have said why not, if each errors was reported or returned

Of course, this feedback is based on my understanding of the code. I can be wrong 😅

rule/add_constant.go
```diff
@@ -42,7 +42,11 @@ type AddConstantRule struct {
 
 // Apply applies the rule to given file.
 func (r *AddConstantRule) Apply(file *lint.File, arguments lint.Arguments) []lint.Failure {
-	r.configureOnce.Do(func() { r.configure(arguments) })
+	var configureErr error
+	r.configureOnce.Do(func() { configureErr = r.configure(arguments) })
+	if configureErr != nil {
+		return []lint.Failure{lint.NewInternalFailure(configureErr.Error())}
+	}
```
ccoVeille, line 49: This idea is great, but it doesn't address this

https://github.com/mgechev/revive/pull/1126#discussion_r1874704591

But it could be addressed separately

chavacava: Yes, I know. It is not my intention to address this problem right now.
IMO configureOnce is cumbersome (vs just lock/unlock) for dealing with errors

ccoVeille: Yes, I agree

Maybe linter could call configure if there is a Configure  interface on the Rule

But, some tests will need to be rewritten

lint/package.go
chavacava: Thanks for the feedback 
(I've edited your comment to add numbers to the list of problems)
My meta-feedback:
1. Fair (even if I personally prefer explicit code over "hidden magic")
2. True (more on errorgroup below)
3. Not necessary _all_ but I get the point
4. True
5. On error, control exits the whole program thus gorutines will die.

On errgroup:  
1. as used in the PR #1126, errgroup does not necessarily prevent for configuring "all linters, no matter if one fail"
```
revive -config defaults.toml ./...
>>>>> configure addconstant
>>>>> configure arguments limits
Error during linting:Invalid argument to the add-constant rule, expecting a k,v map. Got string
```
2. as used in the PR #1126, errgroup does not necessarily prevent for "goroutines keep being called"
(each `>>>> linting ...` is a goroutine)
```
revive -config defaults.toml ./...
>>>>>> linting file  internal/typeparams/typeparams.go
>>>>>> linting file  internal/typeparams/typeparams_go118.go
>>>>> configure addconstant
>>>>>> linting file  config/config.go
>>>>>> linting file  config/config_test.go
>>>>>> linting file  internal/ifelse/branch.go
>>>>>> linting file  internal/ifelse/branch_kind.go
>>>>>> linting file  internal/ifelse/chain.go
>>>>>> linting file  internal/ifelse/doc.go
>>>>>> linting file  internal/ifelse/func.go
>>>>>> linting file  internal/ifelse/rule.go
>>>>>> linting file  internal/ifelse/target.go
>>>>>> linting file  internal/ifelse/args.go
>>>>>> linting file  main.go
>>>>> configure arguments limits
>>>>> configure banned chars
>>>>>> linting file  formatter/ndjson.go
>>>>>> linting file  formatter/plain.go
>>>>>> linting file  formatter/sarif.go
>>>>>> linting file  formatter/stylish.go
>>>>>> linting file  formatter/default.go
>>>>>> linting file  formatter/doc.go
>>>>>> linting file  formatter/json.go
>>>>>> linting file  formatter/severity.go
>>>>>> linting file  formatter/unix.go
>>>>>> linting file  formatter/checkstyle.go
>>>>>> linting file  formatter/friendly.go
>>>>>> linting file  internal/astutils/ast_utils.go
>>>>>> linting file  cli/main.go
>>>>>> linting file  cli/main_test.go
>>>>>> linting file  lint/doc.go
>>>>>> linting file  lint/file.go
>>>>>> linting file  lint/formatter.go
>>>>>> linting file  lint/package.go
>>>>>> linting file  lint/linter_test.go
>>>>>> linting file  lint/name_test.go
>>>>>> linting file  lint/config.go
>>>>>> linting file  lint/failure.go
>>>>>> linting file  lint/filefilter.go
>>>>>> linting file  lint/linter.go
>>>>>> linting file  lint/name.go
>>>>>> linting file  lint/rule.go
Error during linting:Invalid argument to the add-constant rule, expecting a k,v map. Got string
```
To resume, errgroup is (subjectively) simpler but still shares the same drawbacks of explicitly using channels. (and to be fair, these drawbacks are also present in the current -with panics- implementation of the linting machinery)

Anyway, the single significant point of my proposal is: use the already available failure mechanism instead of adding an error return value to rules.

rule/add_constant.go
chavacava: I've studied before the alternative of adding a `Configure` function to rules: it simplifies configuration of rules but it has the drawback of changing the interface.

lint/package.go
ccoVeille: Thanks for your detailed reply and feedback.

I agree with your conclusion

👍

rule/add_constant.go
ccoVeille: My point was to add an interface for Configure only.

type Configure interface {
   Configure() error
}

If detected, you call it after a type assertion.

But it's maybe part of what you considered

mfederowicz: ok @chavacava nice concept, can you deploy that changes for lint/* files without add_constant.go rule, and in #1126 do the rest of job (add ``` []lint.Failure{lint.NewInternalFailure(configureErr.Error())} ``` where it is needed) :P

PR #1209, opened 2025-01-26
fix: invalid message reported for directive comment

### LINKED ISSUE #1202: Improve "exported" rule
Right now, exported rule will report this

```go
//nolint:lll // foo
func Whatever() {}
```

>Blah blah commented should be form like ...

It also reports this

```go
//go:embed foo.json
var foo string
```

Or anything that is a Go directive

I spotted it by working on this https://github.com/gofr-dev/gofr/pull/1411 and especially this code

```go
		// code before
}

//nolint:exhaustive // We just want to take care of slice and struct in this case. 
func (c *Client) QueryWithCtx(ctx context.Context, dest any, stmt string, values ...any) error {

		// code after
```

https://github.com/gofr-dev/gofr/pull/1411/files#diff-5931c571fb913c21edcd5cd8b864b2fee691ebda14070081ad5765696b51646dR120-123

I had to add a comment for the method to make it works.

The problem occurs with golangci-lint by using EXC0014 only

It might be related to the fact exported already report a comment should be present for exported

But here, for me, suggesting "should be form" with a Go directive is a false positive

ccoVeille:
They were invalid but not reported with the correct message.

So they were wrongly caught by golangci-lint that excluded them by default.

Fixes #1202

### DIFF
--- a/rule/exported.go
+++ b/rule/exported.go
@@ -187,7 +187,7 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
 		return
 	}
 
-	if fn.Doc == nil {
+	if !hasTextComment(fn.Doc) {
 		w.onFailure(lint.Failure{
 			Node:       fn,
 			Confidence: 1,
@@ -252,7 +252,7 @@ func (w *lintExported) lintTypeDoc(t *ast.TypeSpec, doc *ast.CommentGroup) {
 		return
 	}
 
-	if doc == nil {
+	if !hasTextComment(doc) {
 		w.onFailure(lint.Failure{
 			Node:       t,
 			Confidence: 1,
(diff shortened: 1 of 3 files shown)


ccoVeille:
This is the first iteration of a wider refactoring I want to conduct

rule/exported.go
```diff
@@ -364,6 +364,19 @@ func (w *lintExported) lintValueSpecDoc(vs *ast.ValueSpec, gd *ast.GenDecl, genD
...
+// hasTextComment returns true if the comment contains a text comment
+// e.g. //go:embed foo.txt a directive comment, not a text comment
+// e.g. //nolint:whatever is a directive comment, not a text comment
+func hasTextComment(comment *ast.CommentGroup) bool {
+	if comment == nil {
+		return false
+	}
+
+	// a comment could be directive and not a text comment
+	text := comment.Text()
```
chavacava, line 376: nit: in the next PR please add a comment explaining that ` .Text`  removes directives from the comment block

ccoVeille:
> This is the first iteration of a wider refactoring I want to conduct.
> 
> I found other false positives, but there is no need for me to fix them for now, if we don't agree on the current PR.

For records, I opened the issues for them, please review them:
- #1231 
- #1234 
- #1235
- #1236

Issue #1233, opened 2025-02-12
vendor is not skipped


marcelloh:
**Describe the bug**
revive -formatter friendly -exclude ./vendor/... ./...
This should skip any vendor messages

**To Reproduce**
revive -version                                       
version 1.6.1

```shell
revive -formatter friendly -exclude ./vendor/... ./...
```

**Expected behavior**
to have no messages like:
  ⚠  https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#var-naming  don't use underscores in Go names; struct field single_quoted_allowed should be singleQuotedAllowed  
  vendor/gopkg.in/yaml.v3/yamlh.go:779:3

**Desktop (please complete the following information):**
 - OS: MacOS  15.3.1
 - Go: 1.24

denisvmedia:
Duplicate: #1230

PR #1237, opened 2025-02-13
fix: exclusions are no longer working as before 

### LINKED ISSUE #1230: Exclusions broken
**Describe the bug**
With v1.6.1 which includes https://github.com/mgechev/revive/pull/1060 the exclusions are no longer working as before. I have previously been using

```
revive -exclude dir1/... -exclude dir2/... ./...
```

I see the globbing patterns have replaced the dots, however I have been unable to make anything work. I've tried:

- `'dir1/*'`
- `'dir1/**'`
- `'./dir1/**'`
- - `'dir1/**.go'`

**To Reproduce**
See above

**Expected behavior**
Should be able to exclude entire directories

**Logs**
no

 - Ubuntu 24.04.1 LTS 
 - Go 1.23.6

**Additional context**
If this does work, some better samples in the CLI help would be good.
Also, this was clearly a breaking change and not just a patch version.

chavacava:
Fixes #1230 and #1233 by reverting #1060

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,6 @@ go 1.22.1
 
 require (
 	github.com/BurntSushi/toml v1.4.0
-	github.com/bmatcuk/doublestar/v4 v4.8.1
 	github.com/chavacava/garif v0.1.0
 	github.com/fatih/color v1.18.0
 	github.com/fatih/structtag v1.2.0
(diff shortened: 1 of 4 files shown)


denisvmedia (approved):
We have some weird test failing issue though 🤔

chavacava:
> We have some weird test failing issue though 🤔

go 1.24 changed the build info and now the version has a " (devel)" ... thus tests passing in 1.23 do not pass in 1.24 and the other way around. I'll fix

mfederowicz:
sorry for that @chavacava maybe i should back to the drawing board with this issue (make more tests, especially with magic ```./...``` patterns). Maybe good point will be adding some kind of switch in config (by default classic way, with other value new way with regexp), but i need more time for tests :)

PR #1239, opened 2025-02-16
feature: add support for URL struct tags in struct-tag rule

chavacava:
This PR extends struct-tag rule to check URL tags as defined at https://github.com/google/go-querystring (cf https://go.dev/wiki/Well-known-struct-tags#list-of-well-known-struct-tags)

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -102,6 +102,7 @@ const (
 	keyJSON     = "json"
 	keyProtobuf = "protobuf"
 	keyRequired = "required"
+	keyURL      = "url"
 	keyXML      = "xml"
 	keyYAML     = "yaml"
 )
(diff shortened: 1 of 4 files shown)


rule/struct_tag.go
```diff
@@ -102,6 +102,7 @@ const (
 	keyJSON     = "json"
 	keyProtobuf = "protobuf"
 	keyRequired = "required"
+	keyURL      = "url"
```
denisvmedia, line 105: I'd maybe update the docs to include all supported struct tags explicitly.

rule/struct_tag.go
```diff
@@ -322,6 +328,28 @@ func (w lintStructTagRule) checkYAMLTag(options []string) (string, bool) {
 	return "", true
 }
 
+func (w lintStructTagRule) checkURLTag(options []string) (string, bool) {
+	var delimiter = ""
+	for _, opt := range options {
+		switch opt {
+		case "int", "omitempty", "numbered", "brackets":
```
denisvmedia, line 335: It looks like we miss the following: `unix`, `unixmilli`, `unixnano`

Also, I'd consider adding support for `layout` and `del` struct tags, since they are a part of the spec for [the ](https://pkg.go.dev/github.com/google/go-querystring/query).

denisvmedia:
Thanks. Looks nice and simple. Just a few comments.

rule/struct_tag.go
chavacava: I' ve added unix* options.

`layout` and `del` : [almost no-body uses them](https://github.com/search?q=+layout%3A%22+path%3A*.go+language%3AGo&type=code&ref=advsearch) so I think it does not worth supporting these tags.

rule/struct_tag.go
chavacava: Yes, this PR is one in a series of PR related to `struct-tag`, I'll will update the doc at the end.

PR #1240, opened 2025-02-17
feature: add support of datastore struct tags in struct-tag rule

chavacava:
This PR extends struct-tag rule to check datastore tags as defined at https://pkg.go.dev/cloud.google.com/go/datastore (cf https://go.dev/wiki/Well-known-struct-tags#list-of-well-known-struct-tags)

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -98,15 +98,16 @@ func (w lintStructTagRule) Visit(node ast.Node) ast.Visitor {
 }
 
 const (
-	keyASN1     = "asn1"
-	keyBSON     = "bson"
-	keyDefault  = "default"
-	keyJSON     = "json"
-	keyProtobuf = "protobuf"
-	keyRequired = "required"
-	keyURL      = "url"
-	keyXML      = "xml"
-	keyYAML     = "yaml"
+	keyASN1      = "asn1"
+	keyBSON      = "bson"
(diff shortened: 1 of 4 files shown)


rule/struct_tag.go
```diff
@@ -358,6 +364,21 @@ func (w lintStructTagRule) checkURLTag(options []string) (string, bool) {
...
 
+func (w lintStructTagRule) checkDatastoreTag(options []string) (string, bool) {
+	for _, opt := range options {
+		switch opt {
+		case "flatten", "noindex", "omitempty":
+		default:
+			if w.isUserDefined(keyDatastore, opt) {
+				continue
+			}
+			return fmt.Sprintf("unknown option '%s' in Datastore tag", opt), false
```
ccoVeille, line 375: Nitpicking

```suggestion
			return fmt.Sprintf("unknown option '%s' in datastore tag", opt), false
```

Or a variation of this, with quotes maybe.

I'm unsure google datastore tag should be considered as an acronym or a brand.

JSON, BSON, XML were acronym. But look at the way "protobuf" tag is reported.

chavacava: I have also hesitated with the name :)
The [web site](https://cloud.google.com/products/datastore) says "Datastore"

Anyway, I'll write it all lowercases

denisvmedia (approved):

PR #1244, opened 2025-02-20
feature: add suport of validate struct tags in struct-tag rule

chavacava:
This PR extends struct-tag rule to check validate tags as defined at https://github.com/go-playground/validator (cf https://go.dev/wiki/Well-known-struct-tags#list-of-well-known-struct-tags)


denisvmedia (approved):
LGTM

PR #1255, opened 2025-03-03
feature: add support for TOML struct tags in struct-tag rule

chavacava:
This PR extends struct-tag rule to check TOML tags as defined at https://pkg.go.dev/github.com/pelletier/go-toml (cf https://go.dev/wiki/Well-known-struct-tags#list-of-well-known-struct-tags)


denisvmedia (approved):
LGTM

PR #1268, opened 2025-03-12
fix: cognitive complexity nesting with if-else chains

### LINKED ISSUE #1267: Cognitive complexity adds incorrect nesting for if-else chains
**Describe the bug**

Currently, an if-else chain will increase the nesting level and add the nesting increment for every addition `else if` statement in an if-else chain. This is incorrect behaviour; an `else if` statement should increment complexity by 1 (regardless of current nesting level) and leave the nesting level as-is.
    
For example, the following should yield a total complexity of 5:
```
    for { // +1
        if a { // +2 (nesting = 1)
            foo()
        } else if b { // +1
            bar()
        } else if c { // +1
            baz()
        }
    }
```
but the current implementation incorrectly increments the nesting level with each `else if` and adds the nesting increment where it shouldn't:
```
    for { // +1
        if a { // +2 (nesting = 1)
            foo()
        } else if b { // +3 (nesting = 2)
            bar()
        } else if c { // +4 (nesting = 3)
            baz()
        }
    }
```

AxiomaticFixedChimpanzee:
Fixes #1267

### DIFF
--- a/rule/cognitive_complexity.go
+++ b/rule/cognitive_complexity.go
@@ -98,8 +98,7 @@ func (v *cognitiveComplexityVisitor) subTreeComplexity(n ast.Node) int {
 func (v *cognitiveComplexityVisitor) Visit(n ast.Node) ast.Visitor {
 	switch n := n.(type) {
 	case *ast.IfStmt:
-		targets := []ast.Node{n.Cond, n.Body, n.Else}
-		v.walk(1, targets...)
+		v.walkIfElse(n)
 		return nil
 	case *ast.ForStmt:
 		targets := []ast.Node{n.Cond, n.Body}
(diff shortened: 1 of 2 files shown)


rule/cognitive_complexity.go
```diff
@@ -156,6 +155,29 @@ func (v *cognitiveComplexityVisitor) walk(complexityIncrement int, targets ...as
 	v.nestingLevel = nesting
 }
 
+func (v *cognitiveComplexityVisitor) walkIfElse(n *ast.IfStmt) {
+	var w func(n *ast.IfStmt)
+	w = func(n *ast.IfStmt) {
```
chavacava, line 160:
```suggestion
	w := func(n *ast.IfStmt) {
```

rule/cognitive_complexity.go
```diff
@@ -156,6 +155,29 @@ func (v *cognitiveComplexityVisitor) walk(complexityIncrement int, targets ...as
...
+		ast.Walk(v, n.Cond)
+		ast.Walk(v, n.Body)
+		if n.Else != nil {
+			if elif, ok := n.Else.(*ast.IfStmt); ok {
+				v.complexity += 1
```
ccoVeille, line 165:
```suggestion
				v.complexity++
```

testdata/cognitive_complexity.go
```diff
@@ -291,3 +291,16 @@ func Walk(t *Tree, ch chan int) { // MATCH /function Walk has cognitive complexi
...
+	if a { // +1
+		foo()
+	} else if b && c { // +2
+		bar()
+	} else if d { // +1
+		baz()
+	} else {
+		qux()
+	}
+}
```
ccoVeille, line 306:
```suggestion
// Test if-else if chains
func chainedIfElse(a, b, c, d bool) { // MATCH /function chainedIfElse has cognitive complexity 4 (> max enabled 0)/
	if a { // +1
		foo()
	} else if b && c { // +2
		bar()
	} else if d { // +1
		baz()
	} else {
		qux()
	}
}
```

rule/cognitive_complexity.go
```diff
@@ -156,6 +155,29 @@ func (v *cognitiveComplexityVisitor) walk(complexityIncrement int, targets ...as
...
+				w(elif)
+			} else {
+				ast.Walk(v, n.Else)
+			}
+		}
+	}
+
+	// The nesting level is incremented within the if and else blocks, but the nesting increment is only added to the complexity
+	// for the first if in an if-else-if chain.
```
ccoVeille, line 174: I would like to suggest something like this

```suggestion
	// Nesting level is incremented in 'if' and 'else' blocks, but only the first 'if' in an 'if-else-if' chain sees its complexity increased by the nesting level."
```

ccoVeille (approved):
Nitpicking and a rewording suggestion

rule/cognitive_complexity.go
AxiomaticFixedChimpanzee: The function is recursive, so must be declared first otherwise it won't compile.

ccoVeille (approved):

PR #1288, opened 2025-04-04
docs: place `redundant-build-tag` before `redundant-import-alias`

G-Rath:
I noticed this while opening https://github.com/golangci/golangci-lint/pull/5692


chavacava:
Thanks @G-Rath !

PR #1292, opened 2025-04-05
struct-tag: change reported messages and refactor implementation

chavacava:
This PR proposes  a ~full rewriting of the rule `struct-tag` in an attempt to clean-up the implementation.

This PR introduces the following changes:

1. Defines a common signature for tag-checker functions
2. Normalizes failure messages
3. Reduces the implementation complexity of some checks (protobuf, properties, ...)

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -175,7 +175,7 @@ func (w lintStructTagRule) checkTaggedField(ctx *checkContext, f *ast.Field) {
 
 	for _, tag := range tags.Tags() {
 		if msg, ok := w.checkTagNameIfNeed(ctx, tag); !ok {
-			w.addFailure(f.Tag, msg)
+			w.addFailureWithTagKey(f.Tag, msg, tag.Key)
 		}
 
 		checker, ok := w.tagCheckers[tagKey(tag.Key)]
@@ -185,7 +185,7 @@ func (w lintStructTagRule) checkTaggedField(ctx *checkContext, f *ast.Field) {
 
 		msg, ok := checker(ctx, tag, f.Type)
 		if !ok {
-			w.addFailure(f.Tag, msg)
+			w.addFailureWithTagKey(f.Tag, msg, tag.Key)
 		}
 	}
 }
@@ -212,7 +212,7 @@ func (w lintStructTagRule) checkTagNameIfNeed(ctx *checkContext, tag *structtag.
 	// to allow the same tag name in different tag type.
 	mapKey := tag.Key + ":" + tagName
 	if _, ok := ctx.usedTagName[mapKey]; ok {
-		return fmt.Sprintf("duplicate tag name: %q", tagName), false
+		return fmt.Sprintf(msgDuplicatedTagName, tagName), false
 	}
 
 	ctx.usedTagName[mapKey] = true
@@ -259,22 +259,22 @@ func checkCompoundANS1Option(ctx *checkContext, opt string, fieldType ast.Expr) 
 		tagNumber := strings.TrimLeft(opt, "tag:")
 		number, err := strconv.Atoi(tagNumber)
 		if err != nil {
-			return fmt.Sprintf("ASN1 tag must be a number, got %q", tagNumber), false
+			return fmt.Sprintf("tag must be a number but is %q", tagNumber), false
 		}
 		if ctx.usedTagNbr[number] {
-			return fmt.Sprintf("duplicated tag number %v", number), false
+			return fmt.Sprintf(msgDuplicatedTagNumber, number), false
 		}
 		ctx.usedTagNbr[number] = true
 	case "default":
 		if len(parts) < 2 {
-			return "malformed default for ASN1 tag", false
+			return "malformed default", false
 		}
 		if !typeValueMatch(fieldType, parts[1]) {
-			return "field type and default value type mismatch", false
+			return msgTypeMismatch, false
 		}
 	default:
 		if !ctx.isUserDefined(keyASN1, opt) {
-			return fmt.Sprintf("unknown option %q in ASN1 tag", opt), false
+			return fmt.Sprintf(msgUnknownOption, opt), false
 		}
 	}
 	return "", true
@@ -288,7 +288,7 @@ func checkDatastoreTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (messa
 			if ctx.isUserDefined(keyDatastore, opt) {
 				continue
 			}
-			return fmt.Sprintf("unknown option %q in Datastore tag", opt), false
+			return fmt.Sprintf(msgUnknownOption, opt), false
 		}
 	}
 
@@ -297,7 +297,7 @@ func checkDatastoreTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (messa
 
 func checkDefaultTag(_ *checkContext, tag *structtag.Tag, fieldType ast.Expr) (message string, succeded bool) {
 	if !typeValueMatch(fieldType, tag.Name) {
-		return "field type and default value type mismatch", false
+		return msgTypeMismatch, false
 	}
 
 	return "", true
@@ -311,7 +311,7 @@ func checkBSONTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (message st
 			if ctx.isUserDefined(keyBSON, opt) {
 				continue
 			}
-			return fmt.Sprintf("unknown option %q in BSON tag", opt), false
+			return fmt.Sprintf(msgUnknownOption, opt), false
 		}
 	}
 
@@ -325,7 +325,7 @@ func checkJSONTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (message st
 		case "":
 			// special case for JSON key "-"
 			if tag.Name != "-" {
-				return "option can not be empty in JSON tag", false
+				return "option can not be empty", false
 			}
 		case "omitzero":
 			if ctx.isAtLeastGo124 {
@@ -336,7 +336,7 @@ func checkJSONTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (message st
 			if ctx.isUserDefined(keyJSON, opt) {
 				continue
 			}
-			return fmt.Sprintf("unknown option %q in JSON tag", opt), false
+			return fmt.Sprintf(msgUnknownOption, opt), false
 		}
 	}
 
@@ -351,7 +351,7 @@ func checkMapstructureTag(ctx *checkContext, tag *structtag.Tag, _ ast.Expr) (me
 			if ctx.isUserDefined(keyMapstructure, opt) {
 				continue
 			}
-			return fmt.Sprintf("unknown option %q in Mapstructure tag", opt), false
+			return fmt.Sprintf(msgUnknownOption, opt), false
 		}
 	}
 
@@ -373,7 +373,7 @@ func checkPropertiesTag(ctx *checkContext, tag *structtag.Tag, fieldType ast.Exp
 		case 2:
 			msg, ok = checkCompoundPropertiesOption(parts[0], parts[1], fieldType, seenOptions)
 		default:
-			msg, ok = fmt.Sprintf("unknown or malformed option %q in properties tag", opt), false
+			msg, ok = fmt.Sprintf("unknown or malformed option %q", opt), false
 		}
 		if !ok {
 			return msg, false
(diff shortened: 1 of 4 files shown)


rule/struct_tag.go
```diff
@@ -16,6 +16,61 @@ type StructTagRule struct {
...
+	keyProtobuf     tagKey = "protobuf"
+	keyRequired     tagKey = "required"
+	keyTOML         tagKey = "toml"
+	keyURL          tagKey = "url"
+	keyValidate     tagKey = "validate"
+	keyXML          tagKey = "xml"
+	keyYAML         tagKey = "yaml"
+)
+
+type tagChecker func(ctx *checkContext, tag *structtag.Tag, fieldType ast.Expr) (message string, succeded bool)
```
alexandear, line 38: Please fix a typo here and in all places below:
```suggestion
type tagChecker func(ctx *checkContext, tag *structtag.Tag, fieldType ast.Expr) (message string, succeeded bool)
```

testdata/go1.24/struct_tag.go
alexandear, line 1: I think it would be better to rephrase the PR's title. This is definitely not a simple refactoring but a breaking change.

chavacava: Did you spot something this PR breaks ? If the PR breaks something it was not my intention :/

alexandear: The rule's messages have changed, and technically it's a breaking change (albeit a minor one). If someone disables this rule by text in their [golangci-lint config](https://golangci-lint.run/usage/configuration/#linters-configuration), it will break:

```yaml
linters:
  exclusions:
    rules:
      - path: internal/hmac/
        text: "field's type and default value's type mismatch"
        linters:
          - revive
```

rule/struct_tag.go
```diff
@@ -134,22 +210,23 @@ func (w lintStructTagRule) checkTagNameIfNeed(tag *structtag.Tag) (string, bool)
 
 	// We concat the key and name as the mapping key here
 	// to allow the same tag name in different tag type.
-	key := tag.Key + ":" + tagName
-	if _, ok := w.usedTagName[key]; ok {
-		return fmt.Sprintf("duplicate tag name: '%s'", tagName), false
+	mapKey := tag.Key + ":" + tagName
+	if _, ok := ctx.usedTagName[mapKey]; ok {
+		return fmt.Sprintf(msgDuplicatedTagName, tagName), false
```
alexandear, line 215: The constant `msgDuplicatedTagName` is used only once and we can remove it. Perhaps the following is more readable:
```suggestion
		return fmt.Sprintf("duplicated tag name %q", tagName), false
```

testdata/go1.24/struct_tag.go
chavacava: Messages were intentionally modified.
Indeed, modifying a text output (even a typo) can be defined as a breaking change.

rule/struct_tag.go
```diff
@@ -16,6 +16,61 @@ type StructTagRule struct {
...
+	keyRequired:     checkRequiredTag,
+	keyTOML:         checkTOMLTag,
+	keyURL:          checkURLTag,
+	keyValidate:     checkValidateTag,
+	keyXML:          checkXMLTag,
+	keyYAML:         checkYAMLTag,
+}
+
+type checkContext struct {
+	userDefined    map[string][]string // map: key -> []option
```
ccoVeille, line 59: Couldn't it help to use this here


```suggestion
	userDefined    map[tagKey][]string // map: key -> []option
```

rule/struct_tag.go
```diff
@@ -73,9 +131,8 @@ func (*StructTagRule) Name() string {
 type lintStructTagRule struct {
 	onFailure      func(lint.Failure)
 	userDefined    map[string][]string // map: key -> []option
```
ccoVeille, line 133: Here also it could be a tagKey


```suggestion
	userDefined    map[tagKey][]string // map: key -> []option
```

testdata/go1.24/struct_tag.go
ccoVeille: I think I'm the root cause of this change

I reported it in the PR about tag `properties` addition

https://github.com/mgechev/revive/pull/1289#discussion_r2028211614

rule/struct_tag.go
```diff
@@ -158,286 +235,296 @@ func (lintStructTagRule) getTagName(tag *structtag.Tag) string {
...
 		case "keys":
 			if previousOption != "dive" {
-				return "option 'keys' must follow a 'dive' option in validate tag", false
+				return `option "keys" must follow a "dive" option`, false
 			}
 			seenKeysOption = true
 		case "endkeys":
 			if !seenKeysOption {
-				return "option 'endkeys' without a previous 'keys' option in validate tag", false
+				return `option "endkeys" without a previous "keys" option`, false
```
ccoVeille, line 522: Here the quote changes is also "a breaking change" but I'm fine with it

testdata/struct_tag.go
```diff
@@ -6,39 +6,41 @@ type decodeAndValidateRequest struct {
...
 	MandatoryStruct4 mandatoryStruct `json:"mandatoryStruct4" required:"false"`
 	OptionalStruct   *optionalStruct `json:"optionalStruct,omitempty"`
 	OptionalQuery    string          `json:"-" querystring:"queryfoo"`
 	optionalQuery    string          `json:"-" querystring:"queryfoo"` // MATCH /tag on not-exported field optionalQuery/
 	// No-reg test for bug https://github.com/mgechev/revive/issues/208
 	Tiret       string `json:"-,"`
-	BadTiret    string `json:"other,"`               // MATCH /option can not be empty in JSON tag/
-	ForOmitzero string `json:"forOmitZero,omitzero"` // MATCH /unknown option 'omitzero' in JSON tag/
+	BadTiret    string `json:"other,"`               // MATCH /option can not be empty in json tag/
+	ForOmitzero string `json:"forOmitZero,omitzero"` // MATCH /unknown option "omitzero" in json tag/
```
ccoVeille, line 28: It's great to test this one, but while you work on refactoring everything, can't we add a message more like


```suggestion
	ForOmitzero string `json:"forOmitZero,omitzero"` // MATCH /option "omitzero" is unsupported in json tag prior Go 1.24/
```

testdata/go1.24/struct_tag.go
ccoVeille, line 1: I'm a bit worried by this filename

It relies on build tag
 It's OK, but then as soon as Go 1.25 will be out, this file won't be used, and the structtag_test.go will, so it will fail.

Unless I'm misunderstanding the concept of how build tags works

rule/struct_tag.go
```diff
@@ -86,42 +143,61 @@ func (w lintStructTagRule) Visit(node ast.Node) ast.Visitor {
...
 		}
 
-		w.usedTagNbr = map[int]bool{}
-		w.usedTagName = map[string]bool{}
+		ctx := &checkContext{
+			userDefined:    w.userDefined,
+			usedTagNbr:     map[int]bool{},
+			usedTagName:    map[string]bool{},
+			isAtLeastGo124: w.isAtLeastGo124,
+		}
```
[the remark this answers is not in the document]
chavacava: I had the same feeling but then I thought "why that pretty name will be for exclusive property of the standard library context? We also have contexts!" (I must admit my argument might have a little bit of influence from "Le prenom" that I (re)watched last we)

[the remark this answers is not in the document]
ccoVeille: Good movie, and good counter example. I get your point.

But, I'm not that convinced. I'm curious about other maintainer point of view anyway

@alexandear @denisvmedia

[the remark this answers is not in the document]
chavacava: Don't worry, I will rename into checkCtx

testdata/go1.24/struct_tag.go
chavacava: I don't get it. The PR doesn't modify the filename and the filename follows the convention of testdata files in `revive`: `<rule_name>(_other_details)?.go`

BTW: there is also `testdata/struct_tag.go`

rule/struct_tag.go
```diff
@@ -158,286 +235,292 @@ func (lintStructTagRule) getTagName(tag *structtag.Tag) string {
...
-	}
+	return "", true
+}
 
-	for _, tag := range tags.Tags() {
-		if msg, ok := w.checkTagNameIfNeed(tag); !ok {
-			w.addFailure(f.Tag, msg)
+func checkCompoundANS1Option(checkCtx *checkContext, opt string, fieldType ast.Expr) (message string, succeeded bool) {
+	parts := strings.Split(opt, ":")
+	switch parts[0] {
```
ccoVeille, line 257: Here also, using strings.Cut would help for readability

It would avoid code like


                tagNumber := strings.TrimLeft(opt, "tag:")


                if len(parts) < 2 { 
                         return "malformed default", false 
                 }


And the use of parts[0] and parts[1]

rule/struct_tag.go
```diff
@@ -158,286 +235,292 @@ func (lintStructTagRule) getTagName(tag *structtag.Tag) string {
...
+	hasName := false
 	for _, opt := range options {
-		switch opt {
-		case "int", "omitempty", "numbered", "brackets":
-		case "unix", "unixmilli", "unixnano": // TODO : check that the field is of type time.Time
-		case "comma", "semicolon", "space":
-			if delimiter == "" {
-				delimiter = opt
-				continue
+		opt := strings.Split(opt, "=")[0]
```
ccoVeille, line 422:
```suggestion
		opt := strings.CutSuffix(opt, "=")
```

rule/struct_tag.go
```diff
@@ -158,286 +235,292 @@ func (lintStructTagRule) getTagName(tag *structtag.Tag) string {
...
 }
 
-func (w lintStructTagRule) checkDatastoreTag(options []string) (string, bool) {
-	for _, opt := range options {
+func checkRequiredTag(_ *checkContext, tag *structtag.Tag, _ ast.Expr) (message string, succeeded bool) {
+	if tag.Name != "true" && tag.Name != "false" {
+		return `required should be "true" or "false"`, false
+	}
+
+	return "", true
```
ccoVeille, line 464: 🎨 A switch would be more readable, I think

rule/struct_tag.go
```diff
@@ -448,47 +531,62 @@ func (w lintStructTagRule) checkValidateTag(options []string) (string, bool) {
...
+}
+
+func checkValidateOptionsAlternatives(checkCtx *checkContext, alternatives []string) (message string, succeeded bool) {
 	for _, alternative := range alternatives {
 		alternative := strings.TrimSpace(alternative)
 		parts := strings.Split(alternative, "=")
 		switch len(parts) {
 		case 1:
 			badOpt, ok := areValidateOpts(parts[0])
```
ccoVeille, line 570: strings.Cut would help here

testdata/go1.24/struct_tag.go
alexandear: I propose to change the PR's title to `struct-tag: change reported messages and refactor implementation`.

alexandear: > When Go 1.25 will go out, the tests will call testdata/struct_tag.go, right?

When Go 1.24 will go out, the tests will call all tests from [`test/struct_tag_test.go`](https://github.com/mgechev/revive/blob/f08317847bb858db10377cbab49b3c1f2a44101e/test/struct_tag_test.go#L1-L30):

- testdata/struct_tag.go
- testdata/struct_tag_user_options.go
- testdata/go1.24/struct_tag.go

rule/struct_tag.go
ccoVeille: Indeed, it's not as obvious as other ones

rule/struct_tag.go
```diff
@@ -512,115 +605,19 @@ func (lintStructTagRule) typeValueMatch(t ast.Expr, val string) bool {
...
-				continue
-			}
-			return fmt.Sprintf("unknown option '%s' in protobuf tag", k), false
-		}
-	}
-
-	return "", true
+func (w lintStructTagRule) addFailureWithTagKeyf(n ast.Node, msg string, tagKey string) {
+	w.addFailuref(n, "%s in %s tag", msg, tagKey)
 }
```
ccoVeille, line 611: It should be this

```suggestion
func (w lintStructTagRule) addFailureWithTagKey(n ast.Node, msg string, tagKey string) {
	w.addFailuref(n, "%s in %s tag", msg, tagKey)
}
```

There is no extra param sent to addFailureWithTagKey

Unless if you rewrite it like this

```suggestion
func (w lintStructTagRule) addFailureWithTagKey(n ast.Node, tagKey string, msg string, args ...any) {
	msg += fmt.Sprintf(" in %s tag", tagKey)
	w.addFailuref(n, msg, args)
}
```

But I'm unsure there is a need for it

PR #1295, opened 2025-04-05
exported: correctly report deprecation-only comments

### LINKED ISSUE #1231: exported: comment starting with "Deprecated:” should be ignored when checking if a comment is present
```go

// Foo does the magic
func Foo() {}

// Bar is blah blah
//
// Deprecated: use Foo
func Bar() {}

//nolint:whatever
func Baz() {}

// Deprecated: is deprecated
func Qux() {}

func Quux() {}
```

The first two are valid

The third one is fixed by a previous PR:
- #1209

And is now reported as: "exported func Baz should have comment or be unexported"

The last one is reported as: "exported func Quux should have comment or be unexported"

The issue is with `Qux`, the error reported is "comment should be like Qux ...", but in fact here it should be "exported function should have a comment", the very same way `Baz` is reported

semihbkgr:
Closes #1231

Fixed the exported rule to correctly handle and exclude deprecation-only comments.

Previously, when only a deprecation comment was present, it incorrectly reported:
`comment on exported type ... should be of the form`; 
now it reports: 
`exported type ... should have comment or be unexported.`

### DIFF
--- a/rule/exported.go
+++ b/rule/exported.go
@@ -182,7 +182,7 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
 		return
 	}
 
-	if !hasTextComment(fn.Doc) {
+	if !hasDocComment(fn.Doc) {
 		w.onFailure(lint.Failure{
 			Node:       fn,
 			Confidence: 1,
@@ -247,7 +247,7 @@ func (w *lintExported) lintTypeDoc(t *ast.TypeSpec, doc *ast.CommentGroup) {
 		return
 	}
 
-	if !hasTextComment(doc) {
+	if !hasDocComment(doc) {
 		w.onFailure(lint.Failure{
 			Node:       t,
 			Confidence: 1,
@@ -314,7 +314,7 @@ func (w *lintExported) lintValueSpecDoc(vs *ast.ValueSpec, gd *ast.GenDecl, genD
 		return
 	}
 
-	if !hasTextComment(vs.Doc) && !hasTextComment(gd.Doc) {
+	if !hasDocComment(vs.Doc) && !hasDocComment(gd.Doc) {
 		if genDeclMissingComments[gd] {
 			return
 		}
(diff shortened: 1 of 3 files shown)


chavacava:
Hi @semihbkgr thanks for the PR!

I've added a test case to the PR. The case is that of a function with a deprecation comment line followed by a wrong comment line (`Minimum ...` where `Min ...` is expected). The expected failure message for that case is `comment on exported function Min should be of the form "Min ..."`

semihbkgr:
Hi @chavacava

I assumed that everything following the `Deprecated:` tag is part of the deprecation comment. However, I'm unsure whether placing the deprecation comment before the doc comment—as done in the new test case you added—is correct. I couldn’t find any examples of this ordering in popular codebases. Typically, the doc comment appears first, followed by the deprecation comment. I'm not certain whether this is a formal rule or just a widely followed convention.
Also, if the doc comment comes after the deprecation comment, it becomes unclear how to distinguish between the deprecation comment and the doc comment.

chavacava:
The official [doc on deprecation comments](https://go.dev/doc/comment) says:

> Deprecation notices are followed by some information about the deprecation, and a recommendation on what to use instead, if applicable. The paragraph does not have to be the last paragraph in the doc comment.

```go
// Package rc4 implements the RC4 stream cipher.
//
// Deprecated: RC4 is cryptographically broken and should not be used
// except for compatibility with legacy systems.
//
// This package is frozen and no new functionality will be added.
package rc4
```

Let's assume that the deprecation "section" is always after the "documentation" section.
Feel free to modify the test case I've added to the PR.

semihbkgr:
So we can assume that the doc comment comes before the deprecation comment - this isn’t a strict rule, but it’s a commonly followed convention.
Alternatively, we can treat only the paragraph starting with the `Deprecated:` tag as the deprecation comment, and consider the following paragraphs as part of the doc comment.
This would be another assumption, but it allows us to handle cases where the doc comment appears after the deprecation comment. For example:

```go
// Deprecated: this is deprecated, use min instead -> deprecation comment
//                                                                                   -> end of paragraph
// Min returns a if a <= b, b otherwise.                      -> doc comment
func Min(a, b int) int {
```

chavacava:
Let's keep it simple and stick to the convention (comment + deprecation)

ccoVeille:
Thank you guys for working on this. I was out with family, I only review it now.

PR #1299, opened 2025-04-06
Refactor/exported

chavacava:
IMO the rule deserved a refactoring to keep it under control.

The main modification introduced by this PR is the replacement of `hasDocComment`, `normalizeText` and `isOnlyDeprecationComment` by a single method `firstCommentLine`. This replacement is based on the fact that the only thing of interest when doing the check on comments is the first line of the actual documentation; that is the first line of comment that is neither a directive nor a deprecation comment block.

The PR introduces other changes to reduce the complexity of some methods and gain in clarity.

### DIFF
--- a/rule/exported.go
+++ b/rule/exported.go
@@ -19,7 +19,7 @@ type disabledChecks struct {
 	Method           bool
 	PrivateReceivers bool
 	PublicInterfaces bool
-	Stuttering       bool
+	RepetitiveNames  bool
 	Type             bool
 	Var              bool
 }
@@ -46,7 +46,7 @@ func (dc *disabledChecks) isDisabled(checkName string) bool {
 	case checkNamePublicInterfaces:
 		return dc.PublicInterfaces
 	case checkNameStuttering:
-		return dc.Stuttering
+		return dc.RepetitiveNames
 	case "type":
 		return dc.Type
 	default:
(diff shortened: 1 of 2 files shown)


rule/exported_test.go
```diff
@@ -170,8 +170,8 @@ func TestExportedRule_Configure(t *testing.T) {
 			if rule.disabledChecks != tt.wantDisabledChecks {
 				t.Errorf("unexpected disabledChecks: got = %+v, want %+v", rule.disabledChecks, tt.wantDisabledChecks)
 			}
-			if rule.stuttersMsg != tt.wantStuttersMsg {
-				t.Errorf("unexpected stuttersMsg: got = %v, want %v", rule.stuttersMsg, tt.wantStuttersMsg)
+			if rule.isRepetitiveMsg != tt.wantStuttersMsg {
```
ccoVeille, line 173: I would rename the wantStuttersMsg too

rule/exported.go
```diff
@@ -155,69 +151,51 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
...
-			}
-		}
+		kind = "method"
+		recv := typeparams.ReceiverType(fn)
 		name = recv + "." + name
 	}
 
 	if w.disabledChecks.isDisabled(kind) {
 		return
 	}
```
ccoVeille, line 166: Shouldn't this if be moved before the previous `if` ?

chavacava: I asked very same question to myself when refactoring... I didn't find a nice way of moving the `if` because it depends on `kind`  that is potentially set in the body of previous one.

ccoVeille (approved):

rule/exported.go
```diff
@@ -446,23 +422,51 @@ func (w *lintExported) lintInterfaceMethod(typeName string, m *ast.Field) {
...
+}
+
+func (w *lintExported) addFailure(node ast.Node, confidence float64, category lint.FailureCategory, message string) {
+	w.onFailure(lint.Failure{
+		Node:       node,
+		Confidence: confidence,
+		Category:   category,
+		Failure:    message,
+	})
 }
```
denisvmedia, line 472: WDYT?

```suggestion
func (w *lintExported) addFailure(node ast.Node, confidence float64, category lint.FailureCategory, message string, args ...string) {
	w.onFailure(lint.Failure{
		Node:       node,
		Confidence: confidence,
		Category:   category,
		Failure:    fmt.Sprintf(message, args...),
	})
}
```

ccoVeille: I'm down for it. I would like to suggest renaming the method to addFailuref 

With the logic of sprintf method should ends with an f

chavacava: Indeed!

Issue #1277, opened 2025-03-19
Some rules don't work as expected in some environments


ldez:
**Describe the bug**

The following rules don't work as expected in some environments:
- `context-keys-type`
- `errorf`
- `modifies-value-receiver`
- `range-val-address`
- `string-of-int`
- `time-equal`
- `time-naming`
- `unhandled-error`
- `var-declaration`

**To Reproduce**

```bash
cd /tmp
git clone -branch revive https://github.com/ldez/fiber.git
cd fiber

cat > revive.toml  <<EOF
[rule.unhandled-error]
EOF

# version 1.7.0 (I used the official binary)
revive --version

# Output: 24
revive -config=revive.toml ./... | wc -l

# Output: 3
GOROOT= revive -config=revive.toml ./... | wc -l
```

```toml
[rule.unhandled-error]
```

**Expected behavior**

I expect the rules to be less dependent on the environment.

**Logs**

NA

**Desktop (please complete the following information):**
 - OS: Linux
 - go 1.24

**Additional context**

The problem relates to `file.Pkg.TypeCheck()`: the default importer uses `GOROOT` and `GOPATH`.

If you log the error returns by `file.Pkg.TypeCheck()`, you can see what happened.

Note, `go env GOROOT` always returns a value even if the env var is not set.

For example, GitHub Action runners don't have an explicit env var `GOROOT`.

On GitHub Action runners, the problem is increased if the binary is built with `-trimpath`.
For now, I haven't found a way to reproduce the same environment locally as GitHub Action runners, but I'm using an SSH connection to a runner to test it.

```console
runner@fv-az1337-715:~/work/fiber/fiber$ go version
go version go1.24.1 linux/amd64
runner@fv-az1337-715:~/work/fiber/fiber$
runner@fv-az1337-715:~/work/fiber/fiber$ env | grep GOROOT
GOROOT_1_21_X64=/opt/hostedtoolcache/go/1.21.13/x64
GOROOT_1_23_X64=/opt/hostedtoolcache/go/1.23.7/x64
GOROOT_1_22_X64=/opt/hostedtoolcache/go/1.22.12/x64
runner@fv-az1337-715:~/work/fiber/fiber$
runner@fv-az1337-715:~/work/fiber/fiber$ go env GOROOT
/opt/hostedtoolcache/go/1.24.1/x64
runner@fv-az1337-715:~/work/fiber/fiber$
runner@fv-az1337-715:~/work/fiber/fiber$ cd /tmp
[...]

PR #1308, opened 2025-04-09
fix(unexported-return): add missing TypeCheck call

### LINKED ISSUE #1306: Some reports are "randomly" missing
**Describe the bug**

Some reports are "randomly" missing

**To Reproduce**

```bash
git clone -q git@github.com:prometheus/prometheus.git
cd prometheus
git checkout 701d13abf
```

```console
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
720
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
723
$ revive ./... | wc -l
720
$ revive ./... | wc -l
723
...
```

As you can see, sometimes there are 723 reports and sometimes there are only 720 reports.

**Expected behavior**

Report consistency.

**Additional context**

For now, I didn't find the root cause but I think the problem is related to a concurrency issue.
I will try to find the root cause.

ldez:
**This PR needs PR #1307.**

The problem of consistency was related to a missing call to `Package.TypeCheck()`.

The `UnexportedReturnRule` calls `Package.TypeOf()`, but this requires the data from `Package.TypeCheck()`: depending on the order of the rule, the data may exist or may not.

I checked the calls to `Package.TypeOf()`, and it is only missing for `unexported-return`.

The bad news is that `unexported-return` is now a part of the [other issue](https://github.com/mgechev/revive/issues/1277) about `Package.TypeCheck()`.
Concretely, it was already a part of this issue as it relies on `Package.TypeOf()`.

The regression comes from [#1170](https://github.com/mgechev/revive/pull/1170/files) (v1.6.0); this explains why the number of issues has increased since the beginning of the year.

Note: I also fixed a problem with a missing call to `types.Unalias()`.

Fixes #1306

Note: numbers have changed (723 -> 722) because I replaced `wc` with `jq` (so no extra empty line).

```console
$ ./revive-bench.sh
722
722
722
722
...
722
722
722
```

<details>
<summary>revive-bench.sh</summary>

```bash
#!/bin/bash -e

for a in {1..1000}; do
    revive -formatter json ./... | jq '. |length'
done

```

</details>

### DIFF
--- a/rule/unexported_return.go
+++ b/rule/unexported_return.go
@@ -16,6 +16,8 @@ type UnexportedReturnRule struct{}
 func (*UnexportedReturnRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 	var failures []lint.Failure
 
+	file.Pkg.TypeCheck()
+
 	for _, decl := range file.AST.Decls {
 		fn, ok := decl.(*ast.FuncDecl)
 		if !ok {
@@ -70,7 +72,7 @@ func (*UnexportedReturnRule) Name() string {
 // It is imprecise, and will err on the side of returning true,
 // such as for composite types.
 func exportedType(typ types.Type) bool {
-	switch t := typ.(type) {
+	switch t := types.Unalias(typ).(type) {
 	case *types.Named:
 		obj := t.Obj()
 		switch {


ccoVeille:
Thank you so much for the investigation, time spent on it, and multiple fixes you provided

ldez:
@chavacava, sorry to ask that, but do you plan to create a release soon?

Before updating golangci-lint with a pseudo version to revive, I prefer to check with you.

ldez:
ok, so I will use a pseudo-version. Thank you for the answer.

ldez:
Thank you for the release.

I didn't want to force a version to be created, so thank you again :heart:

PR #1309, opened 2025-04-10
chore: remove edited trigger

ldez:
The `edited` event is related to the edition of the title or the description of a PR.
I think you don't need to re-launch the CI after an edition of a title.


chavacava:
Thanks! Nice catch. I didn't notice checks were triggered when modifying the PR title.

PR #998, opened 2024-06-03
Fix a performance regression #995

dominiquelefevre:
* Support go workspaces when detecting the go version.
* Do not invoke go list for every package.

Closes  #995

PR #1316, opened 2025-04-13
refactor: change usage of package `hashicorp/go-version` to `go/version`

vbvictor:
Refactor:
- Changed usage of `hashicorp/go-version` package to `go/version`.
- Removed `goVersion` field from `config.Config` because it was no longer used in `.toml` configuration file.

Closes https://github.com/mgechev/revive/issues/1302.


lint/config.go
```diff
@@ -65,7 +61,4 @@ type Config struct {
 	WarningCode           int              `toml:"warningCode"`
 	Directives            DirectivesConfig `toml:"directive"`
 	Exclude               []string         `toml:"exclude"`
-	// If set, overrides the go language version specified in go.mod of
-	// packages being linted, and assumes this specific language version.
-	GoVersion *goversion.Version
```
alexandear, line 70: This field shouldn't be deleted. It's an important feature added in #998

alexandear (changes requested):

lint/config.go
denisvmedia: I'd also put a comment there, to avoid future confusions.

vbvictor: I've looked at original PR, hence the question: Can we break compatibility for consumers revive? Mainly speaking about `golangci-lint`. I can make a PR in `golangci-lint` when a new version of `revive` is released to address the changes, but I don't know about any other users of `lint.Config`.
 
For golangci-lint it must be a 1-line change
 https://github.com/golangci/golangci-lint/blob/3afce899b2a3b477168b847c34d565c62d4fad18/pkg/golinters/revive/revive.go#L93.

alexandear: I found another usage in public code:
https://github.com/pingcap/tidb/blob/74b342591b107495c0f34fd334cb3de3e432d491/build/linter/revive/analyzer.go#L105

alexandear: Maybe it's better to deprecate this field for now and remove it after a few revive releases.

lint/package.go
```diff
@@ -36,13 +36,13 @@ var (
...
+	Go115 = "go1.15"
 	// Go121 is a constant representing the Go version 1.21
-	Go121 = goversion.Must(goversion.NewVersion("1.21"))
+	Go121 = "go1.21"
 	// Go122 is a constant representing the Go version 1.22
-	Go122 = goversion.Must(goversion.NewVersion("1.22"))
+	Go122 = "go1.22"
 	// Go124 is a constant representing the Go version 1.24
-	Go124 = goversion.Must(goversion.NewVersion("1.24"))
+	Go124 = "go1.24"
```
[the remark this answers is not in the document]
vbvictor: If we break so many public API without significant boost to performance/binsize/etc, should we even refactor it?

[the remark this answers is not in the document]
ccoVeille: > Now, I think towards keeping `hashicorp/go-version` as is.

Yes, I also feel the same, I'm pretty sad about it.

But as @alexandear said they could be marked deprecated and removed in a few version, with a breaking change, but without a major

Issue #1313, opened 2025-04-11
Remove regex from code when possible


ccoVeille:
Regex are slow var-naming contains a few of them

```
var anyCapsRE = regexp.MustCompile(`[A-Z]`)

var allCapsRE = regexp.MustCompile(`^[A-Z0-9_]+$`)

// regexp for constant names like `SOME_CONST`, `SOME_CONST_2`, `X123_3`, `_SOME_PRIVATE_CONST` (#851, #865)
var upperCaseConstRE = regexp.MustCompile(`^_?[A-Z][A-Z\d]*(_[A-Z\d]+)*$`)
```
https://github.com/mgechev/revive/blob/v1.8.0/rule%2Fvar_naming.go#L13-L18

They could be replaced by function that will use simple checker based on the rune.

This was found when working on

- https://github.com/mgechev/revive/pull/1312#discussion_r2037863798

I found some other usages of regex in the repository, some can be replaced, some cannot as there are part of configuration variables.

We could start with var-naming and do a benchmark.

I'm suggesting this because var-naming is enabled by default and lint almost everything in the code, while some rules are more specialized. So I think a small optimization here could be useful. But it still has to be benchmarked on large code base.

ccoVeille:
This one will require the pending PR on var-naming is merged

- #1312

PR #1322, opened 2025-04-19
var-naming: remove regex from code

mfederowicz:
related with #1313 

replaced regexp with functions eqivalent, and add some tests and benchmarks

```
go test -bench=.
goos: linux
goarch: amd64
pkg: github.com/mgechev/revive/rule
cpu: 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
BenchmarkAnyCapsRE-8            35753906                34.16 ns/op
BenchmarkAnyCaps-8              1000000000               0.8155 ns/op
BenchmarkAllCapsRE-8             4175530               311.3 ns/op
BenchmarkAllCaps-8              89929506                13.28 ns/op
BenchmarkUpperCaseConstRE-8      9525739               133.2 ns/op
BenchmarkIsUpperCaseConst-8     211988328                5.434 ns/op
PASS
ok      github.com/mgechev/revive/rule  9.191s
```
as we can see above, results of benchmarks using regexp are much slower (benchmarks with RE suffix)

ps. sorry for previous versions of that PR, but I had old not synced master on my fork :(


chavacava (commented):
Hi @mfederowicz, thanks for the PR.
I've left some comments in the code.
Some general remarks and questions:
1. As benchmarks show, and it was expected, RE are slower than the new functions, that is a fact. How much faster is the rule when it uses the new functions?
2. The standard package [unicode](https://pkg.go.dev/unicode) provides utility functions like `IsDigit`, `IsLower`, `IsUpper` ... We could use it (even if it introduces a small runtime overhead)

ccoVeille:
> The standard package unicode provides utility functions like IsDigit, IsLower, IsUpper ... 

I'm fine with this. But then all characters from non-latin languages (Greek,  Russian, CJK…) would be allowed.

I'm fine with this, as I said by opening:
- https://github.com/mgechev/revive/issues/1315

rule/var_naming.go
```diff
@@ -197,12 +189,12 @@ func (w *lintNames) check(id *ast.Ident, thing string) {
...
 	// #851 upperCaseConst support
 	// if it's const
-	if thing == token.CONST.String() && w.upperCaseConst && upperCaseConstRE.MatchString(id.Name) {
+	if thing == token.CONST.String() && w.upperCaseConst && isUpperCaseConst(id.Name) {
 		return
 	}
 
 	// Handle two common styles from other languages that don't belong in Go.
-	if len(id.Name) >= 5 && allCapsRE.MatchString(id.Name) && strings.Contains(id.Name, "_") {
+	if len(id.Name) >= 5 && allCaps(id.Name) && strings.Contains(id.Name, "_") {
```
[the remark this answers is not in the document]
mfederowicz: maybe it is related with that we want to check const longer than 4 char+'_' prefix? just guesing :)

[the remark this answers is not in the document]
mfederowicz: i changed to ```if len(id.Name) >= 5 && isUpperUnderScore(id.Name)```  for more readability :)

rule/utils.go
```diff
@@ -140,3 +140,77 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+	for _, r := range s {
+		if (r < 'A' || r > 'Z') && (r < '0' || r > '9') && r != '_' {
+			return false
+		}
+	}
+	return true
+}
+
+// isUpperCaseConst checks if string is in constant name format like `SOME_CONST`, `SOME_CONST_2`, `X123_3`, `_SOME_PRIVATE_CONST` (#851, #865)
+func isUpperCaseConst(s string) bool {
```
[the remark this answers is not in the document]
mfederowicz: in new functions change caps to upper :)

mfederowicz:
> 1. As benchmarks show, and it was expected, RE are slower than the new functions, that is a fact. How much faster is the rule when it uses the new functions?

maybe using ```time``` for measuring execution time is not the best, but i notice something interesting (increase number of warnings), when i switch revive version from 1.9.0 to 1.9.0-2-g17527c9 (builded from git source: master)

results from test executions on kubernetes master branch (https://github.com/kubernetes/kubernetes)

test command: ```time revive --formatter friendly ./...```

version 1.8.0: installed from go install -x github.com/mgechev/revive@v1.8.0
upperCaseConst=false
Warnings:1928  var-naming  

real    0m2.311s
user    0m6.862s
sys     0m0.757s

upperCaseConst=true
Warnings:1923  var-naming  

real    0m2.376s
user    0m6.930s
sys     0m0.788s

version:1.9.0: installed from go install -x github.com/mgechev/revive@latest
upperCaseConst=false
Warnings:1928  var-naming  
real    0m2.459s
user    0m6.971s
sys     0m0.837s

upperCaseConst=true
Warnings:1923  var-naming  

real    0m2.360s
user    0m6.877s
sys     0m0.761s


version dev:v1.9.0-2-g17527c9 (build from git:master)
upperCaseConst=false
Warnings: 2131  var-naming  

real    0m2.382s
user    0m6.998s
sys     0m0.765s

upperCaseConst=true
Warnings: 2126  var-naming  

real    0m2.328s
user    0m6.861s
sys     0m0.715s

version dev:v1.9.0-4-g5f63cc2 (build from git branch without regexp)
upperCaseConst=false
Warnings: 2131  var-naming  

real    0m2.196s
user    0m6.726s
sys     0m0.732s

upperCaseConst=true
Warnings: 2126  var-naming  

real    0m2.412s
user    0m6.836s
sys     0m0.723s

chavacava:
> maybe using time for measuring execution time is not the best

[hyperfine](https://github.com/sharkdp/hyperfine) might be the right tool for this kind of benchmarks.

>  i notice something interesting (increase number of warnings), 
> when i switch revive version from 1.9.0 to 1.9.0-2-g17527c9

PR #1312 (merged after 1.9.0) adds new failures

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+	for _, r := range s {
+		if r >= '0' && r <= '9' {
+			continue // skip digits
+		}
+
+		if r >= 'a' && r <= 'z' {
+			return true
+		}
+	}
+	return false
```
ccoVeille, line 173:
```suggestion
	for _, r := range s {
		if r >= 'a' && r <= 'z' {
			return true
		}
	}
	return false
```

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+	}
+
+	// Must start with an uppercase letter
+	if s[i] < 'A' || s[i] > 'Z' {
+		return false
+	}
+	i++
+
+	// Continue parsing the rest
+	i = 0 // we can remove all the preceding code
```
ccoVeille, line 199: I don't get the `i++` followed by `i = 0`

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+		i++
+		if i >= len(s) {
+			return false // underscore only is not valid
+		}
+	}
+
+	// Must start with an uppercase letter
+	if s[i] < 'A' || s[i] > 'Z' {
+		return false
+	}
```
ccoVeille, line 195: Here you should use isUpper

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+
+	i := 0
+
+	// Optional leading underscore
+	if s[0] == '_' {
+		i++
+		if i >= len(s) {
+			return false // underscore only is not valid
+		}
+	}
```
ccoVeille, line 190:
```suggestion
	if s[0] == '_' {
		if s == "_" {
			return false // underscore only is not valid
		}
		i++
	}
```

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+	// Must start with an uppercase letter
+	if s[i] < 'A' || s[i] > 'Z' {
+		return false
+	}
+	i++
+
+	// Continue parsing the rest
+	i = 0 // we can remove all the preceding code
+	for i < len(s) {
+		c := s[i]
```
ccoVeille, line 201: Using i index on a string is a bad idea, it could lead to issue with non A-Za-z character

I would recommend using

```suggestion
	for i, c := range s {
```

This way it loops on rune

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+				return false
+			}
+		} else if !isUpperOrDigit(c) {
+			return false
+		}
+		i++
+	}
+
+	return true
+}
```
ccoVeille, line 218: Please consider this implementation

```suggestion
// isUpperCaseConst checks if string is in constant name format like `SOME_CONST`, `SOME_CONST_2`, `X123_3`, `_SOME_PRIVATE_CONST` (#851, #865)
func isUpperCaseConst(s string) bool {
	if len(s) == 0 {
		return false
	}

	r := []rune(s)
	if !IsUpper(r[0]) {
    // Must start with an uppercase letter
		return false
	}

	for i, c := range r {
		if c == '_' {
			if i +1 >= len(s) {
				// underscore is the last character
				return false
			}
			// Underscore must be followed by at least one uppercase letter or digit
			if !isUpperOrDigit(r[i+1]) {
				return false
			}
			continue
		}
		if !isUpperOrDigit(c) {
			return false
		}
	}

	return true
}
```

Pseudo code written on a phone

Or this one

```suggestion
// isUpperCaseConst checks if string is in constant name format like `SOME_CONST`, `SOME_CONST_2`, `X123_3`, `_SOME_PRIVATE_CONST` (#851, #865)
func isUpperCaseConst(s string) bool {
	if len(s) == 0 {
		return false
	}

	r := []rune(s)
	if !IsUpper(r[0]) {
    // Must start with an uppercase letter
		return false
	}

	for i, c := range r {
		switch {
		case isUpperOrDigit(c):
			continue
		
		case c == '_':
			// Underscore must be followed by at least one uppercase letter or digit
			if i +1 >= len(s) || !isUpperOrDigit(r[i+1]) {
				return false
			}

		default:
			return false
	}

	return true
}
```

rule/utils.go
ccoVeille: Please take a look at the implementation I suggested

https://github.com/mgechev/revive/pull/1322#discussion_r2051750662

rule/utils.go
ccoVeille: Please take a look at the implementation I suggested

https://github.com/mgechev/revive/pull/1322#discussion_r2051750662

rule/utils.go
ccoVeille: Please take a look at the implementation I suggested

https://github.com/mgechev/revive/pull/1322#discussion_r2051750662

rule/utils.go
ccoVeille: https://github.com/mgechev/revive/pull/1322#discussion_r2051750662

rule/utils.go
```diff
@@ -140,3 +140,84 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+
+// isUpperOrDigit checks if character is capital or digit
+func isUpperOrDigit(c byte) bool {
+	return (c >= 'A' && c <= 'Z') || (c >= '0' && c <= '9')
+}
```
ccoVeille, line 223:
```suggestion
// isUpperOrDigit checks if character is capital or digit
func isUpperOrDigit(c rune) bool {
	return isUpper(c) || (c >= '0' && c <= '9')
}
```

ccoVeille (changes requested):

rule/utils_test.go
```diff
@@ -30,3 +31,119 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+
+		{"X509B", true},
+		{"FOO", true},
+		{"_FOO123_BAR456", true},
+		{"A1_B2_C3", true},
+		{"A1_b2", false},
+		{"__FOO", false},
+		{"FOO_", false},
+		{"foo", false},
+		{"_", false},
```
ccoVeille, line 124: Some edge cases that the code handles should be tested

```suggestion
		{"X509B", true},
		{"FOO", true},
		{"1FOO", false},
		{"_FOO123_BAR456", true},
		{"A1_B2_C3", true},
		{"A1_b2", false},
		{"__FOO", false},
		{"FOO_", false},
		{"foo", false},
		{"_", false},
		{"__", false},
		{"", false},
```

rule/utils_test.go
```diff
@@ -30,3 +31,122 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+		{"SOME_CONST_2", true},
+		{"__FOO", false},
+		{"__", false},
+		{"X509B", true},
+		{"FOO", true},
+		{"1FOO", false},
+		{"_FOO123_BAR456", true},
+		{"A1_B2_C3", true},
+		{"A1_b2", false},
+		{"FOO_", false},
```
ccoVeille, line 124: About this one @chavacava @alexandear don't you think we could consider it's an uppercase constants ?

mfederowicz: @ccoVeille did you think that ```A1_b2``` or ```FOO_``` should be processed as a proper consts?

ccoVeille: I don't know. But I think it's worth raising the point.

If we had to make a detector about the fact something is in exact screamimg snake case, I would say these are invalid.

But then, here we are dealing with code that are not respecting Go standard.

Here it's a linter rule, I would report them


Also think about these technology acronym

- 3DES
- 386
- 2FA

Why wouldn't we report the following constants as uppercase constants

- 3DES_CIPHER
- 386_ARCH
- 2FA_TTL

While the following are reported

- CIPHER_3DES
- ARCH_386
- TTL_2FA

Please note these could be part of a separate issue and PR.

I'm raising these points because I think about it while reviewing the PR.

rule/var_naming.go
```diff
@@ -234,6 +226,10 @@ func (w *lintNames) check(id *ast.Ident, thing string) {
 	})
 }
 
+func isUpperUnderScore(s string) bool {
+	return !hasLowerCaseLetter(s) && strings.Contains(s, "_")
+}
```
ccoVeille, line 231: There is a problem here, now you moved to Unicode function you will face problems

Some unicode letter are neither upper or lower case, so not being lower don't make it upper case.

This is true for at least CJK and Arabic

But it's also true with digits

```suggestion
// isUpperUnderScore detects variable that are made from upper case letters, underscore, or digits.
//
// short variable names are considered OK
func isUpperUnderScore(s string) bool {
	if !strings.Contains(s, "_") {
		return false
	}

	if len(s) <= 5 {
		// avoid false positive
		return false
	}

	for _, r := range s { 
		if r == '_' || isUpperOrDigit(r) {
			continue
		}
		return false
	}
	return true
}
```

This change includes my suggestion to check the length in the method: https://github.com/mgechev/revive/pull/1322#discussion_r2054390869

If you discard it, you will have to check s is not empty or underscore

rule/var_naming.go
```diff
@@ -234,6 +226,10 @@ func (w *lintNames) check(id *ast.Ident, thing string) {
 	})
 }
 
+func isUpperUnderScore(s string) bool {
```
ccoVeille, line 229: Please test this method with

- _
- empty string
- 404_404
- FOO_BAR
- FO
- F_O
- _FOOBAR
- FOOBAR_
- FOO123
- FOO_123

ccoVeille (changes requested):

rule/var_naming.go
mfederowicz: after changes testset for ```isUpperUnderScore``` looks like this:

```
{"_", false},
{"", false},
{"empty string", false},
{"_404_404", true},
{"FOO_BAR", true},
{"FOOBAR", false},
{"FO", false},
{"F_O", false},
{"_FOOBAR", true},
{"FOOBAR_", true},
{"FOO123", false},
{"FOO_123", true},
```
of course for ```isUpperCaseConst``` some cases from above testset especially:

```
{"FOOBAR", true},
{"FO", true},
{"F_O", true},
{"FOO123", true}
```
results as true

rule/utils.go
```diff
@@ -140,3 +141,81 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+
+// hasLowerCaseLetter checks if string contains at least one lower case letter
+func hasLowerCaseLetter(s string) bool {
+	for _, r := range s {
+		if unicode.IsLower(r) {
+			return true
+		}
+	}
+	return false
+}
```
ccoVeille, line 163: now you refactored, is this used except in the tests?

Otherwise, it can be removed

 ```suggestion
```

File: rule/var_naming.go (hunk 13 above)
Comment (ccoVeille) [reply rule/var_naming.go]: Thanks for adding these tests.

They are valuable for any future refactoring

**Review by ccoVeille — COMMENTED** [review]
Thanks for dealing with my multiple feedbacks.

I appreciate.

One last thing to check and we are good for me

https://github.com/mgechev/revive/pull/1322#discussion_r2054614427

File: rule/utils.go (hunk 14 above)
Comment (mfederowicz) [reply rule/utils.go]: @ccoVeille good point, hasLowerCaseLetter removed :P

File: rule/var_naming.go (hunk 13 above)
Comment (mfederowicz) [reply rule/var_naming.go]: no problem :P

**Review by ccoVeille — APPROVED** [review]

File: rule/utils.go [hunk 15]
```diff
@@ -140,3 +141,71 @@ func isCallToExitFunction(pkgName, functionName string) bool {
 func newInternalFailureError(e error) []lint.Failure {
 	return []lint.Failure{lint.NewInternalFailure(e.Error())}
 }
+
+// hasUpperCaseLetter checks if string contains at least one upper case letter
+func hasUpperCaseLetter(s string) bool {
+	for _, r := range s {
+		if unicode.IsUpper(r) {
```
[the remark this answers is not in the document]
ccoVeille: Please note, it's not totally his fault

- https://github.com/mgechev/revive/pull/1322#pullrequestreview-2779884804

😅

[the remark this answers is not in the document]
mfederowicz: @ccoVeille  no problem, added to isASCIIUpper and some non ascii test cases :)

rule/utils.go
```diff
@@ -140,3 +141,76 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+	if len(s) == 0 {
+		return false
+	}
+	r := []rune(s)
+	c := r[0]
+	if len(r) == 1 {
+		return unicode.IsUpper(c)
+	}
+	switch {
+	case c == '_', unicode.IsUpper(c):
```
ccoVeille, line 171: Should it be this


```suggestion
		return isASCIIUpper(c)
	}
	switch {
	case c == '_', isASCIIUpper(c):
```

rule/utils.go
```diff
@@ -140,3 +141,76 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+		default:
+			return false
+		}
+	}
+	return true
+}
+
+// isUpperOrDigit checks if rune is upper case letter or digit
+func isUpperOrDigit(c rune) bool {
+	return unicode.IsUpper(c) || unicode.IsDigit(c)
```
ccoVeille, line 195:
```suggestion
	return isASCIIUpper(c) || unicode.IsDigit(c)
```

ccoVeille:
@mfederowicz if you agree, I would appreciate a Co-Authored-By tag.

ccoVeille:
> > maybe using time for measuring execution time is not the best
> 
> [hyperfine](https://github.com/sharkdp/hyperfine) might be the right tool for this kind of benchmarks.

Also, did you have a chance to use tools like hyperfine like what @chavacava suggested here?

mfederowicz:
@ccoVeille yes what we want to measure (what scenario)

maybe 

run ```hyperfine revive --formatter friendly ./...``` on https://github.com/kubernetes/kubernetes with flag: upperCaseConst=true and false

with which versions?

ccoVeille:
I would compare the code from this branch ( 2fa9077146104a1731749994cc9a2afacea1b507 ) and the code from master ( 17527c928baeac86c96a6b87909ef9a5539dd8e3 )

Also, I would limit the rules to var-naming only via a bench.toml config file, then use -- config=bench.toml

mfederowicz:
@ccoVeille iam confused, below results of tests executed on master: https://github.com/kubernetes/kubernetes

branch/revive - v1.9.0-12-g2fa9077
master/revive - v1.9.0-2-g17527c9

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| `branch/revive --config=bench.false.toml --formatter friendly ./...` | 2.376 ± 0.064 | 2.286 | 2.477 | 1.00 |
| `branch/revive --config=bench.true.toml --formatter friendly ./...` | 2.355 ± 0.058 | 2.248 | 2.443 | 1.00 |
| `master/revive --config=bench.false.toml --formatter friendly ./...` | 2.387 ± 0.042 | 2.328 | 2.470 | 1.00 |
| `master/revive --config=bench.true.toml --formatter friendly ./...` | 2.329 ± 0.069 | 2.206 | 2.414 | 1.00 |

rule/utils.go
```diff
@@ -140,3 +141,76 @@ func isCallToExitFunction(pkgName, functionName string) bool {
 func newInternalFailureError(e error) []lint.Failure {
 	return []lint.Failure{lint.NewInternalFailure(e.Error())}
 }
+
+// isASCIIUpper checks if rune is ascii upper case letter
+func isASCIIUpper(r rune) bool {
```
alexandear, line 146: Can be simply:
```suggestion
func isUpper(r rune) bool {
```

rule/utils.go
```diff
@@ -140,3 +141,76 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+			return false
+		}
+	}
+	return true
+}
+
+// isUpperOrDigit checks if rune is upper case letter or digit
+func isUpperOrDigit(c rune) bool {
+	return isASCIIUpper(c) || unicode.IsDigit(c)
+}
```
alexandear, line 196: I propose to create `isDigit` function instead of `unicode.IsDigit` because `unicode.IsDigit('৩') == true`.
```suggestion
func isUpperOrDigit(r rune) bool {
	return isUpper(r) || isDigit(c)
}
```

Also, let's use `r` for consistency with `isUpper`.

alexandear (approved):

ccoVeille (approved):

mfederowicz:
ok @alexandear, @ccoVeille, @chavacava  i resolved conflict from master, so we can merge this PR :P

ccoVeille:
> @mfederowicz if you agree, I would appreciate a Co-Authored-By tag.
> 

Could you ?

```
Co-authored-by: ccoVeille <3875889+ccoVeille@users.noreply.github.com>
```

rule/utils.go
```diff
@@ -140,3 +140,81 @@ func isCallToExitFunction(pkgName, functionName string) bool {
 func newInternalFailureError(e error) []lint.Failure {
 	return []lint.Failure{lint.NewInternalFailure(e.Error())}
 }
+
+// isUpper checks if rune is digit
+func isDigit(r rune) bool {
+	return r >= '0' && r <= '9'
```
denisvmedia, line 146: Technically in go we can name the symbols in unicode. Should we use unicode.IsDigit instead? If we deliberately don't want to use the unicode package, I'd then put a comment on the reasoning.

rule/utils.go
```diff
@@ -140,3 +140,81 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+}
+
+// isUpper checks if rune is ascii upper case letter
+func isUpper(r rune) bool {
+	return r >= 'A' && r <= 'Z'
```
denisvmedia, line 151: Ditto

rule/utils.go
```diff
@@ -140,3 +140,81 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+
+// isUpperOrDigit checks if rune is upper case letter or digit
+func isUpperOrDigit(r rune) bool {
+	return isUpper(r) || isDigit(r)
+}
+
+// isUpperUnderScore detects variable that are made from upper case letters, underscore, or digits.
+//
+// short variable names are considered OK
+func isUpperUnderScore(s string) bool {
```
denisvmedia, line 205: Nit:

```suggestion
func isUpperUnderscore(s string) bool {
```

rule/utils.go
ccoVeille: Please read resolved comments

Everything was almost about this remark

Such ad 
https://github.com/mgechev/revive/pull/1322#discussion_r2057855880

So yes, adding a comment and a unit test about a "unicode digit"

rule/utils.go
ccoVeille: Same here, it's normal to do not use unicode.IsUpper

See previous discussion.

ccoVeille:
@denisvmedia if you feel there is a need for allowing "extended Unicode character" in the same way Go allows us to use via the spec.

You can take a look at #1315

-#1315

rule/utils_test.go
```diff
@@ -30,3 +31,195 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+		_ = allUpperCaseRE.MatchString(input)
+	}
+}
+
+func BenchmarkAllCaps(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		input := "_SOME_PRIVATE_CONST_2"
+		_ = hasUpperCaseLetter(input)
+	}
+}
```
ccoVeille, line 93: I would remove all benchmark before merging

rule/utils_test.go
```diff
@@ -30,3 +31,195 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+		_ = upperCaseConstRE.MatchString(input)
+	}
+}
+
+func BenchmarkIsUpperCaseConst(b *testing.B) {
+	for i := 0; i < b.N; i++ {
+		input := "A1_B2_C3"
+		_ = isUpperCaseConst(input)
+	}
+}
```
ccoVeille, line 141: Same

mfederowicz:
ok @ccoVeille I added you as Co-authored in last commit, as some one said: last but not least :)

rule/utils_test.go
```diff
@@ -30,3 +31,195 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+	}{
+		{'0', false},
+		{'5', false},
+		{'a', false},
+		{'A', true},
+		{'Ą', false},
+		{'Ć', false},
+		{'B', true},
+		{'C', true},
+		{'Z', true},
```
ccoVeille, line 213: Please add CJK rune here

rule/utils_test.go
```diff
@@ -30,3 +31,195 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+		{'0', true},
+		{'1', true},
+		{'2', true},
+		{'3', true},
+		{'4', true},
+		{'5', true},
+		{'6', true},
+		{'7', true},
+		{'8', true},
+		{'9', true},
```
ccoVeille, line 184: Please add a unicode digit here

You can find one in the documentation of unicode.IsDigit

rule/utils.go
```diff
@@ -140,3 +140,81 @@ func isCallToExitFunction(pkgName, functionName string) bool {
 func newInternalFailureError(e error) []lint.Failure {
 	return []lint.Failure{lint.NewInternalFailure(e.Error())}
 }
+
+// isUpper checks if rune is digit
+func isDigit(r rune) bool {
+	return r >= '0' && r <= '9'
+}
+
+// isUpper checks if rune is ascii upper case letter
```
ccoVeille, line 149:
```suggestion
// isUpper checks if rune is ASCII upper case letter
//
// # Note
// We restrict to A-Z because [Unicode.IsUpper] return true for a large variety of letter
```

rule/utils.go
ccoVeille, line 146:
```suggestion
// isUpper checks if rune is simple digit
//
// # Note
// we don't use [unicode.IsDigit] as it returns true for a large variety of digit that are not 0-9
func isDigit(r rune) bool {
	return r >= '0' && r <= '9'
```

ccoVeille (commented):
Maybe this

rule/utils_test.go
mfederowicz: @ccoVeille did you think about to split TestIsUpperFunction into three tests:

- non ascii (for example digits)
- unicode uppercase
- unicode lowercase/symbols

rule/utils.go
ccoVeille: @denisvmedia the comments were added by @mfederowicz

rule/utils_test.go
mfederowicz: @ccoVeille what do you mean: "it would avoid the true/false also"? should we dont use test cases with input and expected value? can you show me example?

ccoVeille: Here is the current code

```go
func TestIsUpperFunction(t *testing.T) { 
         tests := []struct { 
                 input    rune 
                 expected bool 
         }{ 
                 {'0', false}, 
                 {'5', false}, 
                 {'a', false}, 
                 {'A', true}, 
                 {'Ą', false}, 
                 {'Ć', false}, 
                 {'B', true}, 
                 {'C', true}, 
                 {'Z', true}, 
                 {' ', false}, 
                 {'!', false}, 
                 {'🙂', false}, // Emoji to test unicode 
                 {'你', false}, 
                 {'日', false}, 
                 {'本', false}, 
                 {'語', false}, 
                 {'韓', false}, 
                 {'中', false}, 
                 {'文', false}, 
                 {'あ', false}, 
                 {'ア', false}, 
                 {'한', false}, 
         } 
  
         for _, tt := range tests { 
                 result := isUpper(tt.input) 
                 if result != tt.expected { 
                         t.Errorf("isUpper(%q) = %v; want %v", tt.input, result, tt.expected) 
                 } 
         } 
 }
```

Here is what I suggest

```go
func TestIsUpperFunction(t *testing.T) { 
         tests := []struct { 
                 input    rune 
                 expected bool 
         }{ 
                 {'0', false}, 
                 {'5', false}, 
                 {'a', false}, 
                 {'A', true}, 
                 {'Ą', false}, 
                 {'Ć', false}, 
                 {'B', true}, 
                 {'C', true}, 
                 {'Z', true}, 
                 {' ', false}, 
                 {'!', false}, 
                 {'🙂', false}, // Emoji to test unicode 
                 {'你', false}, 
                 {'日', false}, 
                 {'本', false}, 
                 {'語', false}, 
                 {'韓', false}, 
                 {'中', false}, 
                 {'文', false}, 
                 {'あ', false}, 
                 {'ア', false}, 
                 {'한', false}, 
         } 
  
         for _, tt := range tests { 
                 result := isUpper(tt.input) 
                 if result != tt.expected { 
                         t.Errorf("isUpper(%q) = %v; want %v", tt.input, result, tt.expected) 
                 } 
         } 
 }
```

rule/utils_test.go
```diff
@@ -30,3 +30,171 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+		{'한', false},
+	}
+
+	for _, tt := range tests {
+		result := isUpper(tt.input)
+		if result != tt.expected {
+			t.Errorf("isUpper(%q) = %v; want %v", tt.input, result, tt.expected)
+		}
+	}
+}
```
ccoVeille, line 200: Here is what I suggest

```suggestion
func TestIsUpperFunction(t *testing.T) {
	t.Run("non letter", func(t *testing.T) {
	tests := []rune
		'0',
		'5'
		' ',
		'_',
		'!',
		'🙂', // Emoji to test unicode
	}

	for _, r := range tests {
		result := isUpper(r)
		if result {
			t.Errorf("isUpper(%q) = %v; want false", r, result)
		}
	}
})

t.Run("non ASCII letter", func(t *testing.T) {
	tests := []rune
		'Ą',
		'Ć',
		'你',
		'日',
		'本',
		'語',
		'韓',
		'中',
		'文',
		'あ',
		'ア',
		'한',
	}

	for _, r := range tests {
		result := isUpper(r)
		if result {
			t.Errorf("isUpper(%q) = %v; want false", r, result)
		}
	}
})

t.Run("lowercase ASCII letter", func(t *testing.T) {
	tests := []rune
		'a',
		'b',
	}

	for _, r := range tests {
		result := isUpper(r)
		if result {
			t.Errorf("isUpper(%q) = %v; want false", r, result)
		}
	}
})

t.Run("uppercase ASCII letter", func(t *testing.T) {
	tests := []rune
		'A',
		'B',
		'C', 
		'Z', 
	}

	for _, r := range tests {
		result := isUpper(r)
		if !result {
			t.Errorf("isUpper(%q) = %v; want true", r, result)
		}
	}
})
}
```

rule/utils_test.go
ccoVeille: Please take a look at

https://github.com/mgechev/revive/pull/1322#discussion_r2059266875

mfederowicz:
@ccoVeille interesting because in my mailbox i have email with suggested by you changes, but suggested change is the same code (or maybe it is time to sleep :))

ccoVeille:
> @ccoVeille interesting because in my mailbox i have email with suggested by you changes, but suggested change is the same code (or maybe it is time to sleep :))

I simply send too fast, so I deleted,and now I  managed to provide what I initially meant

rule/utils.go
```diff
@@ -140,3 +140,87 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+}
+
+// isUpperOrDigit checks if rune is upper case letter or digit
+func isUpperOrDigit(r rune) bool {
+	return isUpper(r) || isDigit(r)
+}
+
+// isUpperUnderscore detects variable that are made from upper case letters, underscore, or digits.
+//
+// short variable names are considered OK
```
alexandear, line 210:
```suggestion
// Short variable names are considered OK
```

rule/utils_test.go
```diff
@@ -30,3 +30,204 @@ func TestIsCallToExitFunction(t *testing.T) {
 		})
 	}
 }
+
+func TestHasUpperCaseFunction(t *testing.T) {
```
alexandear, line 34:
```suggestion
func TestHasUpperCaseLetter(t *testing.T) {
```

rule/utils_test.go
```diff
@@ -30,3 +30,204 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.varName, func(t *testing.T) {
+			if got := hasUpperCaseLetter(tt.varName); got != tt.expected {
+				t.Errorf("hasCaps(%s) = %v; want %v", tt.varName, got, tt.expected)
+			}
+		})
+	}
+}
+
+func TestIsUpperConstFunction(t *testing.T) {
```
alexandear, line 64:
```suggestion
func TestIsUpperConst(t *testing.T) {
```

rule/utils_test.go
```diff
@@ -30,3 +30,204 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+	for _, tt := range tests {
+		t.Run(tt.varName, func(t *testing.T) {
+			if got := isUpperCaseConst(tt.varName); got != tt.expected {
+				t.Errorf("isUpperCaseConst(%s) = %v; want %v", tt.varName, got, tt.expected)
+			}
+		})
+	}
+}
+
+func TestIsUpperUnderScoreFunction(t *testing.T) {
```
alexandear, line 97:
```suggestion
func TestIsUpperUnderScore(t *testing.T) {
```

rule/utils_test.go
```diff
@@ -30,3 +30,204 @@ func TestIsCallToExitFunction(t *testing.T) {
...
+
+	for _, tt := range tests {
+		result := isDigit(tt.input)
+		if result != tt.expected {
+			t.Errorf("isDigit(%q) = %v; want %v", tt.input, result, tt.expected)
+		}
+	}
+}
+
+func TestIsUpperFunction(t *testing.T) {
```
alexandear, line 165:
```suggestion
func TestIsUpper(t *testing.T) {
```

rule/utils.go
```diff
@@ -140,3 +140,87 @@ func isCallToExitFunction(pkgName, functionName string) bool {
 func newInternalFailureError(e error) []lint.Failure {
 	return []lint.Failure{lint.NewInternalFailure(e.Error())}
 }
+
+// isUpper checks if rune is simple digit
```
alexandear, line 144:
```suggestion
// isUpper checks if rune is a simple digit.
```

rule/utils.go
```diff
@@ -140,3 +140,87 @@ func isCallToExitFunction(pkgName, functionName string) bool {
...
+//
+// # Note
+// we don't use [unicode.IsDigit] as it returns true for a large variety of digit that are not 0-9
```
alexandear, line 147: pkgsite hides unexported docs, so no need to use `[]`. Also, propose some rephrasing:

```suggestion
// The function doesn't use unicode.IsDigit as it returns true for a large variety of digits that are not 0-9.
```

ccoVeille:
> > We spent too much time reviewing the PR for a tiny performance benefit.

I agree.

> IMO, there is no actual performance benefit and, besides the time invested on reviews and coding, now the codebase is bigger and more complex.

I agree too.

But I feel like it's still better as it helped to figures things for #1315

- #1315

Thanks for your work @mfederowicz 

And thank everyone ( @chavacava @alexandear @denisvmedia ) for the time spent on this PR.

PR #1346, opened 2025-05-14
Integrate golangci-lint into CI/CD pipeline

### LINKED ISSUE #1250: Integrate golangci-lint into CI/CD pipeline
I propose to integrate [golangci-lint](https://golangci-lint.run/) into the CI/CD pipeline to ensure code quality and consistency across the project.

## Steps

1. Add golangci-lint config with minimal configuration. The config could be based on https://github.com/ccoVeille/golangci-lint-config-examples
2. Extend GitHub Actions Workflow `.github/workflows/lint.yaml` with `golangci-lint` job. It should use [`golangci-lint-action`](https://github.com/marketplace/actions/golangci-lint).
3. Run golangci-lint locally, identify, and fix existing linting issues.

ArjunRajpal123:
## Motivation/Why
Integrate [golangci-lint](https://golangci-lint.run/) into the CI/CD pipeline to ensure code quality and consistency across the project.

## Steps followed
1. Add golangci-lint config with minimal configuration. The config could be based on https://github.com/ccoVeille/golangci-lint-config-examples
2. Extend GitHub Actions Workflow .github/workflows/lint.yaml with golangci-lint job. It should use [golangci-lint-action](https://github.com/marketplace/actions/golangci-lint).
3. Run golangci-lint locally, identify, and fix existing linting issues.
4. Disabled Revive in the golangci-lint config

**Important Notes**
- I initially attempted to follow the [basic golangci-lint config](https://github.com/ccoVeille/golangci-lint-config-examples/tree/main/02-basic), but quickly ran into 34 linting issues (errcheck: 24, staticcheck: 10). To keep this PR focused and manageable, I decided to enable only staticcheck and address all related issues.

- I’ve disabled rule SA1019 (usage of deprecated identifiers) for now, as the only trigger was ast.Object, which is a known case of a deprecated field. Handling this seems better suited for a dedicated follow-up PR.

# Footer
Closes #1250

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -45,9 +45,6 @@ jobs:
       with:
         go-version: stable
 
-    - name: Install revive
-      run: go install
-
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v8
       with:
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -48,9 +48,6 @@ jobs:
     - name: Install revive
       run: go install
 
-    - name: Run revive
-      run: revive --config revive.toml ./...
-
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v8
       with:
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -45,7 +45,13 @@ jobs:
       with:
         go-version: stable
 
-    - name: golangci-lint
+    - name: Install revive
+      run: go install
+
+    - name: Run revive
+      run: revive --config revive.toml ./...
+
+    - name: Run golangci-lint
       uses: golangci/golangci-lint-action@v8
       with:
         version: v2.1.6
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -48,4 +48,4 @@ jobs:
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v8
       with:
-        version: latest
+        version: v2.1.6
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -50,3 +50,8 @@ jobs:
 
     - name: Run revive
       run: revive --config revive.toml ./...
+
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.1.6
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -48,4 +48,4 @@ jobs:
     - name: golangci-lint
       uses: golangci/golangci-lint-action@v8
       with:
-        version: v2.1.6
+        version: latest
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -7,3 +7,30 @@
 # Version: v2.0.0
 #
 version: "2"
+
+linters:
+  # some linters are enabled by default
+  # https://golangci-lint.run/usage/linters/
+  #
+  enable:
+    # Vet examines Go source code and reports suspicious constructs.
+    - govet
+
+    # Detects when assignments to existing variables are not used.
+    - ineffassign
+
+    # It's a set of rules from staticcheck. See https://staticcheck.io/
+    - staticcheck
+
+    # Checks Go code for unused constants, variables, functions and types.
+    - unused
+
+  # TODO: Phase in the following
+  disable:
+    # Errcheck is a program for checking for unchecked errors in Go code.
+    - errcheck
+
+  exclusions:
+    rules:
+      - linters: staticcheck
+        text: "SA1019:"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,9 @@
+---
+# golangci-lint configuration file made by @ccoVeille
+# Source: https://github.com/ccoVeille/golangci-lint-config-examples/
+# Author: @ccoVeille
+# License: MIT
+# Variant: 01-defaults
+# Version: v2.0.0
+#
+version: "2"
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,16 +24,62 @@ linters:
 
     # Checks Go code for unused constants, variables, functions and types.
     - unused
+
+    - revive
  
   disable:
     # TODO: enable errcheck and resolve all errors
     # Errcheck is a program for checking for unchecked errors in Go code.
     - errcheck
 
-    - revive
+  settings:
+    # the following settings should match revive.toml
+    revive:
+      severity: warning
+      confidence: 0.8
 
-  exclusions:
-    rules:
-      - linters: [staticcheck]
-        # SA1019: Using a deprecated function, variable, constant or field
-        text: "SA1019:"
+      rules:
+        - name: bare-return
+        - name: blank-imports
+        - name: comment-spacings
+        - name: constant-logical-expr
+        - name: context-as-argument
+        - name: context-keys-type
+        - name: deep-exit
+        - name: dot-imports
+        - name: empty-block
+        - name: empty-lines
+        - name: enforce-map-style
+          arguments: ["literal"]
+        - name: enforce-slice-style
+          arguments: ["literal"]
+        - name: error-naming
+        - name: error-return
+        - name: error-strings
+        - name: errorf
+        - name: exported
+        - name: filename-format
+          arguments: ["^[_a-z][_a-z0-9]*\\.go$"]
+        - name: identical-branches
+        - name: increment-decrement
+        - name: indent-error-flow
+        - name: line-length-limit
+          arguments: [200]
+        - name: package-comments
+        - name: range
+        - name: receiver-naming
+        - name: redefines-builtin-id
+        - name: redundant-build-tag
+        - name: superfluous-else
+        - name: time-naming
+        - name: unexported-naming
+        - name: unexported-return
+        - name: unnecessary-stmt
+        - name: unreachable-code
+        - name: unused-parameter
+        - name: unused-receiver
+        - name: useless-break
+        - name: use-any
+        - name: var-declaration
+        - name: var-naming
+
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,6 +26,8 @@ linters:
     - unused
 
     - revive
+
+    - nolintlint
  
   disable:
     # TODO: enable errcheck and resolve all errors
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,45 +1,33 @@
----
-# golangci-lint configuration file made by @ccoVeille
-# Source: https://github.com/ccoVeille/golangci-lint-config-examples/
-# Author: @ccoVeille
-# License: MIT
-# Variant: 01-defaults
-# Version: v2.0.0
-#
+# This configuration for golangci-lint.
+# See https://golangci-lint.run/usage/configuration/ for details.
+
 version: "2"
 
 linters:
-  # some linters are enabled by default
-  # https://golangci-lint.run/usage/linters/
-  #
+  default: none
   enable:
-    # Vet examines Go source code and reports suspicious constructs.
     - govet
-
-    # Detects when assignments to existing variables are not used.
     - ineffassign
-
-    # It's a set of rules from staticcheck. See https://staticcheck.io/
+    - misspell
+    - nolintlint
+    - revive
     - staticcheck
-
-    # Checks Go code for unused constants, variables, functions and types.
     - unused
 
-    - revive
-
-    - nolintlint
- 
-  disable:
-    # TODO: enable errcheck and resolve all errors
-    # Errcheck is a program for checking for unchecked errors in Go code.
-    - errcheck
-
   settings:
-    # the following settings should match revive.toml
+    govet:
+      enable-all: true
+      disable:
+        - fieldalignment
+        - shadow
+    nolintlint:
+      allow-unused: true
+      require-explanation: true
+      require-specific: true
+    # The following settings should match revive.toml
     revive:
       severity: warning
       confidence: 0.8
-
       rules:
         - name: bare-return
         - name: blank-imports
@@ -61,6 +49,7 @@ linters:
         - name: errorf
         - name: exported
         - name: filename-format
+          # Override the default pattern to forbid .go files with uppercase letters and dashes.
           arguments: ["^[_a-z][_a-z0-9]*\\.go$"]
         - name: identical-branches
         - name: increment-decrement
@@ -84,4 +73,3 @@ linters:
         - name: use-any
         - name: var-declaration
         - name: var-naming
-
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -24,13 +24,16 @@ linters:
 
     # Checks Go code for unused constants, variables, functions and types.
     - unused
-
-  # TODO: Phase in the following
+ 
   disable:
+    # TODO: enable errcheck and resolve all errors
     # Errcheck is a program for checking for unchecked errors in Go code.
     - errcheck
+
+    - revive
 
   exclusions:
     rules:
       - linters: staticcheck
+        # SA1019: Using a deprecated function, variable, constant or field
         text: "SA1019:"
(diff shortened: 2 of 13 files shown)


chavacava:
Hi @ArjunRajpal123 thanks for the PR!

IMO we should not disable `revive` but run it with the configuration we used until now (`revive.toml` at the root of the repository) (what do you think @denisvmedia  @alexandear @ccoVeille ?)

.golangci.yml
```diff
@@ -0,0 +1,39 @@
...
+    # Errcheck is a program for checking for unchecked errors in Go code.
+    - errcheck
+
+    - revive
+
+  exclusions:
+    rules:
+      - linters: [staticcheck]
+        # SA1019: Using a deprecated function, variable, constant or field
+        text: "SA1019:"
```
ccoVeille, line 39: So here is some context

> I’ve disabled rule SA1019 (usage of deprecated identifiers) for now, as the only trigger was ast.Object, which is a known case of a deprecated field. Handling this seems better suited for a dedicated follow-up PR.

There multiple ways to handle this:
- the way you chose
- the fact to disable SA1019 from `linters.settings.staticcheck.check` by using -SA1019
- use //nolint:staticcheck at each place ast.Object is used.

I would have used the last option, and add a TODO if there is something we can do.

ccoVeille:
> IMO we should not disable revive but run it with the configuration we used until now

I would say it depends whether we plan to remove revive from the CI.

Here is an option I would suggest if we want to keep revive in the CI:

- define everything we need to have revive with same settings as we had
- but configure golangci-lint-action to disable revive
- this way IDE would use golangci-lint settings.

chavacava:
> > IMO we should not disable revive but run it with the configuration we used until now
>  I would say it depends whether we plan to remove revive from the CI.

When I say "we should not disable revive" I mean running it through `golangci-lint`

.golangci.yml
ArjunRajpal123: `//nolint:staticcheck` I like this solution a lot more

ccoVeille:
> > > IMO we should not disable revive but run it with the configuration we used until now
> >  I would say it depends whether we plan to remove revive from the CI.
> 
> When I say "we should not disable revive" I mean running it through `golangci-lint` 

I wasn't sure, so @ArjunRajpal123 here is what you could do:

enable revive linter via the .golangci.yaml file

Then add a node for linters.settings.revive and adapt the setting you will find in 

https://github.com/mgechev/revive/blob/master/revive.toml

denisvmedia:
I'm wondering if it's the same when we use revive directly and revive via golangci-lint. As far as I understand currently we check with the most recent version of revive, while in golangci-lint we won't have the most recent of revive.

ArjunRajpal123:
Some recent Changes to this pr
- [x] used `//nolint:staticcheck` at each place `ast.Object` is used (also verified that each place where apt.Object is used has a todo)
- [x]  enabled revive linter via the .golangci.yaml file
- [x] added a node for linters.settings.revive in [this file](https://github.com/mgechev/revive/blob/master/revive.toml) and adapt the setting you will find in
- [x] removed revive from GHA workflow


Sidenote: found two revive lint errors after enabling revive in `golangci`
image

rule/error_strings_test.go
```diff
@@ -34,7 +34,8 @@ func TestErrorStringsRule_Configure(t *testing.T) {
 		{
 			name:      "Invalid function",
 			arguments: lint.Arguments{"errors."},
-			wantErr:   errors.New("found invalid custom function: errors."),
+			//nolint: revive
```
ccoVeille, line 37: It should be this

```suggestion
			//nolint:revive
```

Please enable `nolintlint` linter via `.golangci.yaml` 

It would lint, the `//nolint` directives

formatter/formatter_test.go
```diff
@@ -61,11 +61,15 @@ Warnings:
...
+				`"Category":"cat","Position":{"Start":{"Filename":"test.go","Offset":0,"Line":2,"Column":5},` +
+				`"End":{"Filename":"test.go","Offset":0,"Line":2,"Column":10}},"Confidence":0,"ReplacementLine":""}]`,
 		},
 		{
 			formatter: &formatter.NDJSON{},
-			want:      `{"Severity":"warning","Failure":"test failure","RuleName":"rule","Category":"cat","Position":{"Start":{"Filename":"test.go","Offset":0,"Line":2,"Column":5},"End":{"Filename":"test.go","Offset":0,"Line":2,"Column":10}},"Confidence":0,"ReplacementLine":""}`,
+			want: `{"Severity":"warning","Failure":"test failure","RuleName":"rule",` +
+				`"Category":"cat","Position":{"Start":{"Filename":"test.go","Offset":0,"Line":2,"Column":5},` +
+				`"End":{"Filename":"test.go","Offset":0,"Line":2,"Column":10}},"Confidence":0,"ReplacementLine":""}`,
 		},
```
ccoVeille, line 73: Is this change related to a the line-length rule set to 200?


if it's right, then it means calling revive didn't detect it, while enable the rulevia golangci-lint does  🤔 

Maybe a typo in the revive. toml file?

@ArjunRajpal123 I'm fine with the changes, it improves readability, but I'm unsure why you had to do it.

rule/error_strings_test.go
```diff
@@ -34,7 +34,8 @@ func TestErrorStringsRule_Configure(t *testing.T) {
...
+			wantErr: errors.New("found invalid custom function: errors."),
```
ccoVeille, line 38: Here also, I'm unsure why you had to disable revive rule.

revive was supposed to be enabled on the code

formatter/formatter_test.go
ArjunRajpal123: It is related to the line-length rule being set to 200 and I was a bit confused myself as to why it was showing up myself. I am going to look into this more closely.

.golangci.yml
```diff
@@ -0,0 +1,85 @@
...
+    # Detects when assignments to existing variables are not used.
+    - ineffassign
+
+    # It's a set of rules from staticcheck. See https://staticcheck.io/
+    - staticcheck
+
+    # Checks Go code for unused constants, variables, functions and types.
+    - unused
+
+    - revive
```
ccoVeille, line 28: Please enable `nolintlint` linter, as I recommended here

https://github.com/mgechev/revive/pull/1346#discussion_r2090225629

.github/workflows/lint.yaml
```diff
@@ -48,5 +48,7 @@ jobs:
     - name: Install revive
```
denisvmedia, line 48: Note, you've removed revive run, but didn't remove the installation step. I put a comment [here](https://github.com/mgechev/revive/pull/1346#issuecomment-2881567252), but it might have got lost. I believe running golangci-lint with the latest bundled revive is not the same as running currently built revive.

.github/workflows/lint.yaml
```diff
@@ -48,5 +48,7 @@ jobs:
...
       run: go install
 
-    - name: Run revive
-      run: revive --config revive.toml ./...
+    - name: golangci-lint
+      uses: golangci/golangci-lint-action@v8
+      with:
+        version: v2.1.6
```
denisvmedia, line 54: I'm not sure about explicit version here. Will it require manual maintenance in the future?

.github/workflows/lint.yaml
ccoVeille: I agree, it's why I asked what was expected here

https://github.com/mgechev/revive/pull/1346#issuecomment-2880000622

.golangci.yml
denisvmedia: Actually, using `//nolint:staticcheck` is not the same. It will disable the whole linter, not just a particular check.

.github/workflows/lint.yaml
ArjunRajpal123: It would require additional manual maintenance, I can set version to `latest` to avoid this
https://github.com/golangci/golangci-lint-action?tab=readme-ov-file#version

.golangci.yml
ArjunRajpal123: I have now enabled this linter

.github/workflows/lint.yaml
ccoVeille: That's a usual way to do.

The idea behind this is to avoid someone opening a small change one thing, won't have to fight with all the change the last version of golangci-lint would report.

The very same way to setup-go action is often set up to use go-1. 24, and not stable, oldstable.

But yes, it requires manual maintenance

rule/error_strings_test.go
ArjunRajpal123: image

I ran into the following error above when using golangci+revive, which appears related to the following code linked below. Modifying the `wantErr` wouldn't be the correct fix because that causes tests to fail and modifying the rule below seems out of scope. Also curious as to why this shows up now.

https://github.com/mgechev/revive/blob/f6cc22de4d5c75cf471501affa5522c778078761/rule/error_strings.go#L54-L56

formatter/formatter_test.go
ccoVeille: It might be a bug in the way revive reads the line length from the config file.

I have locally set up the max line length to 5, and go run . reported nothing 😱

Maybe the config in the current revive.toml is wrong, and is not what the code expects.

Or the opposite, there is small bug in this rule.

Anyway, this should be addressed in another PR.

.golangci.yml
ccoVeille: Yes, but here it would be disabled for one line only. Instead of disabling SA1019 everywhere.

Staticcheck has some directives.

https://staticcheck.dev/docs/configuration/#line-based-linter-directives

Maybe it works when staticcheck is used through golangci-lint

rule/error_strings_test.go
ccoVeille: Here I agree the errors strings shouldn't end with a dot.

But it's a special case, it's the unit tests of errors-strings rule 😬😅

So here we might think about how to avoid adding directives that would allow usto write bad code because we want to test it 😁

.github/workflows/lint.yaml
denisvmedia: > But yes, it requires manual maintenance

I bet, this will be rarely updated then.

rule/error_strings_test.go
ccoVeille: > related to the following code linked below. Modifying the `wantErr` wouldn't be the correct fix because that causes tests to fail and modifying the rule below seems out of scope (disabling felt like the right choice). Also curious as to why this lint-error shows up now.

But you might also round a real bug.

.golangci.yml
denisvmedia: Unfortunately, it doesn't wok with golangci-lint. That's a well-known issue. And yes, disabling the whole linter for one line is better than the rule for the whole project. I agree.

.github/workflows/lint.yaml
chavacava: IMO a manual update is not a bad thing, by the contrary (we keep the control on the actual version we apply to our code base)

.golangci.yml
```diff
@@ -0,0 +1,87 @@
+---
+# golangci-lint configuration file made by @ccoVeille
+# Source: https://github.com/ccoVeille/golangci-lint-config-examples/
+# Author: @ccoVeille
+# License: MIT
+# Variant: 01-defaults
+# Version: v2.0.0
+#
```
alexandear, line 8: @ccoVeille do we need this copyright? During a review process we have changed a lot, so this is not the `01-defaults` configuration from https://github.com/ccoVeille/golangci-lint-config-examples/

.golangci.yml
```diff
@@ -0,0 +1,87 @@
...
+version: "2"
+
+linters:
```
alexandear, line 11: A few suggestions from my side as one of the [most active golangci-lint contributors](https://github.com/golangci/golangci-lint/graphs/contributors):

* I'd recommend adding `default: none` to ensure that only explicitly enabled linters are used. See https://github.com/golangci/golangci-lint/blob/deee4e75cd912954538449a2f35d795ea04545ce/.golangci.reference.yml#L15. That way, when a new version of golangci-lint is released, we won't implicitly enable new linters.
* It might be beneficial to enable as many linters as possible from [this list](https://github.com/golangci/golangci-lint/blob/deee4e75cd912954538449a2f35d795ea04545ce/.golangci.reference.yml#L19-L123).
* For easier future maintenance, I suggest sorting the list of enabled linters.
* The `disable` section can be removed, as it's redundant when using `default: none`.

rule/datarace.go
```diff
@@ -23,6 +23,7 @@ func (r *DataRaceRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 		funcResults := funcDecl.Type.Results
 
 		// TODO: ast.Object is deprecated
+		//nolint:staticcheck
```
alexandear, line 26: Maybe better:
```suggestion
		//nolint:staticcheck // TODO: ast.Object is deprecated
```

alexandear: The same below

rule/error_strings_test.go
```diff
@@ -34,7 +34,8 @@ func TestErrorStringsRule_Configure(t *testing.T) {
 		{
 			name:      "Invalid function",
 			arguments: lint.Arguments{"errors."},
-			wantErr:   errors.New("found invalid custom function: errors."),
+			//nolint:revive
```
alexandear, line 37: I think we should adding an explanation we used `//nolint:revive` here.

rule/if_return.go
```diff
@@ -44,7 +44,7 @@ func (w *lintElseError) Visit(node ast.Node) ast.Visitor {
 				continue
 			}
 			assign, ok := s.Init.(*ast.AssignStmt)
-			if !ok || len(assign.Lhs) != 1 || !(assign.Tok == token.DEFINE || assign.Tok == token.ASSIGN) {
+			if !ok || len(assign.Lhs) != 1 || assign.Tok != token.DEFINE && assign.Tok != token.ASSIGN {
```
alexandear, line 47: I'm not sure if this became more readable. Let's revert.

.golangci.yml
ccoVeille: Absolutely not. I added this **to my repository** because:
- I wanted to define the license of my work is MIT
- I have the same headers for all my examples
- some people are copy-pasting my "work" as-is.
- I can track the projects that are using my examples, because I'm curious
- sometimes I contributed to some projects my example files were added, for example I helped a project to migrate to the v2 format.

But then here in the context of this PR:
- this file is the very basic one, so of course it has to be complicated.
- the file diverted and is the results of contributions and feedbacks from many people.
- I chose the MIT license because it permits to be included in any project, no matter its license, and it allows to alter the file in the way, and it can be updated in a way to remove my name.


TL; DR; yes, of course, the whole header can be removed.

The only thing I'm unsure about is if the file doesn't have to keep the MIT license line.

But as revive is licensed under MIT license,
it can be removed.

rule/if_return.go
ccoVeille: I agree

This one is implied by QF0001

Please take a look here, on how to disable the rule that implied this.

https://github.com/gofr-dev/gofr/blob/dac6bdc735c3b60ccfb17d8e4a799c23e3d7383d/.golangci.yml#L141

I had to do it for gofr too

- https://github.com/gofr-dev/gofr/pull/1708

rule/error_strings_test.go
ccoVeille: I agree.

We should enforce using an explanation with nolint 

https://github.com/gofr-dev/gofr/blob/dac6bdc735c3b60ccfb17d8e4a799c23e3d7383d/.golangci.yml#L97-L98

chavacava:
A quick note regarding pull requests from first-time contributors: to promote a welcoming atmosphere, we should aim to streamline and simplify the PR workflow. While I'm deeply convinced that having others review your code is a valuable learning experience, we should be cautious not to overwhelm contributors with excessive comments or lengthy discussions on details that can be addressed in subsequent PRs.

alexandear:
I've addressed all my comments, fixed some additional issues, and merged the PR.

@ArjunRajpal123 — thank you for your contribution!

Thanks everyone for the review.

If you have suggestions for improving the golangci-lint configuration or the CI/CD setup, please open a separate PR. Don't hesitate to contribute!

alexandear (approved):

PR #1354, opened 2025-05-17
test: add missing unit tests for 'unexported-return' rule

semihbkgr:
Add missing unit tests for `unexported-return` rule

I noticed that there were no unit tests for the `unexported-return` rule. I had previously added them in PR #1344, but since that PR was closed, I'm submitting this separate PR specifically for the unit tests.

### DIFF
--- a/test/unexported_return_test.go
+++ b/test/unexported_return_test.go
@@ -1,0 +1,11 @@
+package test
+
+import (
+	"testing"
+
+	"github.com/mgechev/revive/rule"
+)
+
+func TestUnexportedReturn(t *testing.T) {
+	testRule(t, "unexported_return", &rule.UnexportedReturnRule{})
+}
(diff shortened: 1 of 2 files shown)


chavacava:
Hi @semihbkgr thank you for the PR.

`unexported-return` is part of the original set of rules of `golint` and tests are organized a little bit different than tests for other rulers (under `testdata/golint` and triggered by `test/golint_test.go`) 
Could you please check if there is some new test cases in this PR not already covered by https://github.com/mgechev/revive/blob/master/testdata/golint/unexported_return.go ? If that is the case, you could add it.

semihbkgr:
Thank you for the clarification. It's good to know that the original set of rules has its tests organized differently. I initially thought the tests for it were missing, but I now see they're located under `testdata/golint`. I've reviewed the existing test cases, and it looks like all relevant scenarios are already covered.

PR #1355, opened 2025-05-17
tiny refactoring on friendly, sarif and stylish formatters

chavacava:
Replace function call by var ref and move def near utilization in friendly and stylish formatters

### DIFF
--- a/formatter/sarif.go
+++ b/formatter/sarif.go
@@ -27,7 +27,7 @@ func (*Sarif) Format(failures <-chan lint.Failure, cfg lint.Config) (string, err
 	sarifLog := newReviveRunLog(cfg)
 
 	for failure := range failures {
-		sarifLog.AddResult(failure)
+		sarifLog.addResult(failure)
 	}
 
 	buf := new(bytes.Buffer)
@@ -72,7 +72,7 @@ func (l *reviveRunLog) addRules(cfg map[string]lint.RuleConfig) {
 	}
 }
 
-func (l *reviveRunLog) AddResult(failure lint.Failure) {
+func (l *reviveRunLog) addResult(failure lint.Failure) {
 	positiveOrZero := func(x int) int {
 		if x > 0 {
 			return x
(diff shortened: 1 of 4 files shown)


formatter/sarif.go
```diff
@@ -72,7 +72,7 @@ func (l *reviveRunLog) addRules(cfg map[string]lint.RuleConfig) {
 	}
 }
 
-func (l *reviveRunLog) AddResult(failure lint.Failure) {
+func (l *reviveRunLog) addResult(failure lint.Failure) {
```
ccoVeille, line 75: One feedback not directly revlated to this PR, but a linter idea I have by reading this change.

It reminds me unexported-return rule.

Here reviveRunLog is an unexported struct, having unexported methods seems not logic, and I think it's why you made this change

formatter/stylish.go
```diff
@@ -66,6 +62,10 @@ func (*Stylish) Format(failures <-chan lint.Failure, config lint.Config) (string
 		output += table(val) + "\n"
 	}
 
+	ps := "problems"
+	if total == 1 {
+		ps = "problem"
+	}
 	suffix := fmt.Sprintf(" %d %s (%d errors) (%d warnings)", total, ps, totalErrors, total-totalErrors)
```
ccoVeille, line 69: It seems strange you can have 

"1 problem (1 errors) (0 warnings)"
Or

"1 problem (0 errors) (1 warnings)"

You should also handle the singular/plural for "error" and "warning"

yes it's old code unaffected by the PR, but I think you should fix it

ccoVeille (approved):
LGTM 👍

One minor remark ~~, and a linter idea~~

formatter/sarif.go
ccoVeille: I think my idea is stupid if we simply consider a struct like

```go
type fooError struct{}

func (fooError) Error() string {
	return "foo"
}

func whatever() error {
	return fooError{}
}
```

This is valid

alexandear (approved):
LGTM

ccoVeille (approved):

PR #1372, opened 2025-05-23
new rule to spot unnecessary use of formatting functions

chavacava:
The rule spots calls to formatting functions without leveraging formatting directives.

Like:
```golang
log.Printf("UDP server exited") // can be replaced with log.Print
fmt.Fprintf(writer, " }") //  fmt.Fprint
message := fmt.Sprintf("contains an invalid SecurityDefinitionsItem") // fmt.Sprint or just "contains an invalid SecurityDefinitionsItem"
```
The implementation doesn't try to be smart to cope with corner cases, it just checks if the _format_ argument of the functions contains `%`. It also avoids relying on type information and limits itself to check some common patterns (like calling `t` the variable holding `*testing.T` or `logger` the variable holding a standard logger instance)

[In the future someone will ask for adding a configuration to define their own functions to check; and after that, as usual, someone else will ask for adding support for regular expressions.] 

I've tested the rule against k8s repo and it found +280 failures

### DIFF
--- a/rule/unnecessary_format.go
+++ b/rule/unnecessary_format.go
@@ -1,0 +1,92 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"strings"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// UnnecessaryFormatRule spots calls to formatting functions without leveraging formatting directives.
+type UnnecessaryFormatRule struct{}
+
+// Apply applies the rule to given file.
+func (*UnnecessaryFormatRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	fileAst := file.AST
+	walker := lintUnnecessaryFormat{
+		file:    file,
+		fileAst: fileAst,
+		onFailure: func(failure lint.Failure) {
+			failures = append(failures, failure)
+		},
+	}
+
+	ast.Walk(walker, fileAst)
+
+	return failures
+}
+
+// Name returns the rule name.
+func (*UnnecessaryFormatRule) Name() string {
+	return "unnecessary-format"
+}
+
+type lintUnnecessaryFormat struct {
+	file      *lint.File
+	fileAst   *ast.File
+	onFailure func(lint.Failure)
+}
+
+type formattingSpec struct {
+	formatArgPosition byte
+	alternative       string
+}
+
+var formattingFuncs = map[string]formattingSpec{
+	"fmt.Appendf": {1, "fmt.Append"},
+	"fmt.Errorf":  {0, "errors.New"},
+	"fmt.Fprintf": {1, "fmt.Fprint"},
+	"fmt.Fscanf":  {1, "fmt.Fscan or fmt.Fscanln"},
+	"fmt.Printf":  {0, "fmt.Print or fmt.Println"},
+	"fmt.Scanf":   {0, "fmt.Scan"},
+	"fmt.Sprintf": {0, "fmt.Sprint or just the string itself"},
+	"fmt.Sscanf":  {1, "fmt.Sscan"},
+}
+
+func (w lintUnnecessaryFormat) Visit(n ast.Node) ast.Visitor {
+	ce, ok := n.(*ast.CallExpr)
+	if !ok || len(ce.Args) < 1 {
+		return w
+	}
+
+	funcName := gofmt(ce.Fun)
+	spec, ok := formattingFuncs[funcName]
+	if !ok {
+		return w
+	}
+
+	pos := int(spec.formatArgPosition)
+	if len(ce.Args) <= pos {
+		return w // not enough params /!\
+	}
+
+	format := gofmt(ce.Args[pos])
+
+	if !(format[0] == '"') || strings.Contains(format, `%`) {
+		return w
+	}
+
+	failure := lint.Failure{
+		Category:   lint.FailureCategoryOptimization,
+		Node:       ce.Fun,
+		Confidence: 1,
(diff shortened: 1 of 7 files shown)


ccoVeille:
Isn't there something about this in go vet or go-critic, or maybe perfsprint ?

chavacava:
> Isn't there something about this in go vet or go-critic, or maybe perfsprint ?

with 

```go
package fixtures

import (
	"fmt"
	"testing"
)

func unnecessaryFormat(t *testing.T) {
	t.Errorf("expected 4")
	fmt.Printf("nothing")
}
```

neither `go vet` or `perfsprint` detect issues. 
(didn't check with `go-critic`)

ccoVeille:
> Isn't there something about this in go vet or go-critic, or maybe perfsprint ?

I checked this morning with my daredevil config

https://github.com/ccoVeille/golangci-lint-config-examples/tree/main/90-daredevil

And nothing was reported about what you tried to detect. So it sounds good 👍

rule/unnecessary_format.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+func (w lintUnnecessaryFormat) Visit(n ast.Node) ast.Visitor {
+	ce, ok := n.(*ast.CallExpr)
+	if !ok || len(ce.Args) < 1 {
+		return w
+	}
+
+	funcName := gofmt(ce.Fun)
+	spec, ok := formattingFuncs[funcName]
```
ccoVeille, line 82: I feel this works, but I'm worried it relies on parsing the text available in code?

Reusing what is done with go vet printf would be mode robust, no?

https://cs.opensource.google/go/x/tools/+/refs/tags/v0.33.0:go/analysis/passes/printf/printf.go

chavacava: Well the situations triggering a false-positive should be rare in the wild.
From the +270 failures on k8s I didn't find a false positive

rule/unnecessary_format.go
```diff
@@ -0,0 +1,108 @@
...
+	"fmt.Scanf":   {0, "fmt.Scan"},
+	"fmt.Sprintf": {0, "fmt.Sprint or just the string itself"},
+	"fmt.Sscanf":  {1, "fmt.Sscan"},
+	// standard logging functions
+	"log.Fatalf":    {0, "log.Fatal"},
+	"log.Panicf":    {0, "log.Panic"},
+	"log.Printf":    {0, "log.Print"},
+	"logger.Fatalf": {0, "logger.Fatal"},
+	"logger.Panicf": {0, "logger.Panic"},
+	"logger.Printf": {0, "logger.Print"},
```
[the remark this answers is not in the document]
ccoVeille: Some context in the other thread

https://github.com/mgechev/revive/pull/1372#discussion_r2105898857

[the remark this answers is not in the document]
chavacava: > I mean it won't detect s.Logger.Printf

Then your worries are not about [_false positives_](https://github.com/mgechev/revive/pull/1372#discussion_r2105872916) but about missing some cases.

As I said in the description of the PR, the rule implementation tries to be _simple enough_ to detect most common cases. To detect other cases the rule will need type information and some [no-so-simple logic](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.33.0:go/analysis/passes/printf/printf.go). I think the gain in scope induced by type-info does not worth the effort (dev and maintenance)  

> Also, many non std lib have Debugf, Infof, Warnf, Noticef, Logf

From the description of the PR:
_In the future someone will ask for adding a configuration to define their own functions to check; and after that, as usual, someone else will ask for adding support for regular expressions._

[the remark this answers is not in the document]
ccoVeille: I'm sorry if my ideas mixed each other 😱 I created the conditions for creating confusion.

The logger.Printf is the one that confused me.

Because this one is not an std lib one, but a variable name.

You tested it the right way and explained it in the PR description.

Yes, the other like Warnf and co are addition  I thought about later.

ccoVeille (approved):

ccoVeille:
This rule might evolve, but for now we are good

rule/unnecessary_format.go
```diff
@@ -0,0 +1,123 @@
...
+
+	if format[0] != '"' || strings.Contains(format, `%`) {
+		return w
+	}
+
+	failure := lint.Failure{
+		Category:   lint.FailureCategoryOptimization,
+		Node:       ce.Fun,
+		Confidence: 1,
+		Failure:    fmt.Sprintf("unnecessary use of formatting function %s, you can replace it with %s", funcName, spec.alternative),
```
alexandear, line 117: I propose to quote function names like we [do in some rules](https://github.com/mgechev/revive/blob/924c70da4585d296915a5b9424a9ee97a93a6bb6/rule/enforce_repeated_arg_type_style.go#L140): `unnecessary use of formatting function "fmt.Printf", you can replace it with "fmt.Print or fmt.Println"`

```suggestion
		Failure:    fmt.Sprintf("unnecessary use of formatting function %q, you can replace it with %q", funcName, spec.alternative),
```

alexandear (approved):
Thank you for the rule, it's great.

LGTM

rule/unnecessary_format.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	pos := int(spec.formatArgPosition)
+	if len(ce.Args) <= pos {
+		return w // not enough params /!\
+	}
+
+	format := gofmt(ce.Args[pos])
+
+	if format[0] != '"' || strings.Contains(format, `%`) {
```
ccoVeille, line 109: this code is a naive approach for detecting:
- it's about a string (so not a variable)
- that doesn't contain a `%`

Can't we have a problem with a literal string ?

I added something about it in the test file

PR #1373, opened 2025-05-23
refactor: removes `get` from getters names

chavacava:
This PR removes the prefix `get` from getter methods (see [Getters](https://go.dev/doc/effective_go#Getters))

Note: this PR includes modifications in 3 functions from `main` that are not strictly getters but removing the `get` prefix makes them read better.

### DIFF
--- a/lint/failure.go
+++ b/lint/failure.go
@@ -81,7 +81,8 @@ type Failure struct {
 }
 
 // GetFilename returns the filename.
-// Deprecated: Use Filename.
+//
+// Deprecated: Use [Filename].
 func (f *Failure) GetFilename() string {
 	return f.Filename()
 }
--- a/lint/failure.go
+++ b/lint/failure.go
@@ -80,6 +80,12 @@ type Failure struct {
 	ReplacementLine string
 }
 
+// GetFilename returns the filename.
+// Deprecated: Use Filename.
+func (f *Failure) GetFilename() string {
+	return f.Filename()
+}
+
 // Filename returns the filename.
 func (f *Failure) Filename() string {
 	return f.Position.Start.Filename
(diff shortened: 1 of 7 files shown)


lint/failure.go
```diff
@@ -81,7 +81,13 @@ type Failure struct {
 }
 
 // GetFilename returns the filename.
+// Deprecated: Use Filename.
 func (f *Failure) GetFilename() string {
+	return f.Filename()
+}
```
ccoVeille, line 87: If you want to respect specs, it should be in a separated paragraph.

```suggestion
 // GetFilename returns the filename.
//
// Deprecated: Use Filename.
 func (f *Failure) GetFilename() string {
	return f.Filename()
}
```

Also, you could use godoc link to help navigation


```suggestion
 // GetFilename returns the filename.
//
// Deprecated: Use [Filename].
 func (f *Failure) GetFilename() string {
	return f.Filename()
}
```

revivelib/pattern.go
```diff
@@ -12,7 +12,13 @@ func (p *LintPattern) IsExclude() bool {
 }
 
 // GetPattern - returns the actual pattern
+// Deprecated: Use Pattern.
```
ccoVeille, line 15:
```suggestion
// GetPattern - returns the actual pattern.
//
// Deprecated: Use [Pattern].
```

revivelib/pattern.go
```diff
@@ -11,8 +11,8 @@ func (p *LintPattern) IsExclude() bool {
 	return p.isExclude
 }
 
-// GetPattern - returns the actual pattern
-func (p *LintPattern) GetPattern() string {
+// Pattern - returns the actual pattern
+func (p *LintPattern) Pattern() string {
```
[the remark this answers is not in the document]
ccoVeille: Thanks for the changes you made

alexandear (approved):
Please fix comments and can be merged.

ccoVeille (approved):

PR #1389, opened 2025-05-27
feature: rule use-fmt-print

chavacava:
`print` and `println` built-in functions are not recommended for use-cases other than [language boostraping and are not guaranteed to stay in the language](https://go.dev/ref/spec#Bootstrapping).

This PR adds a new rule, `use-fmt-print`, that spots uses of `print` and `println`, and recommends using their `fmt` equivalents.

The rule might generate false positives if the package under analysis has its own definition of `println` or `print` **functions**.

WIP: tests, doc will be added by future commits.

### DIFF
--- a/rule/use_fmt_print.go
+++ b/rule/use_fmt_print.go
@@ -70,13 +70,13 @@ func (w lintUseFmtPrint) Visit(node ast.Node) ast.Visitor {
 		Confidence: 1,
 		Node:       node,
 		Category:   lint.FailureCategoryBadPractice,
-		Failure:    fmt.Sprintf(`avoid using built-in function %q, replace by "fmt.F%s(os.Stderr, %s)"`, name, name, callArgs),
+		Failure:    fmt.Sprintf(`avoid using built-in function %q, replace it by "fmt.F%s(os.Stderr, %s)"`, name, name, callArgs),
 	})
 
 	return w
 }
 
-func (w lintUseFmtPrint) callArgsAsStr(args []ast.Expr) string {
+func (lintUseFmtPrint) callArgsAsStr(args []ast.Expr) string {
 	strs := []string{}
 	for _, expr := range args {
 		strs = append(strs, astutils.GoFmt(expr))
--- a/testdata/use_fmt_print.go
+++ b/testdata/use_fmt_print.go
@@ -16,6 +16,6 @@ func useFmtPrint() {
 	t.print("just testing")
 	t.println("just testing")
 
-	println("just testing", something)   // MATCH /avoid using built-in function "println", replace by "fmt.Fprintln(os.Stderr, "just testing", something)"/
-	print("just testing", some, thing+1) // MATCH /avoid using built-in function "print", replace by "fmt.Fprint(os.Stderr, "just testing", some, thing + 1)"/
+	println("just testing", something)   // MATCH /avoid using built-in function "println", replace it by "fmt.Fprintln(os.Stderr, "just testing", something)"/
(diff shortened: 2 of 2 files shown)


rule/use_fmt_print.go
```diff
@@ -0,0 +1,53 @@
...
+
+func (w lintUseFmtPrint) Visit(node ast.Node) ast.Visitor {
+	ce, ok := node.(*ast.CallExpr)
+	if !ok {
+		return w // nothing to do, the node is not a call
+	}
+
+	if !astutils.IsIdent(ce.Fun, "println") && !astutils.IsIdent(ce.Fun, "print") {
+		return nil // nothing to do, the call is not println(...) nor print(...)
+	}
```
ccoVeille, line 43: A switch would be more readable

rule/use_fmt_print.go
```diff
@@ -0,0 +1,64 @@
...
+
+	name := id.Name
+	replacement := "Println"
+	switch name {
+	default:
+		return nil // nothing to do, the call is not println(...) nor print(...)
+	case "println":
+	case "print":
+		replacement = "Print"
+	}
```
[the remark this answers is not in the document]
ccoVeille: https://github.com/golang/go/blob/c8c3d661b084c01242fa7837e3853d7492ba1717/src/builtin/builtin.go#L294

chavacava:
Thank you @alexandear @ccoVeille @denisvmedia for the feedback.

rule/use_fmt_print.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/mgechev/revive/internal/astutils"
+	"github.com/mgechev/revive/lint"
+)
+
+// UseFmtPrintRule lints calls to print and println.
+type UseFmtPrintRule struct{}
+
+// Apply applies the rule to given file.
+func (r *UseFmtPrintRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
```
ccoVeille, line 16: This comment is out of the scope of the PR.

This reminds me something we talked about months ago @alexandear 

The lint.Arguments is now almost, if not totally  useless.

We could add this to a list of things for a v2 (I'm unsure we can do it now)

I remember we also talked about removing goversion dependencies, and mention it as a candidate for a v2.

The v2 is only an idea now, but we could start taking a note of all ideas.

The v2 could be also a moment to move out of code that use deprecated GOROOT thing.

So here it's more a general feedback

rule/use_fmt_print.go
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		if fnDecl.Recv != nil {
+			continue // it´s a method (not function) declaration
+		}
+
+		switch fnDecl.Name.Name {
+		case "print":
+			redefinesPrint = true
+		case "println":
```
ccoVeille, line 102: Nitpicking:

print and println could have been constants, they are used twice.

testdata/use_fmt_print.go
```diff
@@ -0,0 +1,21 @@
...
+func useFmtPrint() {
+	fmt.Println("just testing")
+	fmt.Print("just testing")
+	t := useFmtPrintT{}
+	t.print("just testing")
+	t.println("just testing")
+
+	println("just testing") // MATCH /avoid using built-in function "println", use "fmt.Println" instead/
+	print("just testing")   // MATCH /avoid using built-in function "print", use "fmt.Print" instead/
+}
```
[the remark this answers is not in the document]
ccoVeille: I think I have proved many times (the PR about fmt.Printf argument for example), and here again I don't read issue or PR description.

I read them quickly to get an idea of what it talks about, then I jump into to the code review.

I'm sorry, I know I shouldn't 😅

Issue #1235, opened 2025-02-12
exported: report differently a method that is now exported


ccoVeille:
```go
package sandbox

// Foo is OK
func Foo() {}

// bar was not exported but is now exported
func Bar() {}
```

![Image](https://github.com/user-attachments/assets/37143ba2-ee31-4624-bb57-46e401aeeccb)

revive reports the right things yet, but it could be improved.

Here we are 100% sure the issue is the fact the method is now exported, but wasn't before

But because golangci-lint exclude [`EXC0012`](https://golangci-lint.run/usage/false-positives/#exc0012) by default.
This one is not reported via golangci-lint.

We should consider trying to detect it, and report another message that won't be caught by `EXC0012`

Here the idea is to compare that only the case of the first letter differ.

chavacava:
@ccoVeille what message could we report for these cases?

ccoVeille:
Let's consider #1236 with this issue for providing a wording.

So we already have this error messages

- exported (.+) should have comment( \(or a comment on this block\))? or be unexported

- comment on exported (.+) should be of the form "(.+)..."

- package comment should be of the form "(.+)..."

- should have a package comment

These are respectively, exc0012, exc0013, exc0014, exc0015

Here are some suggestions:

- invalid identifier in comment: expected 'FooBarQuux', not 'fooBarQuux'.
- invalid identifier in comment: expected 'FooBarQuux', got 'fooBarQuux'.
- comment should start with 'FooBarQuux', not 'fooBarQuux'
- comment should start with 'FooBarQuux', got 'fooBarQuux'

ccoVeille:
The main reason I wanted to work on this was the fact golangci-lint v1 was excluding almost all errors reported by the `exported` rule within revive.

It was difficult for someone to know that they had to "exclude the exclusions" via `issues.include" filed by using the obscure and cryptic EXC0012-15. Everything was documented, but it wasn't easy. 

Now golangci-lint v2 is here, thanks to the huge amount of work of @ldez and some other people like @alexandear (revive maintainer)

golangci-lint v2 cakes with a lot of changes. One of them is the concept of presets

Now, people who doesn't care about having clean godoc comments will see they are excluded.

Of course, people might have migrated blindly their golangci-lint configuration, so they do have the comments preset, right now.

But I would say, it's OK. They can see it if they look at their configuration then documentation.

TL; DR; I'm unsure we need this refactoring anymore.

So here, is the point do you guys think there is a need to report differently than now issues when:

[...]

Issue #1226, opened 2025-02-08
exported: currently only support package prefix detection, package suffix could be also useful


ccoVeille:
Currently, revive detects this

```go
// Package option comment
package option

// OptionWhatever is prefixed with the package name which is repetitive
func OptionWhatever() {} // MATCH /func name will be used as option.OptionWhatever by other packages, and that is repetitive; consider calling this Whatever/
```

I would like to detect this

```go
// Package option comment
package option

// WhateverOption is suffixed with the package name which is repetitive
func WhateverOption() {} // MATCH /func name will be used as option.WhateverOption by other packages, and that is repetitive; consider calling this Whatever/
```

The changes are pretty limited.

But I would like to open the debate. I was surprised to only see that prefix where detected.

chavacava:
Hi @ccoVeille thanks for the proposal, Indeed, checking suffixes might be interesting.

ccoVeille:
Please assign it to me. I will handle it

chavacava:
@ccoVeille some avancements on this?

ccoVeille:
No. If you want to work on it, you can of course take it back from me.

Also, let me share a kind of feeling I had: I'm not confident this won't detect a lot of things. This is what stopped me to work on this.

I wanted to do some checks on large repositories. But I didn't take the necessary time.

PR #1403, opened 2025-06-07
exported: improve error message

ccoVeille:
This PR improves the error reporting of exported rules

The idea was developed in #1235, we want to provide more information such as:

- the case changed
- the variable is now exported

Explore:
- #1226 
- #1235

### DIFF
--- a/rule/exported.go
+++ b/rule/exported.go
@@ -165,21 +165,36 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
 		return
 	}
 
-	firstCommentLine := firstCommentLine(fn.Doc)
-
-	if firstCommentLine == "" {
-		w.addFailuref(fn, 1, lint.FailureCategoryComments,
+	status := w.checkGoDocStatus(fn.Doc, fn.Name.Name)
+	switch status {
+	case exportedGoDocStatusOK:
+		return // comment is fine
+	case exportedGoDocStatusMissing:
+		w.addFailuref(fn, status.Confidence(), lint.FailureCategoryComments,
 			"exported %s %s should have comment or be unexported", kind, name,
 		)
 		return
 	}
 
-	prefix := fn.Name.Name + " "
-	if !strings.HasPrefix(firstCommentLine, prefix) {
-		w.addFailuref(fn.Doc, 0.8, lint.FailureCategoryComments,
-			`comment on exported %s %s should be of the form "%s..."`, kind, name, prefix,
-		)
+	firstCommentLine := w.firstCommentLine(fn.Doc)
+	w.addFailuref(fn.Doc, status.Confidence(), lint.FailureCategoryComments,
+		`comment on exported %s %s should be of the form "%s ..."%s`, kind, name, fn.Name.Name, status.CorrectionHint(firstCommentLine),
+	)
+}
+
+func (*lintExported) hasPrefixInsensitive(s, prefix string) bool {
+	return strings.HasPrefix(strings.ToLower(s), strings.ToLower(prefix))
+}
+
+func (*lintExported) stripFirstRune(s string) string {
+	// Decode the first rune to handle multi-byte characters.
+	firstRune, size := utf8.DecodeRuneInString(s)
+	if firstRune == utf8.RuneError {
+		return s // no valid first rune found
 	}
+
+	// Return the string without the first rune.
+	return s[size:]
 }
 
 func (w *lintExported) checkRepetitiveNames(id *ast.Ident, thing string) {
@@ -233,24 +248,24 @@ func (w *lintExported) lintTypeDoc(t *ast.TypeSpec, doc *ast.CommentGroup, first
 		return
 	}
 
+	expectedPrefix := typeName
 	for _, a := range articles {
 		if typeName == a {
 			continue
 		}
 		var found bool
 		if firstCommentLine, found = strings.CutPrefix(firstCommentLine, a+" "); found {
+			expectedPrefix = a + " " + typeName
 			break
 		}
 	}
 
-	// if comment starts with name of type and has some text after - it's ok
-	expectedPrefix := typeName + " "
-	if strings.HasPrefix(firstCommentLine, expectedPrefix) {
+	status := w.checkGoDocStatus(doc, expectedPrefix)
+	if status == exportedGoDocStatusOK {
 		return
 	}
-
-	w.addFailuref(doc, 1, lint.FailureCategoryComments,
-		`comment on exported type %v should be of the form "%s..." (with optional leading article)`, t.Name, expectedPrefix,
+	w.addFailuref(doc, status.Confidence(), lint.FailureCategoryComments,
+		`comment on exported type %v should be of the form "%s ..." (with optional leading article)%s`, t.Name, typeName, status.CorrectionHint(firstCommentLine),
 	)
 }
 
@@ -272,6 +287,7 @@ func (w *lintExported) checkValueNames(names []*ast.Ident, nodeToBlame ast.Node,
 
 	return true
 }
+
 func (w *lintExported) lintValueSpecDoc(vs *ast.ValueSpec, gd *ast.GenDecl, genDeclMissingComments map[*ast.GenDecl]bool) {
 	kind := "var"
 	if gd.Tok == token.CONST {
@@ -292,8 +308,8 @@ func (w *lintExported) lintValueSpecDoc(vs *ast.ValueSpec, gd *ast.GenDecl, genD
 		return
 	}
 
-	vsFirstCommentLine := firstCommentLine(vs.Doc)
-	gdFirstCommentLine := firstCommentLine(gd.Doc)
+	vsFirstCommentLine := w.firstCommentLine(vs.Doc)
+	gdFirstCommentLine := w.firstCommentLine(gd.Doc)
 	if vsFirstCommentLine == "" && gdFirstCommentLine == "" {
 		if genDeclMissingComments[gd] {
 			return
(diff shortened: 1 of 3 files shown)


testdata/exported_issue_1235.go
```diff
@@ -0,0 +1,46 @@
...
+// errNotFound is returned when the requested resource is not found.
+var ErrNotFound = errors.New("not found")
+
+// NEW FORMAT scope changed
+// MATCH:23 /comment on exported method StatusHTTP.Qux should start with "Qux ..." using correct scope not "qux ..."/
+
+// NEW FORMAT case change
+// MATCH:7 /comment on exported function SendJSON should start with "SendJSON ..." using correct capitalization not "SendJson ..."/
+// MATCH:18 /comment on exported method StatusHTTP.FooBar should start with "FooBar ..." using correct capitalization not "Foobar ..."/
+// MATCH:28 /comment on exported function SendJSON should start with "SendJSON ..." using correct capitalization not "SendJson ..."/
```
ccoVeille: The current format is like this

// MATCH:23 /comment on exported method StatusHTTP.Qux should be of the form "Qux ..."/

// MATCH:7 /comment on exported function SendJSON should be of the form "SendJSON ..."/
// MATCH:18 /comment on exported method StatusHTTP.FooBar should be of the form "FooBar ..."/
// MATCH:28 /comment on exported function SendJSON should be of the form "SendJSON ..."/

ccoVeille: I'm unsure there is a need for a special wording when the scope changed, but it's mostly because I dislike all the idea I had about how to report them.

ccoVeille: @alexandear @denisvmedia @chavacava

alexandear: Thanks for the examples.

I think we should avoid changing the error messages in v1 unless there’s a very specific reason to do so.

1. The `exported` rule and its wording come directly from `golint` ([source](https://github.com/golang/lint/blob/6edffad5e6160f5949cdefc81710b2706fbcd4f6/testdata/4.go#L17)) and are well-known to most Go developers.
2. Similar phrasing is also used in `staticcheck` ([example](https://github.com/dominikh/go-tools/blob/bbc2f4dd71ea1dd1f0801fd7b7c4b7a22b9a90ba/stylecheck/st1020/testdata/go1.0/CheckExportedFunctionDocs/CheckExportedFunctionDocs.go#L20C23-L20C49)).
3. Lastly, changing the message format might unintentionally break some `golangci-lint` configurations ([example](https://sourcegraph.com/search?q=context:global+comment+on+exported+method&patternType=standard&sm=1)) that rely on specific message text for filtering or suppressing.

So I’d vote for keeping them as-is unless there’s a compelling reason to diverge. Let's see what others say.

ccoVeille: I'm a bit sad and disappointed.

Unless I have misunderstood your message, your comment goes against the whole discussion I initiated and we discussed in the issue

https://github.com/mgechev/revive/issues/1235#issuecomment-2797320667

And let me think I should close my PR.

My [question](https://github.com/mgechev/revive/pull/1403#discussion_r2133970926) was about the fact I wasn't sure there was a need to distinct case change from scope change, as you can see in the code and error message.

ccoVeille: Another way to consider the wording is to append things to the existing error message

// MATCH:7 /comment on exported function SendJSON should be of the form "SendJSON ..." not "SendJson"/
// MATCH:18 /comment on exported method StatusHTTP.FooBar should be of the form "FooBar ..." not "Foobar" /
// MATCH:28 /comment on exported function SendJSON should be of the form "SendJSON ..." not "sendJSON"/

But if we don't append additional information such as

// MATCH:7 /comment on exported function SendJSON should be of the form "SendJSON ..." not "SendJson" (case differs)/
// MATCH:18 /comment on exported method StatusHTTP.FooBar should be of the form "FooBar ..." not "Foobar" (case differs)/
// MATCH:28 /comment on exported function SendJSON should be of the form "SendJSON ..." not "sendJSON" (scope differs)/

I feel like the change that this PR could bring is pretty useless…

ccoVeille: @alexandear

alexandear: Thanks for the follow-up, @ccoVeille, and I’m sorry to hear that my previous message felt dismissive - that definitely wasn’t the intention.

You're absolutely right that your PR aims to improve the wording when there's a mismatch in either case or scope, and I appreciate the concrete examples you've provided. They do make a strong case for how additional context like `case differs` or `scope differs` could help users better understand what went wrong, especially in ambiguous cases.

My main concern was around changing existing messages in a way that could break users' filtering or expectations, particularly for long-standing rules. But I see now that your proposal is more about augmenting the message with clarifying info rather than changing its core.

So perhaps a compromise could be: keep the existing format intact, but conditionally append hints like `case differs` or `scope differs` only when they apply. That might bring the clarity you're aiming for without introducing regressions.

Does that sound reasonable to you?

ccoVeille:
EDIT: move message in the right thread

https://github.com/mgechev/revive/pull/1403#discussion_r2155205669

testdata/exported_issue_1235.go
ccoVeille: I was also about to suggest this, but I was awaiting for feedbacks from others.

So yes, I'm fine with it. I'll proceed to the changes when I will have some free time.

I'm still interested by feedbacks from @denisvmedia and @chavacava

ccoVeille: I posted wording suggestions into the issue 

https://github.com/mgechev/revive/issues/1235#issuecomment-2985421816

ccoVeille: I updated the PR with the suggestion 1 I talked [here](https://github.com/mgechev/revive/issues/1235#issuecomment-2985454225)

ccoVeille:
@alexandear @denisvmedia @chavacava 

Can I have a feedback about the changes I'm suggesting here? 

Thanks

The PR is in Draft because I have multiple place where I will have to update the wordings.

@alexandear I don't think there is a need to postpone revive 1.11 version for this one.

It's a minor change

rule/exported.go
```diff
@@ -174,12 +174,58 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
...
+		return
 	}
+
+	w.addFailuref(fn.Doc, 1, lint.FailureCategoryComments,
+		`comment on exported %s %s should be of the form "%s..." by using its correct casing, not "%s ..."`,
+		kind, name, expectedPrefix, currentPrefix,
+	)
+}
+
+func (*lintExported) HasPrefixFold(s, prefix string) bool {
```
chavacava: This method could be private.

rule/exported.go
```diff
@@ -174,12 +174,58 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
...
+	return strings.HasPrefix(strings.ToLower(s), strings.ToLower(prefix))
+}
+
+func (*lintExported) StripFirstRune(s string) string {
```
chavacava: This method could be private.

rule/exported.go
```diff
@@ -174,12 +174,58 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
...
+	}
+
+	// Decode the first rune to handle multi-byte characters.
+	firstRune, size := utf8.DecodeRuneInString(s)
+	if firstRune == utf8.RuneError {
+		return s // no valid first rune found
+	}
+
+	// Return the string without the first rune.
+	return s[size:]
```
chavacava: Decoding an empty string returns an error thus, we could write something like:

```suggestion
	// Decode the first rune to handle multi-byte characters.
	firstRune, size := utf8.DecodeRuneInString(s)
	
	switch firstRune {
	case utf8.RuneError: // no valid first rune found
		return s
	default:
		// Return the string without the first rune.
		return s[size:]
	}
```

rule/exported.go
chavacava: The method does not actually fold strings; it's a case-insensitive version of `strings.HasPrefix`
Maybe 
```suggestion
func (*lintExported) HasPrefixInsensitive(s, prefix string) bool {
```
describes better what the method does.

ccoVeille: I inspired myself from strings.EqualFold, but I'm fine with your wording

chavacava: My proposal is because, as I understand,  _fold_ implies using `unicode.SimpleFold` 

https://cs.opensource.google/go/go/+/master:src/strings/strings.go;drc=c0eb7ab3062fae802639545770f6a897d75b0085;l=1260

rule/exported.go
```diff
@@ -174,12 +174,58 @@ func (w *lintExported) lintFuncDoc(fn *ast.FuncDecl) {
 		return
 	}
 
-	prefix := fn.Name.Name + " "
-	if !strings.HasPrefix(firstCommentLine, prefix) {
+	expectedPrefix := fn.Name.Name + " "
```
chavacava: I propose something like 

```go
	expectedPrefix := fn.Name.name+" "
	currentPrefix := strings.Split(firstCommentLine, " ")[0]
	switch w.comparePrefixes(expectedPrefix, actualPrefix) {
	case prefixesAreEqual:
		return
	case prefixesAreEqualCaseInsensitive:
		// They have the same prefix, but the case is different.
		// For example, the comment starts with "SendJson" but should start with "SendJSON".
		w.addFailuref(fn.Doc, 1, lint.FailureCategoryComments,
			`comment on exported %s %s should be of the form "%s..." by using its correct casing, not "%s ..."`,
			kind, name, expectedPrefix, currentPrefix,
		)
	case prefixesOnlyDifferFirstCharacterCase:
		// Only the first character differs, such as "sendJSON" became "SendJSON".
		// so we consider the scope has changed.
		w.addFailuref(fn.Doc, 1, lint.FailureCategoryComments,
			`comment on exported %s %s should be of the form "%s..." to match its exported status, not "%s ..."`,
			kind, name, expectedPrefix, currentPrefix,
		)
	default: // Prefixes are not equal
		// The comment does not start with the expected prefix.
		w.addFailuref(fn.Doc, 0.8, lint.FailureCategoryComments,
			`comment on exported %s %s should be of the form "%s..."`, kind, name, expectedPrefix,
		)
	}

	return
	
type prefixComparationResult int
const (
	prefixesAreNotEqual prefixComparationResult = -1
	prefixesAreEqual prefixComparationResult = 0
	prefixesOnlyDifferFirstCharacterCase prefixComparationResult = 1
	prefixesAreEqualCaseInsensitive prefixComparationResult = 2
)

func (*lintExported) comparePrefixes(p1, p2 string) prefixComparationResult {
	if len(p1) != len(p2) { // just an optim to quit fast if !equal
		return prefixesAreNotEqual
	}

	if p1 == p2 {
		return prefixesAreEqual
	}

	if strings.ToLower(p1) == strings.ToLower(p2) {
		if w.StripFirstRune(p1) == w.StripFirstRune(p2) {
			return prefixesOnlyDifferFirstCharacterCase
		}

		return prefixesAreEqualCaseInsensitive
	}

	return prefixesAreNotEqual
}
```
(/!\ written on notepad thus I don't know if it compiles)

rule/exported.go
ccoVeille: Here we are reaching something that could be about a new rule, or a new case for an existing rule.

The struct is unexported, so should we have only unexported method?

I think we talked about something close to this in a previous PR, no sure which ones

The counter example I can think about are the method like Error() string, and maybe others.

rule/exported.go
ccoVeille: I like the idea, but I'm unsure.

The detector returning a constant doesn't sounds to be idiomatic. I would use sentinel errors maybe.

Also, I feel like there is something wrong with the way the message is repeated.

`comment on exported %s %s should be of the form "%s..."`, kind, name, expectedPrefix,

I'm the one who coded it, I feel like I should have used an optional extra information.

I will have to think about all this, but thanks for the idea

chavacava: Some remarks/comments:
1. I made a _proposal_ ie not something that I consider a necessary condition to go ahead with the PR. In other words: I'm OK with the code as is.
2. The proposed code might have compiler and semantic errors (for example, I'm sure the trailing space in prefixes is not handled correctly)
3. I'm not comfortable with the idea of using errors to encode non-error information (the result of a comparison is not an error)
4. When I refactored the code into the proposed one I was not comfortable with the repetition of `w.addFailuref(...)` in each switch-case. I think we could just set `msg` and `confidence` variables in the switch-cases and add a single call to `w.addFailuref(...)` after the switch. That will make the code more "light" to read.

rule/exported.go
ccoVeille: Here was the thread I think

- https://github.com/mgechev/revive/pull/1344#discussion_r2087127972

ccoVeille: I created an issue from current thread

- https://github.com/mgechev/revive/issues/1425

ccoVeille:
I finally worked on it.

Please review

Side note: I opened #1431 to work on this as it was a nightmare to debug without changes.

- #1431

rule/exported.go
```diff
@@ -325,13 +343,71 @@ func (w *lintExported) lintValueSpecDoc(vs *ast.ValueSpec, gd *ast.GenDecl, genD
...
+
+func (gds exportedGoDocStatus) Confidence() float64 {
+	switch gds {
+	case exportedGoDocStatusOK, exportedGoDocStatusCaseMismatch, exportedGoDocStatusFirstLetterMismatch:
+		return 1
+	}
+	return 0.8
+}
+
+func (gds exportedGoDocStatus) Format(firstLine string) string {
```
ccoVeille: I'm not inspired for this method name, please be kind and suggest me better names

ccoVeille: I reworded to CorrectionHint

chavacava (approved):

alexandear (approved):

ccoVeille:
@babakks maybe you will find interest in this PR for https://github.com/godoc-lint/godoc-lint

PR #1408, opened 2025-06-18
unexported-return: fix type aliases support

### LINKED ISSUE #1406: rule.UnexportedReturn: false positive when return type is exported but an alias to an unexported type

ldez:
Fixes #1406
Closes #1407

### DIFF
--- a/rule/unexported_return.go
+++ b/rule/unexported_return.go
@@ -78,7 +78,18 @@ func (*UnexportedReturnRule) Name() string {
 // It is imprecise, and will err on the side of returning true,
 // such as for composite types.
 func exportedType(typ types.Type) bool {
-	switch t := types.Unalias(typ).(type) {
+	switch t := typ.(type) {
+	case *types.Alias:
+		obj := t.Obj()
+		switch {
+		// Builtin types have no package.
+		case obj.Pkg() == nil:
+		case obj.Exported():
+		default:
+			_, ok := t.Underlying().(*types.Interface)
+			return ok
+		}
+		return true
 	case *types.Named:
 		obj := t.Obj()
 		switch {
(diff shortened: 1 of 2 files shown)


alexandear (approved):
Thank you!

PR #1412, opened 2025-06-30
docs: improve unhandled-error ignore argument examples

scop:
Anchor at beginning/end as appropriate to avoid false positives, include ignorable string.Writer functions, avoid matching bytes.Buffer.WriteTo.


alexandear (approved):
LGTM

PR #1427, opened 2025-07-14
feature: add warnings on 'slices.Delete' in 'modifies-parameter' check

vbvictor:
Add warnings when function parameter was passed to `slices.Delete` or `slices.DeleteFunc` function.
The implementation is kind-of hardcoded right now. I can make it more future-proof by creating a list of function names with positions of arguments it modifies if it even needed, WDYT?

Closes https://github.com/mgechev/revive/issues/1410.

### DIFF
--- a/rule/modifies_param.go
+++ b/rule/modifies_param.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"go/ast"
 
+	"github.com/mgechev/revive/internal/astutils"
 	"github.com/mgechev/revive/lint"
 )
 
@@ -57,11 +58,21 @@ func (w lintModifiesParamRule) Visit(node ast.Node) ast.Visitor {
 		}
 	case *ast.AssignStmt:
 		lhs := v.Lhs
-		for _, e := range lhs {
+		for i, e := range lhs {
 			id, ok := e.(*ast.Ident)
 			if ok {
+				if i < len(v.Rhs) {
+					if callExpr, ok := v.Rhs[i].(*ast.CallExpr); ok && isSlicesDelete(callExpr) {
+						w.checkSlicesDelete(callExpr)
+						continue
+					}
+				}
 				checkParam(id, &w)
 			}
+		}
+	case *ast.ExprStmt:
+		if callExpr, ok := v.X.(*ast.CallExpr); ok && isSlicesDelete(callExpr) {
+			w.checkSlicesDelete(callExpr)
 		}
 	}
 
@@ -75,6 +86,31 @@ func checkParam(id *ast.Ident, w *lintModifiesParamRule) {
 			Node:       id,
 			Category:   lint.FailureCategoryBadPractice,
 			Failure:    fmt.Sprintf("parameter '%s' seems to be modified", id),
+		})
+	}
+}
+
+func isSlicesDelete(callExpr *ast.CallExpr) bool {
+	return astutils.IsPkgDotName(callExpr.Fun, "slices", "Delete") ||
+		astutils.IsPkgDotName(callExpr.Fun, "slices", "DeleteFunc")
+}
+
+func (w *lintModifiesParamRule) checkSlicesDelete(callExpr *ast.CallExpr) {
+	if len(callExpr.Args) == 0 {
+		return
+	}
+
+	if id, ok := callExpr.Args[0].(*ast.Ident); ok && w.params[id.Name] {
+		funcName := "function"
+		if sel, ok := callExpr.Fun.(*ast.SelectorExpr); ok {
+			funcName = fmt.Sprintf("%s.%s", sel.X, sel.Sel.Name)
+		}
+
+		w.onFailure(lint.Failure{
+			Confidence: 1, // slices.Delete/DeleteFunc always modifies
+			Node:       callExpr,
+			Category:   lint.FailureCategoryBadPractice,
+			Failure:    fmt.Sprintf("parameter '%s' is modified by %s", id.Name, funcName),
 		})
 	}
 }
--- a/testdata/modifies_param.go
+++ b/testdata/modifies_param.go
@@ -1,5 +1,7 @@
 package fixtures
 
+import "slices"
+
 func one(a int) {
 	a, b := 1, 2 // MATCH /parameter 'a' seems to be modified/
 	a++          // MATCH /parameter 'a' seems to be modified/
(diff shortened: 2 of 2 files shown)


chavacava:
Hi @vbvictor, thanks for the PR !
Indeed, a more flexible (less hardcoded) approach might be more convenient for easily adding other cases in the future.

The rule `string-format` has something similar for allowing users to define new cases. I think that  we can keep the mechanism internal to `modifies-parameter` implementation and, in the future if required, open it for users through configuration.

vbvictor:
Updated to use a predefined list of function name we are interested in, PTAL

internal/astutils/ast_utils.go
```diff
@@ -159,3 +159,26 @@ func GoFmt(x any) string {
 	gofmtConfig.Fprint(&buf, fs, x)
 	return buf.String()
 }
+
+// GetCallName return the name of a call expression in the form of package.Func or Func.
+func GetCallName(call *ast.CallExpr) (callName string, ok bool) {
```
chavacava: This utilisty function seems not to apply to general cases with more than 2 selectors (something [very common](https://github.com/search?q=%2F.*%5C..*%5C..*%5C..*%5C%28.*%5C%29%2F+language%3AGo&type=code&ref=advsearch))

Does a call to `astutils.GoFmt` is not enough?

vbvictor: Indeed, `astutils.GoFmt` is enough.
I suspected it could be slower, but benchmark tells is no difference.

rule/modifies_param.go
```diff
@@ -78,3 +98,32 @@ func checkParam(id *ast.Ident, w *lintModifiesParamRule) {
 		})
 	}
 }
+
+func isModifyingFunction(callExpr *ast.CallExpr) bool {
```
chavacava: This function seems not necessary. Its body is almost the same as the first lines of `checkModifyingFunction` therefore we could completely remove `isModifyingFunction` and let `checkModifyingFunction` do the job.

rule/modifies_param.go
```diff
@@ -78,3 +98,32 @@ func checkParam(id *ast.Ident, w *lintModifiesParamRule) {
...
+func (w *lintModifiesParamRule) checkModifyingFunction(callExpr *ast.CallExpr) {
+	funcName := astutils.GoFmt(callExpr.Fun)
+	positions, found := modifyingFunctions[funcName]
+	if !found {
+		return
+	}
+
+	for _, pos := range positions {
+		if pos >= len(callExpr.Args) {
+			continue
```
chavacava, line 118: if positions are in ascending order we could stop iterating (`break`) here.

rule/modifies_param.go
```diff
@@ -78,3 +98,32 @@ func checkParam(id *ast.Ident, w *lintModifiesParamRule) {
...
+		}
+
+		if id, ok := callExpr.Args[pos].(*ast.Ident); ok && w.params[id.Name] {
+			w.onFailure(lint.Failure{
+				Confidence: 0.5, // confidence is low because of shadow variables
+				Node:       callExpr,
+				Category:   lint.FailureCategoryBadPractice,
+				Failure:    fmt.Sprintf("parameter '%s' is modified by %s", id.Name, funcName),
```
chavacava: Maybe we can highlight the low confidence in the failure message:
```suggestion
				Failure:    fmt.Sprintf("parameter '%s' seems to be modified by %s", id.Name, funcName),
```

rule/modifies_param.go
ccoVeille: I also feel like checkModifyingFunction could accept an ast.Node and return if it cannot be converted to ast.CallExpr

It would simplify how the method is called

vbvictor: Done

vbvictor:
Made `checkModifyingFunction` return `bool` to skip double-warning for case like 
```go
func (s []int) {
  s = slices.Delete(s, 0, 1) // 's' modified both by '=' and `Delete`
}
```
It seems like test infrastructure does not support multiple diagnostics on the same line (or I did not dig deep enough).

ccoVeille:
> It seems like test infrastructure does not support multiple diagnostics on the same line (or I did not dig deep enough).

You can find an example here

https://github.com/mgechev/revive/blob/master/testdata%2Ftime_date_out_of_bounds.go#L59-L61

rule/modifies_param.go
```diff
@@ -57,12 +67,17 @@ func (w lintModifiesParamRule) Visit(node ast.Node) ast.Visitor {
...
-		for _, e := range lhs {
+		for i, e := range lhs {
 			id, ok := e.(*ast.Ident)
 			if ok {
+				if i < len(v.Rhs) {
+					w.checkModifyingFunction(v.Rhs[i])
+				}
 				checkParam(id, &w)
 			}
 		}
```
ccoVeille:
```suggestion
		for i, e := range lhs {
 			id, ok := e.(*ast.Ident)
 			if !ok {
				continue
			}

			if i < len(v.Rhs) {
				w.checkModifyingFunction(v.Rhs[i])
			}
 			checkParam(id, &w)
 		}
```

rule/modifies_param.go
```diff
@@ -78,3 +93,31 @@ func checkParam(id *ast.Ident, w *lintModifiesParamRule) {
...
+		}
+
+		if id, ok := callExpr.Args[pos].(*ast.Ident); ok && w.params[id.Name] {
+			w.onFailure(lint.Failure{
+				Confidence: 0.5, // confidence is low because of shadow variables
+				Node:       callExpr,
+				Category:   lint.FailureCategoryBadPractice,
+				Failure:    fmt.Sprintf("parameter '%s' seems to be modified by '%s'", id.Name, funcName),
+			})
+		}
```
ccoVeille: What about this?

```suggestion
		id, ok := callExpr.Args[pos].(*ast.Ident)
		if !ok {
			continue
		}

		_, match := w.params[id.Name]
		if !match {
			continue
		}
		
		w.onFailure(lint.Failure{
			Confidence: 0.5, // confidence is low because of shadow variables
			Node:       callExpr,
			Category:   lint.FailureCategoryBadPractice,
			Failure:    fmt.Sprintf("parameter '%s' seems to be modified by '%s'", id.Name, funcName),
		})
```

vbvictor:
Before merging, workflows of the PR should be approved to run.
Unfortunately, you need to press the button for each commit I make.

PR #1436, opened 2025-07-24
new rule: enforce-else

chavacava:
**The rule was removed by #1495**

---

This PR implements `enforce-else`, a new rule, that enforces the presence of an `else` branch in the last `if` of an `if ... else if` chain.

This is the equivalent of enforcing `default` case in `switch` statements.

The rule will not report `if ... else if` chains where all the (then) branches end with `return` or `break`

P.S.
In practice, the rule generates too many failures (for example, +1800 failures in the Kubernetes code base)
Even if the rule enforces explicit completeness of conditionals, it evidently doesn't fit with the Go way of doing things.

### DIFF
--- a/.github/instructions/go.instructions.md
+++ b/.github/instructions/go.instructions.md
@@ -1,0 +1,247 @@
+---
+description: 'Instructions for writing Go code following idiomatic Go practices and community standards'
+applyTo: '**/*.go,**/go.mod,**/go.sum'
+---
+
+# Go Development Instructions
+
+Follow idiomatic Go practices and community standards when writing Go code.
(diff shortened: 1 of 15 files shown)


ccoVeille (approved):

ccoVeille:
Do you have a real life example of what it would report on real code?

I'm unsure about this rule. I'm not saying I'm against it. I simply doesn't get what this rulebis trying to solve

chavacava:
The rule is semantically equivalent to that enforcing `default` in `switch` but applied to if...else if chains.
cf https://rules.sonarsource.com/go/RSPEC-126

PR #1448, opened 2025-07-30
feature: detect identical-branches in switch statements

### LINKED ISSUE #1442: Switch identical branches detection
#1440 makes me think about use cases with switch 

```go
switch a {
// expected values
case 1:
   foo()
case 2:
   bar()
case 3:
   foo()
default:
   return errors.New("blah")
}
```

I have no computer with me unfortunately, I cannot check, maybe it exists, or ... it becomes a new linter idea.

chavacava:
This PR extends `identical-branches` rule to analyse `switch` statements.

Notes:
1. Only un-tagged switches are not analysed because the order of case evaluation might be important thus two identical branches might exist but can not be merged into one case.
2. Branches ending with a `fallthrough` are not taken into account because the semantics of such branches is defined by the subsequent branch.
3. There are a lot of identical switch branches out there. For example, running the rule against Kubernetes code-base results in ~290 failures. Most of these failures do not spot actual logic problems in the code but more a "style" problem.

Closes #1442

### DIFF
--- a/rule/identical_branches.go
+++ b/rule/identical_branches.go
@@ -1,10 +1,9 @@
 package rule
 
 import (
-	"crypto/md5"
-	"encoding/hex"
 	"fmt"
 	"go/ast"
+	"go/token"
 
 	"github.com/mgechev/revive/internal/astutils"
 	"github.com/mgechev/revive/lint"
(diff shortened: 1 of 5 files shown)


rule/identical_branches.go
```diff
@@ -67,15 +66,67 @@ func (w *lintIdenticalBranches) Visit(node ast.Node) ast.Visitor {
...
+		if n.Tag == nil {
+			return w // do not lint untagged switches (order of case evaluation might be important)
+		}
+
+		w.lintSwitch(n)
+		return nil // switch branches already analyzed
 	default:
 		return w
 	}
 }
```
ccoVeille, line 78: I would write this, but it's opinionated maybe

```go
func (w *lintIdenticalBranches) Visit(node ast.Node) ast.Visitor {
        switch n := node.(type) {
        case *ast.IfStmt:
                if w.isIfElseIf(n) {
                        walker := &lintIfChainIdenticalBranches{
                                onFailure:  w.onFailure,
                                file:       w.file,
                                rootWalker: w,
                        }

                        ast.Walk(walker, n)
                        return nil // the walker already analyzed inner branches
                }

                w.lintSimpleIf(n)

        case *ast.SwitchStmt:
                if n.Tag == nil {
                        break // do not lint untagged switches (order of case evaluation might be important)
                }

                w.lintSwitch(n)
                return nil // switch branches already analyzed
        }

        return w
}
```

testdata/identical_branches.go
```diff
@@ -106,11 +106,80 @@ func identicalBranches() {
...
+	switch a {
+	case 1:
+		foo()
+		fallthrough
+	case 2:
+		fallthrough
+	case 3:
+		foo()
+	case 4:
+		fallthrough
```
ccoVeille, line 181: Testing fallthrought is great.

Could you add an example with a goto ?



I dislike them, but people can use them

testdata/identical_branches.go
```diff
@@ -106,11 +106,80 @@ func identicalBranches() {
...
+
+	// Do not warn on fallthrough
+	switch a {
+	case 1:
+		foo()
+		fallthrough
+	case 2:
+		fallthrough
+	case 3:
+		foo()
```
ccoVeille, line 179: Here it would be interesting to consider reporting empty fallthrought

Maybe with an other PR anyway, and maybe another rule

But this would be expected
```suggestion
	case 2, 3:
		foo()
```

testdata/identical_branches.go
chavacava: Could you please give an example with `goto`?

As I understand, for this analysis, `goto` should be considered as an ordinary statement.

```go
switch a { // the rule must warn on both branches being identical
  case 0:
    goto label
  case 1:
    goto label
}
```

testdata/identical_branches.go
chavacava: I think it could be another rule.

rule/identical_branches.go
chavacava: I'll keep the `default`

testdata/identical_branches.go
ccoVeille: You are right. For this rule everything is fine.

testdata/identical_branches.go
ccoVeille: I opened

- https://github.com/mgechev/revive/issues/1450

PR #1451, opened 2025-07-31
feature: new rule identical-switch-conditions

### LINKED ISSUE #1444: New rule to detect identical conditions in `if ... else if` chains and in `switch`
In long `if ... else if` chains there is the risk of repeating a condition.

```go
if a {
   foo()
} else if b {
    bar()
} else if c {
    foo1()
} else if b { // duplicated condition
    bar2()
}
```

We could add a rule to spot such cases.

chavacava:
This PR adds `identical-switch-conditions`, a new rule to warn on switch case clauses with repeated conditions.

(partially) Closes #1444: I propose to have 2 rules: one for switch and another for if...else if.

/!\ Will add doc in a future commit. Open to review.

cc: @ccoVeille

### DIFF
--- a/rule/identical_switch_conditions.go
+++ b/rule/identical_switch_conditions.go
@@ -1,0 +1,77 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+
+	"github.com/mgechev/revive/internal/astutils"
+	"github.com/mgechev/revive/lint"
+)
+
+// IdenticalSwitchConditionsRule warns on switch case clauses with identical conditions.
+type IdenticalSwitchConditionsRule struct{}
+
+// Apply applies the rule to given file.
+func (*IdenticalSwitchConditionsRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	onFailure := func(failure lint.Failure) {
+		failures = append(failures, failure)
+	}
+
+	w := &lintIdenticalSwitchConditions{file: file, onFailure: onFailure}
+	for _, decl := range file.AST.Decls {
+		fn, ok := decl.(*ast.FuncDecl)
+		if !ok || fn.Body == nil {
+			continue
+		}
+
+		ast.Walk(w, fn.Body)
+	}
+
+	return failures
+}
+
+// Name returns the rule name.
+func (*IdenticalSwitchConditionsRule) Name() string {
+	return "identical-switch-conditions"
+}
+
+type lintIdenticalSwitchConditions struct {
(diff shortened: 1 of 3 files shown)


rule/identical_switch_conditions.go
```diff
@@ -0,0 +1,77 @@
...
+	return failures
+}
+
+// Name returns the rule name.
+func (*IdenticalSwitchConditionsRule) Name() string {
+	return "identical-switch-conditions"
+}
+
+type lintIdenticalSwitchConditions struct {
+	file      *lint.File // only necessary to retrieve the line number of branches
```
ccoVeille: If you only needs the ToPosition method, why not storing only the function ?

So here you could have 

func (token.Pos) token.Position

ccoVeille (approved):

alexandear (approved):
Looks good. You can add a documentation.

Let me try this rule on a few popular repositories and see how it behaves.

rule/identical_switch_conditions.go
chavacava: Good catch!

testdata/identical_switch_conditions.go
```diff
@@ -0,0 +1,41 @@
...
+	case value:
+	default:
+	}
+
+	switch {
+	case a > 0, a < 0:
+	case a == 0:
+	case a < 0: // MATCH /case clause at line 11 has the same condition/
+	default:
+	}
```
ccoVeille, line 15: could you add a test about

```go
switch {
  case incr() > 0:
  case incr() > 0:
}
```

I'm unsure whether it should be reported with switch-identical-conditions. Or this pattern should be discouraged and this could be a candidate for another rule

ccoVeille (approved):
Some minor remarks

testdata/identical_switch_conditions.go
chavacava: The rule catch it as a repeated condition and generates a failure. These are corner cases that deserve a (not so) "false positive"

chavacava: Recently I saw this:
```go
switch {
	case lnOpts.IsSocketOpts():
		// ...
		// check for timeout
		fallthrough
	case lnOpts.IsTimeout(), lnOpts.IsSocketOpts():
		// timeout listener with socket options.
		// ...
	case lnOpts.IsTimeout():
		// ...
	default:
		// ...
	}
```

ccoVeille: My point was to add a test about it.

Your example is good, it could be added

PR #1453, opened 2025-08-01
Split `identical-branches` into 3 rules

chavacava:
The original `identical-branches` targeted simple `if ... else ...` statements; thus when a failure was reported the failure was, most of the time, a bug.

After #1448 the rule also copes with `switch` and `if ... else if`. For these statements, the failures are, most of the time, mainly caused by the "style" of the code and less by logic errors.

This PR proposes to split the rule into 3 rules:
1. `identical-branches` that does was the original rule did,
2. `identical-switch-branches` to cope with switch statements, and
3. `identical-ifelseif-branches` for `if ... else if`

By doing so, users will have fine control on what analysis to activate and allow reducing the (potential) noise from switch and ifelseif chains.


denisvmedia:
I'm wondering if making a single rule configurable can be a better option.

chavacava:
> I'm wondering if making a single rule configurable can be a better option.

In fact, I'm proposing this split because I fear users asking for adding configuration to the rule. :)

veeam-denis:
Do you mean them not knowing how to configure it? Because that's what I'm asking about: having a **configurable** rule to me can be a better option (unless you think this can be more confusing than having separate rules). To me having 3 separate rules sounds more confusing, and I'm afraid we'll end up having false positive bug reports (when the rule does not work how they expect it to work).

P.S. Sorry, again used the wrong account, but it's me (@denisvmedia)

chavacava:
I, as user and developer of `revive`, prefer rules without configuration. The current implementation of the rule will certainly trigger users demanding things like "add a config to deactivate reporting identical branches in switch statements" (even I myself find the rule too noisy)

By splitting the rule we allow deactivating rules with ease (`disabled: true`) and, also important, without adding complexity to the rule itself.

ccoVeille:
From my experience, it's better to have multiple rules. Than having a rule with multiple settings.

People are lazy, and it's a good thing.

People enables things without reading documentation. So if they get too many errors with one rule, they disable it. They don't look for documentation.

I got the issue with enforce-switch-style that expect having a default case, while only being about the fact that default case should be at the end or beginning. People disabled the rule instead of looking for a hypothetical setting.

I know it for sure because this people was me 😅🤣

rule/identical_ifelseif_branches.go
```diff
@@ -0,0 +1,186 @@
...
+	})
+
+	return len(calls) > 0
+}
+
+// identicalBranches yields pairs of (line numbers) of identical branches from the given branches.
+func (w *lintIfChainIdenticalBranches) identicalBranches(branches []ast.Stmt) [][]int {
+	result := [][]int{}
+	if len(branches) < 2 {
+		return result // only one branch to compare thus we return
```
ccoVeille: 🎨

```suggestion
		return result //no other branch to compare thus we return
```

veeam-denis:
Well, I don't have other arguments. I can live with both options :) We should just maybe be pretty explicit in the rule documentation and let users know that for different situations we have 3 different rules.

testdata/identical_switch_branches.go
```diff
@@ -0,0 +1,63 @@
+package fixtures
```
veeam-denis, line 1: Can we have a switch type case?

chavacava: Yes.
(Notice that duplicated clauses on switch type are detected by the compiler)

denisvmedia: Clauses, yes, but I thought we were to detect identical branches, e.g.: https://go.dev/play/p/-C547qnA2Av (if the rule catches it, it would be a false positive)

denisvmedia:
@chavacava thanks for the update. I actually came up to another example: https://go.dev/play/p/788Y5gDbgjP
In this case, they are real dups. The whole switch/case in this case is unnecessary. But if there was no `default` (or it had different instructions), then the other cases would be combinable as `case int, string, bool, float64`.

chavacava:
The AST node for "type switch" is not the same as that for "standard switch"; the rule only analyzes "standard switch" nodes.

testdata/identical_switch_branches.go
```diff
@@ -0,0 +1,77 @@
...
+	// expected values
+	case 1:
+		foo()
+	case 2:
+		bar()
+	case 3:
+		foo()
+	default:
+		bar()
+	}
```
ccoVeille, line 28: Can you also add an example for

```go
switch a {
  case 1:
  case 2:
  case 3:
   a++
  default:
   a=0
}
```

Where case 1,2 should be grouped


It's the opposite of fallthrought logic

denisvmedia:
> The AST node for "type switch" is not the same as that for "standard switch"; the rule only analyzes "standard switch" nodes.

@chavacava yes, I understand. But tests also are to protect us from future refactorings and also serve as self-documentation (also they can be analyzed by AI).

We may put a comment saying that the rule does not mean to process it, but the test case is there to make sure we don't process it by mistake.

chavacava:
That's why I've added the first example you proposed.

denisvmedia:
@chavacava but do we **not** want to detect identical branches in type switch here? Because there is a valid case when we can have it.

chavacava:
> @chavacava but do we **not** want to detect identical branches in type switch here? Because there is a valid case when we can have it.

Yes, we could. Type switches are a little bit trickier than standard switches because case clauses can be combined or not depending on the type you are matching therefore, in some situations (like for example when matching AST nodes by type as we do in revive) there is no alternative to duplicating the branches if you need to do the same thing for 2 types. We could report these cases with a lower confidence (even for those cases a duplicated branch might be a mistake)

I propose to merge this PR to separate the rules, and then work on `identical-switch-branches` to tacle all these détails.

ccoVeille:
> I propose to merge this PR to separate the rules, and then work on `identical-switch-branches` to tacle all these détails.

I agree

> there is no alternative to duplicating the branches if you need to do the same thing for 2 types.

This !

> We could report these cases with a lower confidence (even for those cases a duplicated branch might be a mistake)

Also be careful here, type switch with type can be tricky because you might have to do something like this

```go
package main

func foo[T int64 | float64](v T) {
	// do something with v
	_ = v
}

func main() {
	var a any
	switch v := a.(type) {
	case int64:
		foo(v)
	case float64:
		foo(v)
	default:
		panic("not happy")
	}
}

```

You cannot use `case int64, float64:`

```go
package main

func foo[T int64 | float64](v T) {
	// do something with v
	_ = v
}

func main() {
	var a any
	switch v := a.(type) {
	case int64, float64:
		foo(v) // INVALID: "any does not satisfy int64 | float64"
	default:
		panic("not happy")
	}
}

```

denisvmedia (approved):

denisvmedia:
Agree to all the points. Thank you @chavacava !

denisvmedia:
@ccoVeille but this will work:

```
func main() {
	var a any
	switch a.(type) {
	case int64, float64:
		foo("VALID: "does satisfy int64 | float64")
	default:
		panic("not happy")
	}
}
```

ccoVeille:
Yes, I was just spotting the edge case

PR #1458, opened 2025-08-02
unexported-return: do not allow unexported interfaces

### LINKED ISSUE #1457: unexported-return: unexported interfaces should not be allowed
Currently, `unexported-return` rule allows returning unexported interface types. E.g.

```go
type unexportedInterface interface {
        unexportedMethod()
}

func New() unexportedInterface {
        return nil
}
```

As per discussion in https://github.com/mgechev/revive/pull/1456#issuecomment-3146371336, this is unfriendly to the user of the package. The example above is also “annoying to use“ because the interface type includes an unexported method, so it cannot be assigned to a value that is a subset of the original interface in different package. E.g. https://go.dev/play/p/ohypfiXhADR

**Expected behavior**

The `unexported-return` rule should not allow named unexported return types (except when defined as an exported type alias).

tie:
The current implementation of `unexported-return` rule allows returning unexported interface types.

```go
type unexportedInterface any

func New() unexportedInterface {
        return nil
}
```

Closes #1457

### DIFF
--- a/testdata/golint/unexported_return.go
+++ b/testdata/golint/unexported_return.go
@@ -59,6 +59,15 @@ func ExportedIntReturner() int { // MATCH /exported func ExportedIntReturner ret
 	return int{}
 }
 
+type unexportedInterface interface {
+	UnexportedInterface()
+}
+
+// UnexportedInterfacePointer returns an unexported interface pointer type from this package.
+func UnexportedInterfacePointer() *unexportedInterface { // MATCH /exported func UnexportedInterfacePointer returns unexported type *foo.unexportedInterface, which can be annoying to use/
+	return nil
+}
+
 type config struct {
 	N int
 }
--- a/rule/unexported_return.go
+++ b/rule/unexported_return.go
@@ -80,27 +80,9 @@ func (*UnexportedReturnRule) Name() string {
 func exportedType(typ types.Type) bool {
 	switch t := typ.(type) {
 	case *types.Alias:
-		obj := t.Obj()
-		switch {
-		// Builtin types have no package.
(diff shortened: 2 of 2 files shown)


testdata/golint/unexported_return.go
```diff
@@ -59,6 +59,15 @@ func ExportedIntReturner() int { // MATCH /exported func ExportedIntReturner ret
 	return int{}
 }
 
+type unexportedInterface interface {
+	UnexportedInterface()
+}
+
+// UnexportedInterfacePointer returns an unexported interface pointer type from this package.
+func UnexportedInterfacePointer() *unexportedInterface { // MATCH /exported func UnexportedInterfacePointer returns unexported type *foo.unexportedInterface, which can be annoying to use/
```
veeam-denis: Note, having pointer for an interface type as a result in 99.99% of cases is a mistake. Although, we can check it, I'd focus on a non-pointer result.

ccoVeille: Should we have a rule for detecting this? Or is it something already detected?

I have no computer with me to check unfortunately

chavacava: Returning an interface (vs concrete type) is usually not the best option (unless the func is a struct _factory_ or something like that)

veeam-denis: @chavacava I agree, but here we have a pointer to an interface, which is even worse :) 

I can imagine returning a settable value of an interface type and for that we return a pointer to it, but that's something I personally never saw in real life (neither I had a reason to do that in my own code). I saw this for input arguments that were meant to be the output ones as well, but not for the returned value.

PR #1473, opened 2025-08-13
refactor: reduces the number of references to deprecated *ast.Object

chavacava:
Micro PR that declares the deprecated `*ast.Object` into a private type to "encapsulate" the deprecated type.

### DIFF
--- a/rule/datarace.go
+++ b/rule/datarace.go
@@ -8,6 +8,9 @@ import (
 	"github.com/mgechev/revive/lint"
 )
 
+//nolint:staticcheck // TODO: ast.Object is deprecated
+type nodeUID *ast.Object // type of the unique id for AST nodes
+
 // DataRaceRule lints assignments to value method-receivers.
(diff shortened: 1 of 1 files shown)


ccoVeille:
What are the next step after this PR?

Do you plan a refactoring?

chavacava:
> What are the next step after this PR?
> Do you plan a refactoring?

Rules use `*ast.Object` as an unique identifier of `ast.Ident` nodes. AFAIK there is no other way to uniquely identify identifiers. "uniquely" with some exceptions as described in the deprecation documentation even if these exceptions are very rare for the use case of the rule (negligible functional impact to the rule)

To remove the dependence on `*ast.Object`, the rule will need to use type information and then refactor it to use information on _Uses_. That will be a future PR. As a side effect, the refactoring might help to improve the runtime performance of the rule (the one with the worst performance of all rules)

alexandear (approved):

PR #1483, opened 2025-08-16
feature: detect useless options in spanner struct tag

chavacava:
In spanner struct tags, `-` is used to ignore the field therefore any option following `-` is useless.

This PR evolves the spanner struct tag checker to spot useless options attached to ignored fields.

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -562,13 +562,18 @@ func checkYAMLTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (messa
 
 func checkSpannerTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (message string, succeeded bool) {
 	if tag.Name == "-" {
+		if len(tag.Options) > 0 {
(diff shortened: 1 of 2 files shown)


rule/struct_tag.go
```diff
@@ -562,13 +562,18 @@ func checkYAMLTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (messa
 
 func checkSpannerTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (message string, succeeded bool) {
 	if tag.Name == "-" {
+		if len(tag.Options) > 0 {
+			return fmt.Sprintf("useless option(s) %s for ignored field", strings.Join(tag.Options, ",")), false
+		}
 		return "", true
 	}
```
ccoVeille, line 569: Isn't the same for json and yaml tag then ? And maybe all of them ?

I mean when the value is - what could we expect from the serializer when it has option ?

ccoVeille (approved):

rule/struct_tag.go
chavacava: Yes, I will extend the check to those cases in a future PR

PR #1485, opened 2025-08-16
fix: default formatter generates an extra new line

chavacava:
The default formatter generates an extra empty line at the end of the failure list

image

This PR removes the artifact.

image

### DIFF
--- a/formatter/default.go
+++ b/formatter/default.go
@@ -21,8 +21,14 @@ func (*Default) Name() string {
 // Format formats the failures gotten from the lint.
 func (*Default) Format(failures <-chan lint.Failure, _ lint.Config) (string, error) {
 	var buf bytes.Buffer
+	first := true
+	prefix := ""
 	for failure := range failures {
-		fmt.Fprintf(&buf, "%v: %s\n", failure.Position.Start, failure.Failure)
+		fmt.Fprintf(&buf, "%s%v: %s", prefix, failure.Position.Start, failure.Failure)
+		if first {
+			first = false
+			prefix = "\n"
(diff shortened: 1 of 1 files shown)


formatter/default.go
```diff
@@ -21,8 +21,14 @@ func (*Default) Name() string {
...
 	var buf bytes.Buffer
+	first := true
+	prefix := ""
 	for failure := range failures {
-		fmt.Fprintf(&buf, "%v: %s\n", failure.Position.Start, failure.Failure)
+		fmt.Fprintf(&buf, "%s%v: %s", prefix, failure.Position.Start, failure.Failure)
+		if first {
+			first = false
+			prefix = "\n"
+		}
```
ccoVeille:
```suggestion
		// next line will have a new-line prefix
		prefix = "\n"
```

formatter/default.go
```diff
@@ -21,8 +21,14 @@ func (*Default) Name() string {
 // Format formats the failures gotten from the lint.
 func (*Default) Format(failures <-chan lint.Failure, _ lint.Config) (string, error) {
 	var buf bytes.Buffer
+	first := true
+	prefix := ""
```
ccoVeille:
```suggestion
	// first line has no prefix
	prefix := ""
```

ccoVeille (commented):
I feel this would be simpler

formatter/default.go
chavacava: 20% less CPU performant but simpler. Why not.

ccoVeille (approved):

PR #1487, opened 2025-08-17
feature: struct-tag warns on (useless) options on ignored fields

chavacava:
This PR completes #1483 by detecting useless options on other known struct tags.

For some tags, using `-` to indicate "ignore field" is not formally defined nor documented but open source code bases use it.

### DIFF
--- a/testdata/go1.24/struct_tag.go
+++ b/testdata/go1.24/struct_tag.go
@@ -21,7 +21,7 @@ type decodeAndValidateRequest struct {
 	OptionalQuery    string          `json:"-" querystring:"queryfoo"`
 	optionalQuery    string          `json:"-" querystring:"queryfoo"` // MATCH /tag on not-exported field optionalQuery/
 	// No-reg test for bug https://github.com/mgechev/revive/issues/208
-	Tiret       string `json:"-,"`                   // MATCH /useless empty option for ignored field (remove the comma after -) in json tag/
+	Tiret       string `json:"-,"`
 	BadTiret    string `json:"other,"`               // MATCH /option can not be empty in json tag/
 	ForOmitzero string `json:"forOmitZero,omitzero"` // Go 1.24 introduces omitzero
 }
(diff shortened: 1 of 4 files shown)


chavacava:
> There is one problem `json:"-,-"` is the valid syntax for a JSON field named `-`
> 
> It's the way to support `-` as field name

Zero occurrence [in the wild](https://github.com/search?q=%2F%60json%3A%22-%2C-%2F+language%3AGo&type=code&ref=advsearch). Maybe because using `-` as field name is an heresy.

ccoVeille:
My bad, the syntax is `json:"-,"`

https://github.com/search?q=%2F%60json%3A%22-%2C%22%2F+language%3AGo&type=code&ref=advsearch

And it's used.

`json:"-,-"` is invalid as `-` is not an option

chavacava:
> My bad, the syntax is `json:"-,"`
> 
> https://github.com/search?q=%2F%60json%3A%22-%2C%22%2F+language%3AGo&type=code&ref=advsearch
> 
> And it's used.
> 
> `json:"-,-"` is invalid as `-` is not an option

Then the playground code you linked has also a typo ?
image

ccoVeille:
Yes

Copy pasting is the root of evil 😁

https://go.dev/play/p/xd6JVO2XFf3

alexandear (approved):

testdata/go1.24/struct_tag.go
```diff
@@ -21,7 +21,7 @@ type decodeAndValidateRequest struct {
 	OptionalQuery    string          `json:"-" querystring:"queryfoo"`
 	optionalQuery    string          `json:"-" querystring:"queryfoo"` // MATCH /tag on not-exported field optionalQuery/
 	// No-reg test for bug https://github.com/mgechev/revive/issues/208
-	Tiret       string `json:"-,"`
+	Tiret       string `json:"-,"`                   // MATCH /useless empty option for ignored field (remove the comma after -) in json tag/
```
ccoVeille: This one is valid, it shouldn't be reported

But this won't

```
`json:"-,omitempty"`
`json:"-,-"
```

There are legitimate uses case for using - as a field

https://github.com/search?q=language%3Ajson+%2F%22-%22%3A%2F&type=code

Among them, things around characters matching

ccoVeille:
Thanks @chavacava

PR #1495, opened 2025-08-28
refactor: remove rule enforce-else

chavacava:
`enforce-else` generates too many failures (for example, +1800 failures in the Kubernetes code base)

Even if the rule enforces explicit completeness of conditionals, it evidently doesn't fit with the _Go way_ of doing things.
 
The rule was never released thus we can remove it without backward compatibility issues.


alexandear (approved):
Please update the description of PR where this rule has been added #1436.

Issue #1477, opened 2025-08-13
struct-tag: add support for `codec` tag


ccoVeille:
https://github.com/ugorji/go/codec

This tag can be found in multiple projects

https://github.com/search?q=language%3Ago+%2F%60codec%3A%22%2F&type=code

pashkov256:
Hi @ccoVeille , I'd love to tackle this as a first-time contributor!

chavacava:
> Hi [@ccoVeille](https://github.com/ccoVeille) , I'd love to tackle this as a first-time contributor!

It's all yours. Thanks @pashkov256

chavacava:
Reading [the official doc](https://ugorji.net/blog/go-codec-primer)(?) I understand that `_struct struct{}` is an special field of the struct to provide common options to all other fields.

The example provided in the doc is the following
```go
type My struct {
    _struct struct{} `codec:",omitempty"`   //set omitempty for every field
    Field1 string    `codec:"-"`            //skip this field
    Field2 int       `codec:"myName"`       //Use key "myName" in encode stream
    Field3 int32     `codec:",omitempty"`   //use key "Field3". Omit if empty.
    Field4 bool      `codec:"f4,omitempty"` //use key "f4". Omit if empty.
    field5 bool                             // unexported, so skipped
    Anon                                    // anonymous field, S is inlined.
                                            // stream will contain S as if a regular field.
}
```

I think we could add the following checks:
1. tags for `_struct struct{}` should only have options.
2. if `_struct struct{}` tag sets option _X_ then setting _X_ on other fields is redundant

(please notice that in the above example, both `Field3` and `Field4` have a tag setting `omitempty` even if the tag of `_struct` field already set it. Is a redundancy in the example? Or my understanding of tags for `_struct` is wrong ?)

chavacava:
The option `omitemptyarray` even not official(?) is used in many code bases. 
See 
https://github.com/search?q=language%3Ago+%2F%60codec%3A%22.*omitemptyarray%2F&type=code

How should the rule handle `omitemptyarray`? As an error? Accept it?

chavacava:
~`-` is used to skip a field thus a tag with `-` should not have options.~
~For example, the rule could spot tags like `codec;"-,omitempty"`  and warn the user.~

PS: This is already implemented in the rule by #1487

ccoVeille:
Thanks for checking and providing valuable pointers

pashkov256:
[...]

PR #1507, opened 2025-08-31
struct-tag: support codec struct tag

pashkov256:
This PR adds support for the codec struct tag to the struct-tag rule, as requested in https://github.com/mgechev/revive/issues/1477. The codec tag is widely used in Go projects for serialization/deserialization (e.g., with github.com/ugorji/go/codec), and its absence in the linter could lead to false positives or missed issues.

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -39,7 +39,7 @@ const (
 	keyYAML         tagKey = "yaml"
 )
 
-type tagChecker func(checkCtx *checkContext, tag *structtag.Tag, fields []*ast.Ident, fieldType ast.Expr) (message string, succeeded bool)
+type tagChecker func(checkCtx *checkContext, tag *structtag.Tag, field *ast.Field) (message string, succeeded bool)
 
 var tagCheckers = map[tagKey]tagChecker{
 	keyASN1:         checkASN1Tag,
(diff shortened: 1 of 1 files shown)


ccoVeille:
Please make sure the following points are addressed

- https://github.com/mgechev/revive/issues/1477#issuecomment-3191493076

- https://github.com/mgechev/revive/issues/1477#issuecomment-3191508409

chavacava:
Hi @pashkov256, I made a few changes to get the tests passing and to cover what was missing. Please check out and let me know if it looks good to you.

cc: @ccoVeille

rule/struct_tag.go
```diff
@@ -74,6 +77,23 @@ func (checkCtx checkContext) isUserDefined(key tagKey, opt string) bool {
...
+		return false
+	}
+
+	_, ok := checkCtx.commonOptions[opt]
+	return ok
+}
+
+func (checkCtx *checkContext) addCommonOption(opt string) {
+	if checkCtx.commonOptions == nil {
+		checkCtx.commonOptions = map[string]bool{}
```
ccoVeille, line 91: Isn't a classic pattern to use this instead?


```suggestion
		checkCtx.commonOptions = make(map[string]bool)
```

rule/struct_tag.go
```diff
@@ -165,7 +185,8 @@ func (w lintStructTagRule) Visit(node ast.Node) ast.Visitor {
 // checkTaggedField checks the tag of the given field.
 // precondition: the field has a tag
 func (w lintStructTagRule) checkTaggedField(checkCtx *checkContext, f *ast.Field) {
-	if len(f.Names) > 0 && !f.Names[0].IsExported() {
+	//                                                 do not warn on fields prefixed by _ (used by some tags, like codec)
+	if len(f.Names) > 0 && !f.Names[0].IsExported() && !strings.HasPrefix(f.Names[0].Name, "_") {
 		w.addFailuref(f, "tag on not-exported field %s", f.Names[0].Name)
 	}
```
ccoVeille: Is there a way for us to add this check only for codec tag ?

I feel like it creates a loop especially because now everything prefixed with _ will pass

I feel like only _struct should pass and only for codec

I mean there are a lot of there

https://github.com/search?q=language%3Ago+%2F_%5Ba-z%5D%2B%5Cs%5C*%3F%5Ba-z%5D%2B%5Cs%2B%60%5Ba-z%5D%2B%3A%22%2F&type=code

rule/struct_tag.go
chavacava: cf https://github.com/mgechev/revive/pull/1515#discussion_r2328762902

rule/struct_tag.go
chavacava: You are right. It will require some refactoring. I'll try.

> I mean there are a lot of there

Not so much. The regexp in the search catches things like `My_field` and those things are a huge part of the results.

chavacava: Done

rule/struct_tag.go
```diff
@@ -184,16 +201,49 @@ func (w lintStructTagRule) checkTaggedField(checkCtx *checkContext, f *ast.Field
 			w.addFailureWithTagKey(f.Tag, msg, tag.Key)
 		}
 
-		checker, ok := w.tagCheckers[tagKey(tag.Key)]
+		tagKey := tagKey(tag.Key)
```
ccoVeille: tagKey variable shouldn't shadow its type

rule/struct_tag.go
```diff
@@ -184,16 +201,49 @@ func (w lintStructTagRule) checkTaggedField(checkCtx *checkContext, f *ast.Field
...
+
+	if w.shallWarnOnUnexportedField(f.Names, analyzedTags) {
+		w.addFailuref(f, "tag on not-exported field %s", f.Names[0].Name)
+	}
+}
+
+// tagKeyToSpecialField maps tag keys to their "special" meaning struct fields.
+var tagKeyToSpecialField = map[tagKey]string{
+	"codec": "_struct",
+}
```
ccoVeille:
```suggestion
// specialKeyCodec is the name of the special field for codec tag
const specialFieldCodec = "_struct"

// tagKeyToSpecialField maps tag keys to their "special" meaning struct fields.
var tagKeyToSpecialField = map[tagKey]string{
	keyCodec: specialFieldCodec,
}
```

rule/struct_tag.go
```diff
@@ -38,11 +39,12 @@ const (
 	keyYAML         tagKey = "yaml"
 )
 
-type tagChecker func(checkCtx *checkContext, tag *structtag.Tag, fieldType ast.Expr) (message string, succeeded bool)
+type tagChecker func(checkCtx *checkContext, tag *structtag.Tag, fields []*ast.Ident, fieldType ast.Expr) (message string, succeeded bool)
```
ccoVeille: I would refactor to pass a struct to tagChecker

There are too many fields to make it readable now.

Especially when most checkers ignore 2 arguments

rule/struct_tag.go
```diff
@@ -320,7 +370,29 @@ func checkBSONTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (messa
 	return "", true
 }
 
-func checkJSONTag(checkCtx *checkContext, tag *structtag.Tag, _ ast.Expr) (message string, succeeded bool) {
+func checkCodecTag(checkCtx *checkContext, tag *structtag.Tag, fieldNames []*ast.Ident, _ ast.Expr) (message string, succeeded bool) {
+	mustAddToCommonOptions := len(fieldNames) == 1 && fieldNames[0].Name == "_struct" // see https://github.com/mgechev/revive/issues/1477#issuecomment-3191493076
```
ccoVeille: See https://github.com/mgechev/revive/pull/1507#discussion_r2328804380

```suggestion
	mustAddToCommonOptions := len(fieldNames) == 1 && fieldNames[0].Name == specialFieldCodec // see https://github.com/mgechev/revive/issues/1477#issuecomment-3191493076
```

rule/struct_tag.go
ccoVeille: See also

https://github.com/mgechev/revive/pull/1507#discussion_r2328805091

chavacava:
Thanks @pashkov256 for the PR and @ccoVeille for the feedback.

ccoVeille:
Does this rule find bugs in Go modules known for their high number of issues ?

chavacava:
Do you mean the whole rule struct-tag or the codec tag check?

Because struct tags are very error-prone (mere strings for the compiler) the rule finds many problems in code bases. Mainly bad formatted tags, misspelled options and duplicated id (very common in protobuf tags)

ccoVeille:
i meant the codec tag check that this PR adds.

I was curious if it reported a lot of things.

chavacava:
> i meant the codec tag check that this PR adds.
> 
> I was curious if it reported a lot of things.

No idea. I didn't tested it on open source code bases

ccoVeille:
OK, let's hope we won't report false-positive issues then.

@alexandear or @pashkov256 did you make some tests on your own?

pashkov256:
@ccoVeille I didn't do

PR #1525, opened 2025-09-20
fix: rule `struct-tag` false positive on `validate:omitempty`

### LINKED ISSUE #1524: rule.struct-tag false positive on `validate:omitempty`
**Describe the bug**
revive does not seem to recognize that the 'omitempty' tag is valid in 'validate' tags

**To Reproduce**

```shell
# flags

revive -formatter=stylish -config=revive.toml ./...`
```

```toml
# config file
ignoreGeneratedHeader = false
severity = "error"
confidence = 0.8
errorCode = 0
warningCode = 0

[directive.specify-disable-reason]
[rule.add-constant]
    arguments = [{maxLitCount = "3",allowStrs ="\"\",\"sfn.DescribeExecution: %w\",",allowInts="0,1,2,3",allowFloats="0.0"}]
[rule.argument-limit]
    arguments =[5]
[rule.atomic]
[rule.blank-imports]
[rule.bool-literal-in-expr]
[rule.call-to-gc]
[rule.confusing-naming]
    severity = "warning"
[rule.confusing-results]
    severity = "warning"
[rule.constant-logical-expr]
[rule.context-as-argument]
[rule.context-keys-type]
[rule.cyclomatic]
    arguments =[9]
    severity = "warning"
[rule.deep-exit]
[rule.duplicated-imports]
[rule.empty-block]
    severity = "warning"
[rule.empty-lines]
[rule.error-naming]
[rule.error-return]
[rule.error-strings]
[rule.errorf]
[rule.exported]
[rule.flag-parameter]
[rule.function-result-limit]
    arguments = [3]
[rule.get-return]
[rule.if-return]
[rule.increment-decrement]
[rule.indent-error-flow]
[rule.imports-blacklist]
    arguments =["http://github.com/pkg/errors"]
    severity = "warning"
[rule.import-shadowing]
[rule.package-comments]
    severity = "warning"
[rule.range]
[rule.range-val-in-closure]
[rule.receiver-naming]
[rule.redefines-builtin-id]
[rule.struct-tag]
  arguments = ["json,inline"]
[rule.superfluous-else]
[rule.time-naming]
[rule.var-naming]
    arguments = [["ID"]]
[rule.var-declaration]
[rule.unexported-return]
[rule.unhandled-error]
    arguments = ["fmt.Println","fmt.Printf"]
[rule.unnecessary-stmt]
[rule.unreachable-code]
[rule.unused-parameter]
[rule.unused-receiver]
[rule.waitgroup-by-value]
```
I have tried adding `validate` to the arguments under rule.struct-tag but got the same undesired result

** Actual Behavior **
revive complains about the presence of the `omitempty` value in `validate` tags everywhere it appears, and this did not happen before

**Expected behavior**
using `omitempty` in combination with other validate tags is valid, but on its own revive should probably question it unless the type of the property is a pointer to a custom struct that has its own validate tags

**Logs**

 - OS: [Mac Tahoe 26]
 - Version of Go 1.25.1

**Additional context**

chavacava:
Closes #1524 by adding `omitempty` to the list of accepted options for `validate`

### DIFF
--- a/rule/struct_tag.go
+++ b/rule/struct_tag.go
@@ -843,6 +843,7 @@ var validateSingleOptions = map[string]struct{}{
 	"multibyte":                     {},
 	"number":                        {},
 	"numeric":                       {},
+	"omitempty":                     {},
 	"port":                          {},
 	"postcode_iso3166_alpha2":       {},
 	"postcode_iso3166_alpha2_field": {},
--- a/testdata/struct_tag.go
+++ b/testdata/struct_tag.go
@@ -156,6 +156,7 @@ type MapStruct struct {
 }
 
 type ValidateUser struct {
+	Id          string `validate:"omitempty,min=3,max=32"`
 	Username    string `validate:"required,min=3,max=32"`
 	Email       string `validate:"required,email"`
 	Password    string `validate:"required,min=8,max=32"`


rule/struct_tag.go
```diff
@@ -843,6 +843,7 @@ var validateSingleOptions = map[string]struct{}{
 	"multibyte":                     {},
 	"number":                        {},
 	"numeric":                       {},
+	"omitempty":                     {},
```
ccoVeille, line 846: I feel like validate supports way more validation tags that the one listed here

https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Using_Validator_Tags

I can see at list `omitnil` and `omitzero` are missing.

The complete list is there apparently

https://github.com/go-playground/validator/blob/v10.27.0/baked_in.go

I checked at recent changes in validator I found this recent changes

[omitzero](https://github.com/go-playground/validator/commit/f5f02dcb442ff50ac385d5963e9a7b75b01b6544)
 [validateFn](https://github.com/go-playground/validator/commit/8032f400f4942bd2376b759570abab3dfcb73bbb)
 [ein](https://github.com/go-playground/validator/commit/79c42ad73c1fdebee04c93ac6a0e132353662dc6)

[oneofci](https://github.com/go-playground/validator/commit/d7abf3244e3d4f860c36def311aad53f8e85a45a) 

There might be way more

PR #1533, opened 2025-09-25
fix: rule `struct-tag` add all missing go-validator tags (#1526)

### LINKED ISSUE #1526: `struct-tag`: complete the supported validation tags of `validate`
I feel like validate supports way more validation tags that the one listed here

https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Using_Validator_Tags

I can see at list `omitnil` and `omitzero` are missing.

The complete list is there apparently

https://github.com/go-playground/validator/blob/v10.27.0/baked_in.go

I checked at recent changes in validator I found this recent changes

[omitzero](https://github.com/go-playground/validator/commit/f5f02dcb442ff50ac385d5963e9a7b75b01b6544)
 [validateFn](https://github.com/go-playground/validator/commit/8032f400f4942bd2376b759570abab3dfcb73bbb)
 [ein](https://github.com/go-playground/validator/commit/79c42ad73c1fdebee04c93ac6a0e132353662dc6)

[oneofci](https://github.com/go-playground/validator/commit/d7abf3244e3d4f860c36def311aad53f8e85a45a) 

There might be way more

_Originally posted by @ccoVeille in https://github.com/mgechev/revive/pull/1525#discussion_r2365359777_

focusaurus:
Updating the list of known struct tags from `go-validator`. I came here mostly for `omitnil` but for this PR I grabbed the list from [baked_in.go](https://github.com/go-playground/validator/blob/bc77d03bfbd951ad1299267754e230b0af36a5aa/baked_in.go#L79) at the moment and added them all, sorted, and removed dupes. The sorting shifted a few around but these were already mostly sorted.

I added a test cases for `omitnil` and `omitzero` as I presume that these are widely used, but we don't currently exhaustively test the whole list of these and that seems fine so left a lot of them without a specific test.

Closes #1526


test/struct_tag_test.go
```diff
@@ -20,6 +20,8 @@ func TestStructTagWithUserOptions(t *testing.T) {
 			"datastore,myDatastoreOption",
 			"mapstructure,myMapstructureOption",
 			"validate,displayName",
+			"validate,omitnil",
```
chavacava: The configuration could be written

```suggestion
			"validate,omitnil,omitzero",
```

Anyway, declaring these as user-defined options should not be necessary because the PR add both to the the list of known options for validate.

focusaurus: Ah. I might not have understood what this test was about. Adjusting.

ccoVeille:
Thank you @focusaurus

PR #1536, opened 2025-10-05
fix: rule `confusing-naming` false positive on multiple blank identifiers

### LINKED ISSUE #1535: False positive "differs only by captalization" on alignment fields in struct
**Describe the bug**

A struct with more than one `_` field (e.g. for alignment) is reported to have fields differing only by capitalization.

**To Reproduce**

```shell
% cat a.go 
// package main is a package
package main

type siginfoChild struct {
	signo    int32
	errno    int32
	exitCode int32
	_        int32

	pid    int32
	uid    uint32
	status int32

	_ [25]uint32 // Pad to 128 bytes
}

% go run github.com/mgechev/revive@HEAD -config config.toml a.go
a.go:14:2: Field '_' differs only by capitalization to other field in the struct type siginfoChild
```

```toml
enableAllRules = true
```

**Expected behavior**

A struct may have more than one `_` field.

`go version go1.25.1 darwin/arm64`

emmanuel-ferdman:
## PR Summary
The `confusing-naming` rule was incorrectly reporting false positives when Go structs contained multiple blank identifier (`_`) fields, which are commonly used for padding and alignment. The rule's `checkStructFields` function was treating all field names equally, including blank identifiers, causing it to flag the second occurrence of `_` as differing only by capitalization from the first. This fix adds a simple check to skip blank identifiers before performing the capitalization comparison, as blank identifiers are special in Go and explicitly allowed to appear multiple times in the same struct according to the language specification.

Example:
```go
// test.go
package main

type siginfoChild struct {
    signo    int32
    errno    int32
    exitCode int32
    _        int32

    pid    int32
    uid    uint32
    status int32

    _ [25]uint32 // Pad to 128 bytes
}
```
Old output:
> Field '_' differs only by capitalization to other field in the struct type siginfoChild

Fixes #1535.

### DIFF
--- a/rule/confusing_naming.go
+++ b/rule/confusing_naming.go
@@ -171,7 +171,7 @@ func checkStructFields(fields *ast.FieldList, structName string, w *lintConfusin
 			if id.Name == "_" {
 				continue
 			}
-			
+
 			normName := strings.ToUpper(id.Name)
 			if bl[normName] {
 				w.onFailure(lint.Failure{
--- a/rule/confusing_naming.go
+++ b/rule/confusing_naming.go
@@ -171,6 +171,7 @@ func checkStructFields(fields *ast.FieldList, structName string, w *lintConfusin
 			if id.Name == "_" {
 				continue
 			}
+			
 			normName := strings.ToUpper(id.Name)
 			if bl[normName] {
 				w.onFailure(lint.Failure{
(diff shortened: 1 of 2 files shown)


chavacava:
Hi @emmanuel-ferdman, thanks for the PR.

PR #1544, opened 2025-10-09
deep-exit: detect exit-triggering flag usage

### LINKED ISSUE #1523: deep-exit: add Go standard library flag.Parse call to default exclusions
https://pkg.go.dev/flag#Parse

flag.CommandLine is configured to ExitOnError

https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/flag/flag.go;l=1201

And flag.Flagset.Parse would call os.Exit on error or with --help flag

https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/flag/flag.go;l=1153

uloamaka:
This PR resolves #1523, 
Packages using `flag.ExitOnError` or `flag.Parse()` can exit on invalid flags or `--help`, which violates the principle of reusability the deep-exit rule enforces.

This PR enhances the deep-exit rule to detect indirect program termination caused by the Go flag package.

Specifically, it adds checks for:
- `flag.NewFlagSet` created with `flag.ExitOnError`
- Calls to `flag.Parse()` outside of `main()` or `init()`

These usages can terminate programs unexpectedly, reducing reusability of packages that expose such functions.

Updates:
- testdata/deep_exit.go

* [x]  I added tests

### DIFF
--- a/rule/utils.go
+++ b/rule/utils.go
@@ -17,7 +17,7 @@ type exitFuncChecker func(args []ast.Expr) bool
 var alwaysTrue exitFuncChecker = func([]ast.Expr) bool { return true }
 
 // exitFunctions is a map of std packages and functions that are considered as exit functions.
-var exitFunctions = map[string]map[string]exitFuncChecker {
+var exitFunctions = map[string]map[string]exitFuncChecker{
 	"os":      {"Exit": alwaysTrue},
 	"syscall": {"Exit": alwaysTrue},
 	"log": {


uloamaka:
@alexandear Can you please help review this PR.

rule/utils.go
```diff
@@ -10,16 +11,33 @@ import (
 )
 
 // exitFunctions is a map of std packages and functions that are considered as exit functions.
-var exitFunctions = map[string]map[string]bool{
-	"os":      {"Exit": true},
-	"syscall": {"Exit": true},
+var exitFunctions = map[string]map[string]func(args []ast.Expr) bool{
```
chavacava: `func(args []ast.Expr) bool` could be a type

something like
```go
type exitFuncChecker func(args []ast.Expr) bool`
```

rule/utils.go
```diff
@@ -10,16 +11,33 @@ import (
...
+	"flag": {
+		"Parse": func([]ast.Expr) bool { return true },
+		"NewFlagSet": func(args []ast.Expr) bool {
+			if len(args) != 2 {
+				return false
+			}
+			arg, ok := args[1].(*ast.SelectorExpr)
+			if !ok {
+				return false
+			}
```
chavacava:
```suggestion
			}
			
```

rule/utils.go
```diff
@@ -10,16 +11,33 @@ import (
...
+			id, ok := arg.X.(*ast.Ident)
+			if !ok {
+				return false
+			}
```
chavacava:
```suggestion
			}
			
```

rule/utils.go
```diff
@@ -10,16 +11,33 @@ import (
 )
 
 // exitFunctions is a map of std packages and functions that are considered as exit functions.
-var exitFunctions = map[string]map[string]bool{
-	"os":      {"Exit": true},
-	"syscall": {"Exit": true},
+var exitFunctions = map[string]map[string]func(args []ast.Expr) bool{
+	"os":      {"Exit": func([]ast.Expr) bool { return true }},
```
chavacava: we could create a variable for the expression `func([]ast.Expr) bool { return true }` to avoid defining it multiple times.

rule/utils.go
```diff
@@ -10,16 +11,33 @@ import (
...
+			}
+			arg, ok := args[1].(*ast.SelectorExpr)
+			if !ok {
+				return false
+			}
+			id, ok := arg.X.(*ast.Ident)
+			if !ok {
+				return false
+			}
+			return id.Name == "flag" && arg.Sel.Name == "ExitOnError"
```
chavacava: We could write


```suggestion
			return astutils.IsPkgDotName(args[1], "flag", "ExitOnError")
```

rule/utils.go
```diff
@@ -86,8 +104,16 @@ func isDirectiveComment(line string) bool {
 }
 
 // isCallToExitFunction checks if the function call is a call to an exit function.
-func isCallToExitFunction(pkgName, functionName string) bool {
-	return exitFunctions[pkgName] != nil && exitFunctions[pkgName][functionName]
+func isCallToExitFunction(pkgName, functionName string, ce []ast.Expr) bool {
+	m, ok := exitFunctions[pkgName]
+	if !ok {
+		return false
+	}
```
chavacava, line 109:
```suggestion
	}
	
```

rule/utils.go
```diff
@@ -86,8 +104,16 @@ func isDirectiveComment(line string) bool {
...
+	check, ok := m[functionName]
+	if !ok {
+		return false
+	}
```
chavacava, line 114:
```suggestion
	}
	
```

rule/utils.go
```diff
@@ -86,8 +104,16 @@ func isDirectiveComment(line string) bool {
 }
 
 // isCallToExitFunction checks if the function call is a call to an exit function.
-func isCallToExitFunction(pkgName, functionName string) bool {
-	return exitFunctions[pkgName] != nil && exitFunctions[pkgName][functionName]
+func isCallToExitFunction(pkgName, functionName string, ce []ast.Expr) bool {
```
chavacava: we should find a better name for `ce` argument

`callArgs` ?

cli/main.go
```diff
@@ -173,7 +173,7 @@ func initConfig() {
 	flag.BoolVar(&versionFlag, "version", false, versionUsage)
 	flag.BoolVar(&setExitStatus, "set_exit_status", false, exitStatusUsage)
 	flag.IntVar(&maxOpenFiles, "max_open_files", 0, maxOpenFilesUsage)
-	flag.Parse()
+	flag.Parse() //revive:disable-line:deep-exit
```
chavacava, line 176: The fact we disable the rule in our own code makes me think on the pertinence of checking calls to `flag.Parse` outside the `main` function.

testdata/deep_exit.go
```diff
@@ -36,3 +37,15 @@ func TestMain(m *testing.M) {
 	// must match because this is not a test file
 	os.Exit(m.Run()) // MATCH /calls to os.Exit only in main() or init() functions/
 }
+
+func flagParseOutsideMain() {
+	flag.Parse() // MATCH /calls to flag.Parse only in main() or init() functions/
+}
+
+func flagNewFlagSetExitOnErrorOutsideMain() {
+	flag.NewFlagSet("cmd", flag.ExitOnError) // MATCH /calls to flag.NewFlagSet only in main() or init() functions/
```
chavacava: As I understand from the rule implementation and test, not all calls to `flag.NewFlagSet` are problematic. Only those that have `flag.ExitOnError` as argument are. If my understanding is correct then the failure message is not precise enough.

cli/main.go
ccoVeille: That's because the right way to do is to use either this

```go
func main() {

	// (...)
	flag.BoolVar(&setExitStatus, "set_exit_status", false, exitStatusUsage)
	flag.IntVar(&maxOpenFiles, "max_open_files", 0, maxOpenFilesUsage)
	flag.Parse()

	// (...)
}
```

or this

```go
func main() {
	err := run()
	// (...)
}

func run() error {
	fl := flag.NewFlagSet(os.Args[0], flag.ContinueOnError)
	fl.BoolVar(&setExitStatus, "set_exit_status", false, exitStatusUsage)
	fl.IntVar(&maxOpenFiles, "max_open_files", 0, maxOpenFilesUsage)
	err := fl.Parse(os.Args[1:])
	if err != nil {
		return err
	}
	// (...)

	return nil
}
```

You can take a look at the [code](https://github.com/ccoVeille/gh-reaction/blob/main/main.go) I wrote for [gh-reaction](https://github.com/ccoVeille/gh-reaction)

But maybe, the error message for flag.Parse should be way clearer, about the fact there is a need to use `flag.ContinueOnError`

Please note also, that the same code, using `flag.NewFlagSet(os.Args[0], flag.ExitOnError)` or `flag.NewFlagSet(os.Args[0], flag.PanicOnError)` should be reported as deep-exit, but it would be a next move, and somehow a niche maybe

EDIT: it's not a niche :P https://github.com/search?q=language%3Ago%20%2Fflag%5C.NewFlagSet.*flag%5C.(Exit%7CPanic)OnError%2F%20-is%3Afork%20-is%3Aarchived&type=code

chavacava (approved):

chavacava:
Thank you @uloamaka for the PR and @alexandear @ccoVeille for the feedback

Issue #1538, opened 2025-10-06
New "useless-fallthrough" rule conflicts with "exhaustive" from https://github.com/nishanths/exhaustive


macrael:
**Describe the bug**
https://github.com/nishanths/exhaustive is a commonly used linter than ensures when switching on enum-like constants that every option is included in a switch statement. 

revive has a rule that requires a default case. 

unfortunately golang doesn't actually support exhaustive enums, so often a `default:` case is required even though you have included all possible cases in the list. ALSO you cannot combine the `default` with a `case:`

The new rule alerts when you have a case that has nothing but "fallthrough" in it. 

**To Reproduce**
```
type SomeType int

const (
	One SomeType = iota
	Two
	Three
	Unknown
)

func CheckParsedVal(SomeType i) int {
	switch i {
	case One:
		return 1
	case Two:
		return 2
	case Three:
		return 3
	case Unknown:
		fallthrough
	default:
		return -1
	}

```

**Expected behavior**
an empty fall through before a default may well be accounting for exhaustive switches, this should not be an error nor require a comment to get to a lower threshold.

anupriyakkumari:
Hi, I would like to take this up if available.

PR #1549, opened 2025-10-20
useless-fallthrough: do not report fallthrough before default

anupriyakkumari:
### Description

This PR improves the `useless-fallthrough` rule to allow an empty `fallthrough` before a `default` case in switch statements, especially for enum-like constants. 

Background:

- The [exhaustive](https://github.com/nishanths/exhaustive) linter requires all enum-like constants to be covered in switch cases.
- `revive` currently requires a `default` case in switches.
- Go doesn't support exhaustive enums natively, so a `default:` case is often necessary even when all constants are handled.
- Combining a `default` with a `case` that only contains `fallthrough` is invalid.
- This change prevents the rule from flagging such valid fallthroughs used to account for exhaustive switches.

### To Reproduce

```go
type SomeType int

const (
    One SomeType = iota
    Two
    Three
    Unknown
)

func CheckParsedVal(i SomeType) int {
    switch i {
    case One:
        return 1
    case Two:
        return 2
    case Three:
        return 3
    case Unknown:
        fallthrough
    default:
        return -1
    }
}
```

### Expected behavior

* The `fallthrough` before the `default` case should **not** be reported as useless.
* This allows exhaustive switches to use a `default` case gracefully without requiring comments or lowering confidence scores.

### Checklist

* [x] Added tests to cover the new behavior
* [x] Code follows the coding style of the repository
* [x] GitHub Action build passes

### Footer 

Closes [#1538](https://github.com/mgechev/revive/issues/1538)

### DIFF
--- a/testdata/useless_fallthrough.go
+++ b/testdata/useless_fallthrough.go
@@ -29,14 +29,14 @@ func uselessFallthrough() {
 
 	switch a {
 	case 0:
-		fallthrough 
+		fallthrough
 	default:
 		println()
 	}
 
 	switch a {
 	case 0:
-		fallthrough 
+		fallthrough
 	default:
 		println()
 	case 1:
@@ -47,7 +47,7 @@ func uselessFallthrough() {
 
 	switch a {
 	case 0:
-		fallthrough 
+		fallthrough
 	default:
 		println()
 	}
@@ -70,10 +70,9 @@ func uselessFallthrough() {
 	switch a {
 	case 0:
 		//foo:bar
-		fallthrough 
+		fallthrough
 	default:
 		println()
 	}
-
 
 }


rule/useless_fallthrough.go
```diff
@@ -67,23 +67,30 @@ func (w *lintUselessFallthrough) Visit(node ast.Node) ast.Visitor {
...
+			})
 		}
 
-		w.onFailure(lint.Failure{
-			Confidence: confidence,
-			Node:       branchStmt,
-			Category:   lint.FailureCategoryCodeStyle,
-			Failure:    `this "fallthrough" can be removed by consolidating this case clause with the next one`,
-		})
-
```
ccoVeille, line 86: I don't understand this change.

It looks like the both branches are identical except the confidence level.

So there was no need to change this code

What am I missing ?

anupriyakkumari: You're right that both branches result in the same report and differ only by the confidence level. The main goal of that change was to remove the temporary confidence variable and make the branching behavior more explicit. But it's not functionally different. I'm happy to revert it if we prefer the previous version.

chavacava: > I'm happy to revert it if we prefer the previous version.

Yes, please.

alexandear (approved):

PR #1615, opened 2025-12-20
feature(`var-naming`): allow configuring accepted package names

chavacava:
This PR adds a (yet another) configuration key `acceptedPackageNames` to `var-naming` rule to allow users define a list of accepted package names. The rule will report as bad package names all that names that are in the default (hardcoded) list of bad names except those in the `acceptedPackageNames` list.


alexandear:
@chavacava maybe it would be better to create a new rule, `package-naming`? Currently, `var-naming` is overloaded with too many options. And adding a new one complicates it more.

Issue #1622, opened 2026-01-15
empty-block false positive


scop:
**Describe the bug**

Revive can flag a false empty-block positive where the block contents is irrelevant, but there's important functionality going on in the for loop controls for the block.

**To Reproduce**

```shell
cat <<EOF >main.go
package main

var count int

func step() (int, bool) {
	// This could have all kinds of functionality.
	count++
	return count, count < 5
}

func main() {
	for _, c := step(); c; _, c = step() {
		// Here, the loop body is not important, the control using step() is.
	}
}
EOF
```

```shell
$ docker run -v "$(pwd)":/work ghcr.io/mgechev/revive:v1.13.0  ./work/main.go | grep empty
./work/main.go:12:39: this block is empty, you can remove it
```

**Expected behavior**

No empty-block flagged.

**Logs**
N/A

**Desktop (please complete the following information):**
 - OS: [e.g. Ubuntu 18.04]
 - Version of Go

Using official revive Docker image.

**Additional context**
N/A

denisvmedia:
Thank you for your report. Let me take a look at it.

chavacava:
Hello @scop, thanks for reporting the problem.

The detection of empty blocks is performed by just checking for empty _container_ nodes (bodies, and `BlockStmt`) in the AST (Abstract Syntax Tree).
There is some little ad-hoc logic to avoid warning on things like `for process() {}` but covering cases like the one you report requires adding complementary logic. In these cases we need to trade-off between simplicity of the rule and its "precision" (less false positives)
I don't have actual figures but I suspect the use case is very rare.
Let's see what @denisvmedia proposes.

scop:
Sure, understood. But naively and without having a look at revive's code myself, I suppose expanding the ad-hoc logic to cover cases where there are _any_ function calls (not just the case mentioned above) in the loop controls would help with this without increasing complexity much at all.

denisvmedia:
I created a draft PR with some measurements of the impact. Please take a look @scop @chavacava

alexandear:
Related issue: #386

denisvmedia:
I ended up creating a docs PR instead: #1624

PR #1623, opened 2026-01-15
Fix `empty-block` false positive on for-loop controls

denisvmedia:
## Summary

This change fixes a false positive in the `empty-block` rule for `for` loops where the loop body is intentionally empty but the loop controls contain function calls that do the real work (e.g. `for _, c := step(); c; _, c = step() {}`).

## Motivation

Revive previously reported such loops as empty blocks, even though the side effects and control flow in `Init`/`Cond`/`Post` are the important part. This was reported in issue #1622.

## Changes

- Generalize the `empty-block` rule’s `for` handling:
  - Treat an empty-body `for` loop as **non-empty** if any of `Init`, `Cond`, or `Post` contains a function call (detected via a small `hasCallExpr` helper using `ast.Inspect`).
- Add a regression example to `testdata/empty_block.go` mirroring the reported case:
  - `emptyForWithControlCalls` with `for _, c := step(); c; _, c = step() {}` annotated with `// Must not match`.

## Testing

- `go test ./...`
- Verified `TestEmptyBlock` passes and no new failures appear.

## Benchmarks

`go test ./test -run ^$ -bench BenchmarkEmptyBlockRule -benchmem`

- With fix:
  - `BenchmarkEmptyBlockRule-11  12124  99207 ns/op  64794 B/op  1441 allocs/op`
- Without fix (original upstream code):
  - `BenchmarkEmptyBlockRule-11  12518  95066 ns/op  58974 B/op  1242 allocs/op`

For me, this is an acceptable trade-off: allocations increase slightly, runtime stays about the same, and the rule becomes more precise by no longer flagging a real-world false positive.

### DIFF
--- a/rule/empty_block.go
+++ b/rule/empty_block.go
@@ -79,8 +79,8 @@ func hasCallExpr(node ast.Node) bool {
 
 	found := false
 	ast.Inspect(node, func(n ast.Node) bool {
-		if found || n == nil {
-			return !found
+		if found {
+			return false
 		}
 
 		if _, ok := n.(*ast.CallExpr); ok {


denisvmedia:
@chavacava wdyt about this approach? Seems to work and doesn't seem to bring much overhead (at least it feels to me acceptable).

chavacava:
> @chavacava wdyt about this approach? Seems to work and doesn't seem to bring much overhead (at least it feels to me acceptable).

Thanks for the PR @denisvmedia. The complexity of the problem we face here is not in "detecting when a `for` statement has function calls in the initialization, condition our increment sections". Yes, the proposed patch detects use cases like the one described at #1622 but also will skip things like:

```go
func bar(p any) bool {
   return p > 10
}

func foo() {
  for p := 0; bar(p); p++ {
    // nothing here
  }

  for bar(p) {
    // nothing here
  }
}
```

The main difficulty is that the rule shall warn **depending on the semantics** of the called functions and not simply on the presence of these calls.

The proposed patch will fix the _false positive_ at the price of _false negatives_. It's a fair price? Maybe.

denisvmedia:
Yeah, I agree that the semantics of the called function has a meaning here as well. And yes, not all the cases can be covered. Maybe we can just keep it as is (_current state before this PR_) but documenting the limitations and suggesting the rule to be manually disabled in place. I can prepare and alternative PR that will document the limitations. @chavacava

denisvmedia:
#1624 supersedes this PR.

Issue #1628, opened 2026-02-09
`enforce-switch-style` support for type-switch


weppos:
**Is your feature request related to a problem? Please describe.**

The enforce-switch-style rule only checks regular switch statements (*ast.SwitchStmt) and ignores type switches (*ast.TypeSwitchStmt). This means a type switch without a default case or with default not in the last position is
  not flagged, even though the same style concerns apply.

```go
// This is flagged (regular switch):
switch x {
  case 1:
}

// This is NOT flagged (type switch):
switch v := x.(type) {
  case int:
}
```

> [!NOTE]
> We found this capability was missing, after we investigated a bug in one of our app where the app was incorrectly dropping some payload, due to a type-switch serialization logic that was not correctly implementing a default fallback for unknown types.

**Describe the solution you'd like**

Extend `enforce-switch-style` to also inspect `*ast.TypeSwitchStmt`. Both switch forms share the same `*ast.BlockStmt` structure, so the existing checks (missing default, default not last, all branches jump) can be applied to type switches with minimal changes.

**Describe alternatives you've considered**

A separate rule dedicated to type switches. However, since the style enforcement logic is identical for both switch forms and they share the same AST body structure, handling them in the same rule avoids duplication and is
  consistent from the user's perspective.

**Additional context**

Both `*ast.SwitchStmt` and `*ast.TypeSwitchStmt` expose a `.Body *ast.BlockStmt` field containing `*ast.CaseClause` entries, so the existing default-position and all-branches-jump checks apply without modification to the inspection logic itself.

Also for context, we use revive via [golangci-lint](https://github.com/golangci/golangci-lint).

PR #1629, opened 2026-02-09
enforce-switch-style: handle type switches

weppos:
This PR adds support for type-switch to the rule `enforce-switch-style`.

As described in the associated ticket, this is motivated by a bug we found in our codebase where we missed an else clause for a type-switch. And when trying to enforce the rule via golangci-lint, we found that the type-switch was ignored nonetheless.

Further investigation revealed that Revive is currently skipping type-switches.

This is how a failure will be reported:

```
example.go:766:3: enforce-switch-style: switch must have a default case clause (revive)
switch v := kv.(type) {
case *dns.SVCBMandatory:
        # ...
case *dns.SVCBAlpn:
        # ...
case *dns.SVCBNoDefaultAlpn:
        # ...
case *dns.SVCBPort:
```

Closes https://github.com/mgechev/revive/issues/1628


alexandear:
@weppos thank you! Merging.

Issue #1000, opened 2024-06-11
Support Suggested Fixes (--fix)


StevenACoffman:
Since Revive (and golint) was originally authored, Go analysis has added stable [SuggestedFix](https://pkg.go.dev/golang.org/x/tools/go/analysis#SuggestedFix) support for linters. 

I would like revive to provide suggested fixes with a `--fix` argument, such that they can be automatically applied as in this example:
https://github.com/golang/tools/blob/54dc8c5edb561003330ec7e649333181bbd0a9bd/go/analysis/internal/checker/checker.go#L316

BTW, there is ongoing work in [golangci-lint](https://github.com/golangci/golangci-lint) to support applying multiline suggested fixes golangci/golangci-lint#1779 and it would be nice if revive had such available when that landed.

mgechev:
Hey @StevenACoffman, thanks for the feature request. It'll be a trade-off between performance/complexity and DX. We're currently running rules concurrently over the project and we'll have to modify this logic which will be a performance regression.

I'm leaning towards skipping this feature, but I'll keep the feature request open for longer to collect more feedback on the trade-offs.

StevenACoffman:
Fair enough. We have really enjoyed `gofumpt` and similar tools that avoid needless code review nit discussions.

Some of the rules in revive are similarly straightforwardly inarguable, and we may pull them out into their own "autofixer" utility (`renew` perhaps?) if that's ok.

PR #1646, opened 2026-02-21
docs: improve comments

alexandear:
Fixed a lot of inconsistencies and mistakes in Go comments.

### DIFF
--- a/revivelib/core.go
+++ b/revivelib/core.go
@@ -208,7 +208,7 @@ func (i *ArrayFlags) String() string {
 }
 
 // Set value for array flags.
-func (i *ArrayFlags) Set(value string) error { //nolint:unparam // can't remove error return value because of the interface
+func (i *ArrayFlags) Set(value string) error {
 	*i = append(*i, value)
 
 	return nil
--- a/formatter/stylish.go
+++ b/formatter/stylish.go
@@ -15,6 +15,8 @@ type Stylish struct {
 	Metadata lint.FormatterMetadata
 }
 
+var _ lint.Formatter = (*Stylish)(nil)
+
 // Name returns the name of the formatter.
 func (*Stylish) Name() string {
 	return "stylish"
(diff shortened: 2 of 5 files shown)


denisvmedia (approved):
LGTM

lint/failure.go
```diff
@@ -78,13 +78,13 @@ type Failure struct {
 	Position   FailurePosition `json:"Position"`
 	Node       ast.Node        `json:"-"`
 	Confidence float64         `json:"Confidence"`
-	// For future use
+	// ReplacementLine is for future use.
 	ReplacementLine string `json:"ReplacementLine"`
```
ccoVeille: It's for the "-fix" , right?

Or we already use it, then this comment is outdated?

`range` and `errorf` rules are already using

revivelib/core.go
```diff
@@ -196,6 +196,7 @@ func normalizeSplit(strs []string) []string {
 }
 
 // ArrayFlags type for string list.
+// Implements [flag.Value] interface, to be used in command line arguments.
 type ArrayFlags []string
```
ccoVeille, line 201: Shouldn't we add something like this?


```suggestion
 type ArrayFlags []string

var _ flag.Value = ArrayFlags{}
```

rule/redundant_test_main_exit.go
```diff
@@ -7,7 +7,7 @@ import (
 	"github.com/mgechev/revive/lint"
 )
 
-// RedundantTestMainExitRule suggests removing Exit call in TestMain function for test files.
+// RedundantTestMainExitRule suggests removing [os.Exit] calls in TestMain function for test files.
```
ccoVeille: Maybe we could say it detects os.Exit direct and indirect calls

What do you think ?

alexandear: This rule warns only about direct `os.Exit` calls. I'll mention also `syscall.Exit`.

revivelib/core.go
alexandear: Great suggestion. Will add.

lint/failure.go
alexandear: It looks like `--fix` is not working, according to #1000. I'll remove the comment.

revivelib/core.go
alexandear: Now `//nolint:unparam` is redundant.

PR #1649, opened 2026-02-21
chore: enable reassign linter

alexandear:
Enable [reassign](https://golangci-lint.run/docs/linters/configuration/#reassign) linter to detect `color.NoColor = true` assignments. Fix appeared lint issues.

This automatically fixes failing formatting tests when run with benchmarks `go test -bench . -benchmem -count=1 ./...`

### DIFF
--- a/cli/main.go
+++ b/cli/main.go
@@ -41,9 +41,7 @@ func fail(err string) {
 func RunRevive(extraRules ...revivelib.ExtraRule) {
 	// Move parsing flags outside of init(); otherwise, tests don't work properly.
 	// More info: https://github.com/golang/go/issues/46869#issuecomment-865695953
-	if err := initConfig(); err != nil {
-		fail(err.Error())
-	}
+	initConfig()
 
 	if versionFlag {
 		fmt.Print(getVersion(builtBy, date, commit, version))
@@ -146,7 +144,7 @@ func buildDefaultConfigPath() string {
 	return result
 }
 
-func initConfig() error {
+func initConfig() {
 	if os.Getenv("REVIVE_FORCE_COLOR") == "1" {
 		color.NoColor = false //nolint:reassign // We want to reassign the default value of NoColor to force colorizing for non-TTY environments.
 	}
@@ -175,7 +173,6 @@ func initConfig() error {
 	flag.BoolVar(&setExitStatus, "set_exit_status", false, exitStatusUsage)
 	flag.IntVar(&maxOpenFiles, "max_open_files", 0, maxOpenFilesUsage)
 	flag.Parse() //revive:disable-line:deep-exit
-	return nil
 }
 
 // getVersion returns build info (version, commit, date, and builtBy).


formatter/friendly.go
```diff
@@ -68,15 +68,10 @@ func (f *Friendly) printFriendlyFailure(sb *strings.Builder, failure lint.Failur
...
-)
-
 func (*Friendly) printHeaderRow(sb *strings.Builder, failure lint.Failure, severity lint.Severity) {
-	emoji := warningEmoji
+	emoji := color.YellowString("⚠")
 	if severity == lint.SeverityError {
-		emoji = errorEmoji
+		emoji = color.RedString("✘")
 	}
 	sb.WriteString(table([][]string{{emoji, ruleDescriptionURL(failure.RuleName), color.GreenString(failure.Failure)}}))
```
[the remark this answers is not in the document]
alexandear: Implemented

formatter/friendly.go
```diff
@@ -59,26 +69,17 @@ func (f *Friendly) Format(failures <-chan lint.Failure, config lint.Config) (str
...
-)
-
-func (*Friendly) printHeaderRow(sb *strings.Builder, failure lint.Failure, severity lint.Severity) {
-	emoji := warningEmoji
-	if severity == lint.SeverityError {
-		emoji = errorEmoji
-	}
-	sb.WriteString(table([][]string{{emoji, ruleDescriptionURL(failure.RuleName), color.GreenString(failure.Failure)}}))
+func (*Friendly) printHeaderRow(sb *strings.Builder, firstColumn string, failure lint.Failure) {
+	sb.WriteString(table([][]string{{firstColumn, ruleDescriptionURL(failure.RuleName), color.GreenString(failure.Failure)}}))
```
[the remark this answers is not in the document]
alexandear: We can live with this.

formatter/formatter_test.go
```diff
@@ -193,15 +193,14 @@ file.go
 		},
 	} {
 		t.Run(name, func(t *testing.T) {
+			t.Setenv("NO_COLOR", "true")
```
ccoVeille, line 197: This could become a linter, haha

Using color.NoColor = true should be reported.

But it's an external lib, I don't know if we have something like that in revive

formatter/formatter_test.go
```diff
@@ -193,15 +193,14 @@ file.go
...
 			realStdout := os.Stdout
 			fakeStdout, err := os.Create(filepath.Join(t.TempDir(), "fakeStdout"))
 			if err != nil {
 				t.Fatal(err)
 			}
 			os.Stdout = fakeStdout
-			defer func() {
-				os.Stdout = realStdout
-			}()
+			t.Cleanup(func() { os.Stdout = realStdout })
```
ccoVeille, line 204: At some point, we should refactor to have an io.Writer in the lib that default to os.StdOut allowing us to test without such dirty things that is not clean.

It could become a linter also

formatter/formatter_test.go
alexandear: We have https://golangci-lint.run/docs/linters/configuration/#reassign for this. I'll enable it.

formatter/formatter_test.go
alexandear: This hacky test is intentional, see #921. I added a comment.

formatter/friendly.go
```diff
@@ -59,26 +69,17 @@ func (f *Friendly) Format(failures <-chan lint.Failure, config lint.Config) (str
...
 	if err := f.printFilePosition(sb, failure); err != nil {
 		return err
 	}
 	_, err := sb.WriteString("\n\n")
 	return err
 }
 
-var (
-	errorEmoji   = color.RedString("✘")
-	warningEmoji = color.YellowString("⚠")
```
alexandear, line 73: These lines are problematic and causes go test with bench to fail.

formatter/formatter_test.go
ccoVeille: Indeed. Sometimes solving an issue is as simple as figuring what is already available to fix it

formatter/formatter_test.go
ccoVeille: Thanks, it's clearer

.golangci.yml
```diff
@@ -103,6 +104,9 @@ linters:
       allow-unused: false
       require-explanation: true
       require-specific: true
+    reassign:
+      patterns:
+        - .*
```
ccoVeille, line 109: Thank you for this, I had never noticed there was an option in `reassign`

alexandear: Without this option, `reassign` detects nothing.

PR #1651, opened 2026-02-23
context-keys-type,errorf,modifies-value-receiver: remove unused type checking

alexandear:
Looks like we can remove type checking for `context-keys-type`, `errorf`, and `modifies-value-receiver`.


alexandear:
My bad. We can't.

Tests should have caught this. But they don't.

PR #1650, opened 2026-02-23
refactor: drop dependency on github.com/fatih/structtag

alexandear:
This PR drops one dependency.
Because https://github.com/fatih/structtag seems unmaintained, and we do not utilize all the functions from the library. We need only `Parse`.

Our implementation is a little bit faster:

```console
❯ go test -bench . -benchmem -count=1 -run=^$ ./internal/structtag/...
goos: darwin
goarch: arm64
pkg: github.com/mgechev/revive/internal/structtag
cpu: Apple M4 Pro
BenchmarkParseCompare/single/revive-12          23213113                51.08 ns/op           88 B/op          3 allocs/op
BenchmarkParseCompare/single/fatih-12           18595112                68.07 ns/op          112 B/op          4 allocs/op
BenchmarkParseCompare/with_options/revive-12    16504845                72.00 ns/op          120 B/op          3 allocs/op
BenchmarkParseCompare/with_options/fatih-12     13443501                87.82 ns/op          144 B/op          4 allocs/op
BenchmarkParseCompare/multiple/revive-12         7166770               165.9 ns/op           296 B/op          7 allocs/op
BenchmarkParseCompare/multiple/fatih-12          6108170               195.6 ns/op           352 B/op         10 allocs/op
BenchmarkParseCompare/complex/revive-12          5118066               232.5 ns/op           432 B/op          9 allocs/op
BenchmarkParseCompare/complex/fatih-12           4201783               269.2 ns/op           480 B/op         12 allocs/op
PASS
```

<details><summary>bench_test.go</summary>
<p>

```go
package structtag_test

import (
	"testing"

	fatihst "github.com/fatih/structtag"

	"github.com/mgechev/revive/internal/structtag"
)

func BenchmarkParseCompare(b *testing.B) {
	tests := []struct {
		name string
		tag  string
	}{
		{
			name: "single",
			tag:  `json:"foo"`,
		},
		{
			name: "with_options",
			tag:  `json:"foo,omitempty,string"`,
		},
		{
			name: "multiple",
			tag:  `json:"foo,omitempty" xml:"bar" db:"baz,primarykey"`,
		},
		{
			name: "complex",
			tag:  `json:"id,omitempty,string" xml:"id,attr" db:"id,primarykey" validate:"required,uuid4"`,
		},
	}

	for _, tt := range tests {
		b.Run(tt.name+"/revive", func(b *testing.B) {
			for b.Loop() {
				_, _ = structtag.Parse(tt.tag)
			}
		})
[...]

PR #1663, opened 2026-03-04
chore: Add structtag copyright

alexandear:
```
I'm noticing it now.

Shouldn't we have added a license file

https://github.com/fatih/structtag/blob/master/LICENSE#L1-L8

And a README in this package to explain the context of this "inlined import"
```

_Originally posted by @ccoVeille in https://github.com/mgechev/revive/issues/1650#issuecomment-3995557248_

### DIFF
--- a/THIRD_PARTY_NOTICES
+++ b/THIRD_PARTY_NOTICES
@@ -1,0 +1,37 @@
+This file contains notices for third-party software used in Revive.
+
+---
+
+COMPONENT: structtag (internal/structtag)
+SOURCE: https://github.com/fatih/structtag
+LICENSE: BSD 3-Clause
+
+BSD 3-Clause License
+
+Copyright (c) 2017, Fatih Arslan
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+* Redistributions of source code must retain the above copyright notice, this
+  list of conditions and the following disclaimer.
+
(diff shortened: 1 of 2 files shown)


THIRD_PARTY_NOTICES
```diff
@@ -0,0 +1,37 @@
+This file contains notices for third-party software used in Revive.
```
ccoVeille, line 1: Plural no ?


```suggestion
This file contains notices for third-party softwares used in Revive.
```

internal/structtag/structtag.go
ccoVeille, line 1: I would add README.md rile explaining why the code was imported.

And maybe a license file, in addition if this header.

THIRD_PARTY_NOTICES
ccoVeille, line 1: Is this a classic file name that is normalized somehow ?

It's the first time I'm facing this

internal/structtag/structtag.go
ccoVeille, line 1: Was about moving internal/structtag while folder to Internal/vendored/structtag ?

I would provide a clear namespace

THIRD_PARTY_NOTICES
alexandear: It's common. For example, https://github.com/microsoft/vscode/blob/main/ThirdPartyNotices.txt

ccoVeille:
Closed because of https://github.com/mgechev/revive/pull/1664

PR #1679, opened 2026-03-09
fix: avoid redundant-import-alias false positives for mismatched package names

### LINKED ISSUE #1662: `redudant-import-alias` false positive
**Describe the bug**

Linter reports it as:

```
test.go:19:2: redundant-import-alias: Import alias "utls" is redundant (revive)
	utls "github.com/enetx/utls"
```

But the fact is that the package is actually named tls, and thus aliasing to utls is not redundant (and is preferred if `"crypto/tls"` at the same time).

**To Reproduce**

Try importing a package named boo, but having a package uri as `github.com/orgname/foo.

**Expected behavior**
No reporting.

 - OS: macOS Tahoe 26.3
 - go 1.26.0
 - revive used through golangci-lint 1.10.1

**Additional context**
I have a feeling we already discussed this issue, but I might be mistaken.

SergioChan:
## Summary
- resolve `redundant-import-alias` aliases against the imported package name from type information when available
- keep the existing path-segment fallback for unresolved imports so current behavior remains stable when type info is partial
- add a focused regression unit test covering `github.com/enetx/utls` (import path segment `utls`, package name `tls`)

## Testing
- `go test ./rule -run TestGetImportPackageNameUsesImportedPackageNameWhenTypeInfoAvailable`
- `go test ./test -run TestRedundantImportAlias`

## Related
Closes #1662

### DIFF
--- a/rule/redundant_import_alias.go
+++ b/rule/redundant_import_alias.go
@@ -43,14 +43,8 @@ func (*RedundantImportAlias) Name() string {
 }
 
 func getImportPackageName(imp *ast.ImportSpec, typesInfo *types.Info) string {
-	if typesInfo != nil {
-		if obj, ok := typesInfo.Defs[imp.Name]; ok {
-			if pkgName, ok := obj.(*types.PkgName); ok {
-				if imported := pkgName.Imported(); imported != nil && imported.Name() != "" {
-					return imported.Name()
-				}
-			}
-		}
+	if name := importPackageNameFromTypes(imp, typesInfo); name != "" {
+		return name
 	}
 
 	const pathSep = "/"
@@ -63,4 +57,27 @@ func getImportPackageName(imp *ast.ImportSpec, typesInfo *types.Info) string {
 	}
 
 	return strings.Trim(path[i+1:], strDelim)
+}
+
+func importPackageNameFromTypes(imp *ast.ImportSpec, typesInfo *types.Info) string {
+	if typesInfo == nil || imp.Name == nil {
+		return ""
+	}
+
+	obj, ok := typesInfo.Defs[imp.Name]
+	if !ok {
+		return ""
+	}
+
+	pkgName, ok := obj.(*types.PkgName)
+	if !ok {
+		return ""
+	}
+
+	imported := pkgName.Imported()
+	if imported == nil {
+		return ""
+	}
+
+	return imported.Name()
 }
--- a/rule/redundant_import_alias_test.go
+++ b/rule/redundant_import_alias_test.go
@@ -15,6 +15,7 @@ func TestGetImportPackageNameUsesImportedPackageNameWhenTypeInfoAvailable(t *tes
 	}
 
 	typesInfo := &types.Info{Defs: map[*ast.Ident]types.Object{}}
+	// github.com/enetx/utls intentionally exposes package name "tls".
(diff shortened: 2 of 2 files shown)


rule/redundant_import_alias.go
```diff
@@ -38,7 +42,17 @@ func (*RedundantImportAlias) Name() string {
...
+func getImportPackageName(imp *ast.ImportSpec, typesInfo *types.Info) string {
+	if typesInfo != nil {
+		if obj, ok := typesInfo.Defs[imp.Name]; ok {
+			if pkgName, ok := obj.(*types.PkgName); ok {
+				if imported := pkgName.Imported(); imported != nil && imported.Name() != "" {
+					return imported.Name()
+				}
+			}
+		}
+	}
```
ccoVeille: Can we move this to a dedicated method that will use leave early?

Having 4 imbricated if is a stop for me

rule/redundant_import_alias_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func TestGetImportPackageNameUsesImportedPackageNameWhenTypeInfoAvailable(t *testing.T) {
+	alias := ast.NewIdent("utls")
+	imp := &ast.ImportSpec{
+		Name: alias,
+		Path: &ast.BasicLit{Value: `"github.com/enetx/utls"`},
+	}
+
+	typesInfo := &types.Info{Defs: map[*ast.Ident]types.Object{}}
+	typesInfo.Defs[alias] = types.NewPkgName(token.NoPos, nil, alias.Name, types.NewPackage("github.com/enetx/utls", "tls"))
```
ccoVeille, line 19: Maybe a comment here about the fact this module/package differs somehow from the usual way to name it.

Without it, it's unclear why it should be like this.

SergioChan:
Thanks for the review — addressed in 10271d7.

- Refactored the type-info lookup into `importPackageNameFromTypes` with early returns (no nested conditionals now).
- Added an inline note in the regression test explaining why `github.com/enetx/utls` resolves to package name `tls`.

Validation rerun:
- `go test ./rule -run TestGetImportPackageNameUsesImportedPackageNameWhenTypeInfoAvailable`
- `go test ./test -run TestRedundantImportAlias`

rule/redundant_import_alias.go
```diff
@@ -15,12 +16,15 @@ type RedundantImportAlias struct{}
 func (*RedundantImportAlias) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
 	var failures []lint.Failure
 
+	_ = file.Pkg.TypeCheck()
```
alexandear, line 19: Thanks for this PR. However, we've decided not to make this rule type-aware. This topic came up a few times with other rules, and each time we chose to accept a possible false positive instead of adding a `file.Pkg.TypeCheck` call. Doing so would make the rule unusable for many users (see #1277 for context).

SergioChan:
Thanks for the clear guidance and for linking the prior context.

That makes sense — if `TypeCheck` makes the rule impractical for many users, this approach is not a good fit for revive. I’m going to close this PR instead of pushing further on the type-aware path.

Appreciate the review.

SergioChan:
Closing per maintainer guidance to avoid making this rule type-aware.

PR #1682, opened 2026-03-11
docs: fix untyped rules descriptions

alexandear:
1. Fix the "Typed" column for a few rules in README.
2. Add missing rules to `untyped.toml`.
 
The config file `untyped.toml` has been added in #3 and contains the list of untyped rules. Since then, it has never been updated. This config is also used to measure [Revive's speed](https://revive.run/docs#revives-speed).


ccoVeille:
What about simply deleting the file ?

I mean does this speed reporting is necessary?

It was interesting for when people migrated out of golint

Also, if we have a need for this, we could consider adding a flag that will disable rules by using an interface check.

My point is if this file was forgotten for 8 years, we don't need it. No one did.

alexandear:
> What about simply deleting the file ?
> 
> I mean does this speed reporting is necessary?
> 
> It was interesting for when people migrated out of golint
> 
> Also, if we have a need for this, we could consider adding a flag that will disable rules by using an interface check.
> 
> My point is if this file was forgotten for 8 years, we don't need it. No one did.

Removing the `untyped.toml` was my first thought. But then I found the file is mentioned in the README:

```console
❯ rg untyped.toml
README.md
803:$ time revive -config untyped.toml kubernetes/... > /dev/null
```

And decided to add missing rules instead of rewriting a README.

And after creating this PR, Copilot found that some rules in the "Available Rules" table have the wrong "Typed" value. So this PR fixes it as well.

Removing the "golint" mention from the README is a separate task that we are going to do in the future. We will remove `untyped.toml` as well.

ccoVeille:
I'm fine with this approach

PR #1713, opened 2026-04-29
comment-spacings: ignore //#nosec directive by default

alexandear:
Extend the comment-spacings rule to ignore `//#nosec` comments by default.

https://github.com/securego/gosec/blob/5f4eec95fa28ce5dc6cf555de8c242cb57545f01/cmd/gosecutil/tools.go#L79

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -260,6 +260,11 @@ _Configuration_: ([]string) list of exceptions. For example, to accept comments 
 
 You need to add both `"mypragma:"` and `"+optional"` in the configuration
 
+The following comment prefixes are allowed by default:
+
+- `//#nosec` — [gosec](https://github.com/securego/gosec) security scanner directive
+- Go directive comments matching `//[a-z0-9]+:[a-z0-9]` (e.g. `//nolint:linter`, `//go:generate`, `//revive:disable:rule`)
+
 Configuration example:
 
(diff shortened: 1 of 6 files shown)


denisvmedia:
AFAIU, the comment should be used as `// #nosec` (at least I use it this way in my projects, and it works). Maybe instead warn on that?

alexandear:
> AFAIU, the comment should be used as `// #nosec` (at least I use it this way in my projects, and it works). Maybe instead warn on that?

The `gosec` linter itself in their codebase suggests using both:

- `//#nosec`

https://github.com/securego/gosec/blob/5f4eec95fa28ce5dc6cf555de8c242cb57545f01/testutils/g109_samples.go#L108
https://github.com/securego/gosec/blob/5f4eec95fa28ce5dc6cf555de8c242cb57545f01/testutils/g305_samples.go#L30
https://github.com/securego/gosec/blob/5f4eec95fa28ce5dc6cf555de8c242cb57545f01/cmd/tlsconfig/tlsconfig.go#L67

- `// #nosec`

https://github.com/securego/gosec/blob/5f4eec95fa28ce5dc6cf555de8c242cb57545f01/testutils/g101_samples.go#L286

So, we can't fight with that.

On my work projects I prefer `//gosec:disable` or `//nolint:gosec` directives.

denisvmedia (approved):
LGTM

PR #1731, opened 2026-05-13
docs: update go instructions for modern go patterns

denisvmedia:
This pull request updates the Go code review instructions to ensure recommendations are always aligned with the project's Go version, and to strongly encourage the use of modern Go features when available. It introduces explicit guidance and examples for using new standard library features, and adds new "do/don't" sections for idiomatic usage. The goal is to prevent unnecessary rewrites or regressions to older patterns when the codebase supports newer Go versions.

Key changes include:

### Go Version Awareness and Guidance

- Added a new section instructing reviewers and tools to always check the `go.mod` version before suggesting changes, and to avoid recommending hand-rolled or downgraded alternatives to modern standard library features when the Go version supports them. This section also lists specific features by Go version and provides rules for proposing code that uses features newer than the project's Go version.

### Modern Go Features: Examples and Best Practices

- Introduced a comprehensive "Modern Go Features (Go 1.21+) — Do / Don't" section, with clear before/after code examples for features such as built-in `min`/`max`/`clear`, `slices` and `maps` packages, `errors.Join`, `log/slog`, range-over-integer, loop variable scoping, `cmp.Or`, range-over-function iterators, `omitzero` JSON tag, generic type aliases, `testing/synctest`, `testing.B.Loop`, and container-aware `GOMAXPROCS`. This section instructs not to suggest reverting to older patterns when the modern feature is available.

### Pitfall List Updates

- Expanded the "Common Pitfalls to Avoid" section to explicitly call out suggesting hand-rolled replacements for standard library features, re-introducing unnecessary loop-variable shadowing in Go 1.22+, and replacing structured logging with unstructured logging, reinforcing the new guidance throughout the document.

### DIFF
--- a/.github/instructions/go.instructions.md
+++ b/.github/instructions/go.instructions.md
@@ -16,7 +16,8 @@ and [Google's Go Style Guide](https://google.github.io/styleguide/go/).
 - Always determine the project's Go version from the `go` directive in `go.mod` before suggesting changes.
 - Do **not** propose rewrites that replace modern standard-library features (Go 1.21+) with hand-rolled equivalents.
 - Do **not** flag, downgrade, or "polyfill" the following modern features when the `go.mod` permits them:
-  - Go 1.21: built-ins `min`, `max`, `clear`; stdlib packages `slices`, `maps`, `cmp`; structured logging `log/slog`; `errors.Join`.
+  - Go 1.20: `errors.Join`.
+  - Go 1.21: built-ins `min`, `max`, `clear`; stdlib packages `slices`, `maps`, `cmp`; structured logging `log/slog`.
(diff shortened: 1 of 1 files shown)


ccoVeille:
@denisvmedia if you are interested @samber cc-skills-golang contains some others ideas you might be interested to include 

https://github.com/samber/cc-skills-golang/blob/main/skills%2Fgolang-modernize%2FSKILL.md

denisvmedia:
@ccoVeille thanks, good idea. Maybe we can have it as a follow up PR, unless you want to add any paragraph from there in particular right in this PR

ccoVeille (approved):
Some feedbacks made my Copilot are relevant.

Except that 👍

ccoVeille:
> @ccoVeille thanks, good idea. Maybe we can have it as a follow up PR, unless you want to add any paragraph from there in particular right in this PR

I'm fine with both. Let's wait for @alexandear and @chavacava feedbacks

alexandear (approved):

PR #1739, opened 2026-06-02
feat: add specify-disable-rule directive

### LINKED ISSUE #1720: RFE: option to require specifying rules to disable directives

Revive allows disabling itself for specific code lines using inline comment directives. It is not uncommon to be too lax when doing this by leaving out the specific rules one wants to disable. This can lead to inadvertently disabling other issues there might be on the same line as the intentionally disabled one.

So in essence, this would forbid "naked" `//revive:disable(-line)` and require `//revive:disable(-line):RULES-GO-HERE` instead.

Similarly for `//revive:enable` I suppose.

Add a configuration option similar to the existing `directive.specify-disable-reason` to require specifying at least one rule with disable directives, let's say `directive.specify-disable-rules` (or `*-rule`).

Using golangci-lint's generic linter message text scraping to suppress instead of using revive comments. This works only with golangci-lint, not with plain revive. Having the functionality in revive would work with both.

**Additional context**

N/A

toller892:
## What

Add a new `directive.specify-disable-rules` configuration option that forbids "naked" `//revive:disable` and `//revive:disable-line` directives, requiring at least one rule name to be specified.

## Why

When `//revive:disable` is used without specifying rules, it disables *all* rules for the affected scope. This can inadvertently suppress findings beyond what was intended. The new directive enforces explicit rule specification, similar to the existing `specify-disable-reason` directive.

## Configuration

```toml
[directive.specify-disable-rule]
```

With severity (defaults to warning):

```toml
[directive.specify-disable-rule]
severity = "error"
```

## Behavior

- `//revive:disable` — Error: rule name required (when directive enabled)
- `//revive:disable:cyclomatic` — Allowed
- `//revive:disable-line` — Error: rule name required (when directive enabled)
- `//revive:disable-line:cyclomatic` — Allowed
- `//revive:enable` — Allowed (not affected)

Closes #1720


alexandear (approved):

alexandear:
@toller892 Thank you for the contribution!

PR #1747, opened 2026-06-18
feature: new rule `redundant-canonical-import`

alexandear:
This rule warns on [canonical import path comments](https://go.dev/doc/go1.4#canonicalimports) (e.g. `package pdf // import "rsc.io/pdf"`). In module mode the Go toolchain ignores these comments entirely. So, these canonical import comments can be removed.

See https://pkg.go.dev/cmd/go@go1.11.0#hdr-Import_path_checking:
> Import path checking is also disabled when using modules. Import path comments are obsoleted by the go.mod file's module statement.

### Examples

- https://go.googlesource.com/oauth2/+/refs/tags/v0.36.0/github/github.go#6
  ```go
  package github // import "golang.org/x/oauth2/github"
  ```
- https://github.com/open-telemetry/opentelemetry-go/blob/c2cac8c3d170dbaefe43f6ae5dad650001c5ca54/log/logger.go#L4
  ```go
  package log // import "go.opentelemetry.io/otel/log"
  ```
- https://gitlab.com/cznic/sqlite/-/blob/697300ffaa56b03b79e490dc960f218794cb2a75/sqlite.go#L7
  ```go
  package sqlite // import "modernc.org/sqlite"
  ```

### DIFF
--- a/rule/redundant_canonical_import.go
+++ b/rule/redundant_canonical_import.go
@@ -1,0 +1,64 @@
+package rule
+
+import (
+	"fmt"
+	"regexp"
+	"strings"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// matches a canonical import comment body, e.g.: import "rsc.io/pdf" or import `rsc.io/pdf`
+var canonicalImportCommentRegexp = regexp.MustCompile("^import\\s+(?:\"[^\"]+\"|`[^`]+`)$")
+
+// RedundantCanonicalImport warns on canonical import path comments that are redundant in module mode (Go 1.11+).
+// See https://go.dev/doc/go1.4#canonicalimports.
+type RedundantCanonicalImport struct{}
+
+// Apply applies the rule to given file.
+func (*RedundantCanonicalImport) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	if !file.Pkg.IsAtLeastGoVersion(lint.Go111) {
+		return nil
+	}
+
+	packageLine := file.ToPosition(file.AST.Name.End()).Line
+
+	for _, cg := range file.AST.Comments {
+		for _, c := range cg.List {
+			if c.Pos() < file.AST.Name.End() {
+				continue // before the package name
+			}
+			if file.ToPosition(c.Pos()).Line > packageLine {
+				return nil // past the package clause line; comments are ordered by position
+			}
+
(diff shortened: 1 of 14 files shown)


rule/redundant_canonical_import.go
```diff
@@ -0,0 +1,64 @@
...
+		return nil
+	}
+
+	packageLine := file.ToPosition(file.AST.Name.End()).Line
+
+	for _, cg := range file.AST.Comments {
+		for _, c := range cg.List {
+			if c.Pos() < file.AST.Name.End() {
+				continue // before the package name
+			}
```
Acollie: Nit: Might it be worth extracting this to a function `isBeforePackageName` that way the code can read
```suggestion
			if isBeforePackageName(c, file){
				continue
			}
```

rule/redundant_canonical_import.go
```diff
@@ -0,0 +1,64 @@
...
+	return nil
+}
+
+// commentBody strips the comment markers from c.Text, handling both line and block comments.
+func commentBody(text string) string {
+	text = strings.TrimPrefix(text, "//")
+	text = strings.TrimPrefix(text, "/*")
+	text = strings.TrimSuffix(text, "*/")
+	return strings.TrimSpace(text)
+}
```
Acollie, line 65: handles a few distinct cases (line //, block /* */, whitespace). Could be worth a small table-driven test to pin that behavior down and guard against regressions.

Acollie (changes requested):
Really like this suggestion just a few Nit's on my side

rule/redundant_canonical_import.go
alexandear: `TestRedundantCanonicalImport` already covers your suggestion.

alexandear:
@Acollie thank you for the review.

Acollie (approved):
Thanks for the changes, Looks great thanks

ccoVeille:
For record, it's being discussed to add it to modernizer

- https://github.com/golang/tools/pull/655

ccoVeille:
modernizer now supports it

- https://github.com/golang/tools/pull/655#issuecomment-4897444154

@Alexander @chavacava maybe the linter coded in revive could be removed. It was not released yet in revive

alexandear:
@ccoVeille thank you, removed.

PR #1753, opened 2026-06-26
feat: Export a few config functions for golangci-lint

alexandear:
Golangci-lint integrates Revive by copy-pasting several of Revive's unexported functions and variables (see [pkg/golinters/revive/revive.go](https://github.com/golangci/golangci-lint/blob/a9b484189fa255f772b661b0e8f2e93a02252ae6/pkg/golinters/revive/revive.go#L278)). That duplication is a recurring maintenance burden and a silent source of drift.

With this merged, Golangci-lint can drop its copied Revive internals and call the exported functions directly.


alexandear:
@ldez please review when you have time.

ldez:
I think that the following functions should not be inside Revive:

```go
AllRuleNames() []string
DefaultRuleNames() []string
EnabledRuleNames(conf) []string
```

Those functions are only used by golangci-lint for the debug.

From the POV of revive API, I think better to have the following functions (that return copy of slices):

```go
AllRules() []lint.Rule
DefaultRules() []lint.Rule
EnabledRules(conf) []lint.Rule
```

PR #1770, opened 2026-08-15
fix: checkstyle formatter emits invalid XML for filenames with & or "

### LINKED ISSUE #1769: checkstyle formatter emits invalid XML for filenames containing & or "
## Description

The `checkstyle` formatter emits XML that no parser will accept when a filename contains `&` or `"`.

```
$ revive -formatter checkstyle './a&b.go'
<?xml version='1.0' encoding='UTF-8'?>
<checkstyle version="5.0">
    <file name="./a&b.go">
```

```
$ ... | python3 -c "import sys,xml.dom.minidom as m; m.parseString(sys.stdin.read())"
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 3, column 24
```

A directory containing a `"` breaks the attribute the same way, producing `<file name="./q"dir/a.go">`.

Anything consuming the checkstyle output in CI, which is the whole point of that formatter, fails on the file rather than reporting the lint results.

## Cause

`formatter/checkstyle.go` renders through `text/template`, which does no escaping. One field is escaped explicitly, with an `xml.Escape` call on `failure.Failure`, but the filename used as the map key and `RuleName` are interpolated raw.

That existing escape is what suggests this is an oversight rather than a decision: the need was known and handled for one of the three interpolated values.

## Proposed fix

Apply the same `xml.Escape` idiom to the other two values. Small change, and it keeps the single-escape behaviour of the field that is already handled rather than switching the template engine, which would double-escape it.

Happy to open a PR for this; I have the change and a test that asserts the output is well-formed XML rather than string-matching it.

VXNCXNX:
Closes #1769

## What's broken

The `checkstyle` formatter emits XML that no parser accepts when a filename contains `&` or `"`.

```
$ revive -formatter checkstyle './a&b.go'
    <file name="./a&b.go">

$ ... | python3 -c "import sys,xml.dom.minidom as m; m.parseString(sys.stdin.read())"
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 3, column 24
```

A directory containing a `"` breaks the attribute the same way, giving `<file name="./q"dir/a.go">`.


formatter/checkstyle.go
ccoVeille: Nitpicking: I would have registered the method as a func available in the template and would have used it in the template

See https://pkg.go.dev/text/template#Template.Funcs

Here I feel we mix values and templating.

The difference is slight, but enough to make me react

alexandear: Thanks, done.

alexandear (approved):

alexandear:
@VXNCXNX thank you for the contribution!

formatter/checkstyle.go
ccoVeille: Thank you

PR #1771, opened 2026-08-18
rule: add no-increment-decrement rule

### LINKED ISSUE #910: Opposite of increment-decrement (for non-loop-counter usage)

The increment-decrement rule is quite opinionated, as many rules are. However, its suggestion is problematic in the sense that e.g.
- it's not that common of a practice in my experience to use `++`/`--` in contexts other than loop counters in Go code, e.g. stdlib
- many programming languages don't even have those operators (for example Python, Rust, Swift, Zig)

increment-decrement can be disabled like all others, no problem with that, and that does not require a solution.

But for the above reasons, a rule that would do the *opposite* of increment-decrement for anything but loop counters would be nice to have.

Just disabling increment-decrement, but that won't help with adoption of the `+= 1` style.

**Additional context**

N/A.

ChrisJr404:
This adds `no-increment-decrement`, a new rule that is the mirror image of the existing `increment-decrement` rule, for people who prefer the explicit `i += 1` / `i -= 1` forms over `i++` / `i--`.

As raised in #910, `increment-decrement` is opinionated in the direction of `++`/`--`, but plenty of styles (and languages) go the other way, and in practice `++`/`--` are rarely used in Go outside of loop counters. The new rule spots `i++` and `i--` statements and proposes replacing them with `i += 1` and `i -= 1`. Increment/decrement statements that appear as the post statement of a `for` loop are intentionally left alone, since that is the idiomatic place for a loop counter (see the `for i := 0; i < n; i++` example in the docs).

Like every other revive rule it is opt-in, so the default behaviour is unchanged. It works purely on the AST and requires no type information, so it is registered in `untyped.toml`.

Naming: I went with `no-increment-decrement` to make it easy to find as the counterpart of `increment-decrement`, but I'm happy to rename it if you'd prefer something else.

Changes:
- `rule/no_increment_decrement.go` — the rule.
- `test/no_increment_decrement_test.go` + `testdata/no_increment_decrement.go` — fixtures covering standalone `++`/`--`, a selector operand, a counter incremented inside a loop body (flagged), and `for`-loop counters (not flagged).
- Registered the rule in `config/config.go` and `untyped.toml`, bumped the rule count in `config/config_test.go`, and documented it in `RULES_DESCRIPTIONS.md`.

Testing: `go test ./...` passes, and both `revive --config revive.toml ./...` and `golangci-lint run` are clean on the new files.

Closes #910

### DIFF
--- a/rule/no_increment_decrement.go
+++ b/rule/no_increment_decrement.go
@@ -1,0 +1,76 @@
+package rule
+
+import (
+	"fmt"
+	"go/ast"
+	"go/token"
+
+	"github.com/mgechev/revive/lint"
+)
+
+// NoIncrementDecrementRule suggests replacing `i++` and `i--` with `i += 1` and `i -= 1`,
+// except when they are used as the post statement of a for loop (i.e. as loop counters).
+// It is the opposite of the increment-decrement rule.
+type NoIncrementDecrementRule struct{}
+
+// Apply applies the rule to given file.
+func (*NoIncrementDecrementRule) Apply(file *lint.File, _ lint.Arguments) []lint.Failure {
+	var failures []lint.Failure
+
+	walker := &lintNoIncrementDecrement{
+		file:      file,
+		loopPosts: map[ast.Stmt]struct{}{},
+		onFailure: func(failure lint.Failure) {
+			failures = append(failures, failure)
+		},
+	}
+
+	ast.Walk(walker, file.AST)
+
+	return failures
+}
+
+// Name returns the rule name.
+func (*NoIncrementDecrementRule) Name() string {
+	return "no-increment-decrement"
+}
+
+type lintNoIncrementDecrement struct {
+	file      *lint.File
+	loopPosts map[ast.Stmt]struct{}
+	onFailure func(lint.Failure)
+}
+
+func (w *lintNoIncrementDecrement) Visit(n ast.Node) ast.Visitor {
+	switch stmt := n.(type) {
+	case *ast.ForStmt:
+		if stmt.Post != nil {
+			// The post statement of a for loop is the idiomatic place for a loop
+			// counter, so it is exempted from the rule.
+			w.loopPosts[stmt.Post] = struct{}{}
+		}
+	case *ast.IncDecStmt:
+		if _, isLoopPost := w.loopPosts[stmt]; isLoopPost {
+			return w
+		}
+
+		var replacement string
+		switch stmt.Tok {
+		case token.INC:
+			replacement = "+= 1"
+		case token.DEC:
+			replacement = "-= 1"
+		default:
+			return w
+		}
+
+		w.onFailure(lint.Failure{
+			Confidence: 0.8,
+			Node:       stmt,
+			Category:   lint.FailureCategoryUnaryOp,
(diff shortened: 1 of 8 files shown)


rule/no_increment_decrement.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func (w *lintNoIncrementDecrement) Visit(n ast.Node) ast.Visitor {
+	switch stmt := n.(type) {
+	case *ast.ForStmt:
+		if stmt.Post != nil {
+			// The post statement of a for loop is the idiomatic place for a loop
+			// counter, so it is exempted from the rule.
+			w.loopPosts[stmt.Post] = struct{}{}
+		}
```
chavacava, line 51: I think it's possible to handle the special case of increments in the loop post without keeping track of loopPosts in a list: when visiting a `forStmt` launch the visitor on the for body and then return nil; this will make the visitor to skip the post part of the for.

Something like:


```suggestion
		w.Visit(stmt.Body) //maybe a check stmt.Body != nil is necessary before calling Visit
		return nil
```

rule/no_increment_decrement.go
```diff
@@ -0,0 +1,76 @@
...
+		case token.INC:
+			replacement = "+= 1"
+		case token.DEC:
+			replacement = "-= 1"
+		default:
+			return w
+		}
+
+		w.onFailure(lint.Failure{
+			Confidence: 0.8,
```
chavacava, line 68: confidence could be 1, here we are sure it's an increment/decrement that should be replaced

testdata/no_increment_decrement.go
```diff
@@ -0,0 +1,28 @@
...
+	x-- // MATCH /should replace x-- with x -= 1/
+}
+
+func loopCounters() {
+	for i := 0; i < 10; i++ {
+		_ = i
+	}
+	for j := 10; j > 0; j-- {
+		_ = j
+	}
```
chavacava, line 15: It could be interesting to test with an increment in a for body


```suggestion
	for j := 10; j > 0; j-- {
		_ = j
	}
	
	x := 0
	for j := 10; j > 0; j-- {
		_ = j
		x++ // MATCH /should replace x++ with x += 1/
		x-- // MATCH /should replace x-- with x -= 1/
	}
```

chavacava (commented):
Hi @ChrisJr404, thanks for the PR.
I've left some comments.

chavacava:
@ChrisJr404 please also include the rule in the rule list table on the README.md

ChrisJr404:
Good catch - added the rule to the README rule list table (between `nested-structs` and `optimize-operands-order`).

ccoVeille:
I have a problem here.

First, this rule is purely stylistic. Some may like it, some may not. I feel like 99% of people will be in the second team.

But the real problem is the fact once merged this rule will be part of revive enable all rules settings. And then it would be enabled for anyone.

For me, this should be a linter independent from revive.

The only possible way according to me would be that this feature becomes a setting of an existing revive rule. So here, I feel a setting of increment-decrement should be the way.

Also, if we now consider there is a rule for using `i++` and one for using `i += 1`, using `enable-all-rules` would lead to infinite back and forth fixes

This is the feedback I do, and I'm looking for maintainers feedback.

Do not rush into doing changes.

PR #1779, opened 2026-08-22
feat(identical-switch-branches): add allow-identical-default option

### LINKED ISSUE #1546: identical-switch-branches: exclude default

The `identical-switch-branches` rule is useful but a bit too strict.

When writing switch statements on enums from external libraries, it can be useful to handle known current values but also a default value that matches a branch. This makes it clear what is explicitly mapped and what is a fallback value.

I'd like an option to ignore default branches.

I have considered disabling the rule entirely or commenting out specific switch statements. Unfortunately, either option is undesirable as the check is still useful for other identical branch scenarios.

**Additional context**

vaibhav8a:
Closes #1546.

## What

Adds an `allow-identical-default` option to `identical-switch-branches`.

Switching over values from an external package, a fallback that repeats a listed case is often deliberate — it documents which values are explicitly handled and that everything else lands on the same code. Today the rule reports that, and the only ways out are disabling the rule or annotating each switch, both of which also give up the identical-*case* check, which is the part that finds bugs.

## Behaviour

With the option set, a `default` clause:

- is not reported against an earlier branch, **and**
- does not record its own hash, so a later clause cannot be reported against it either

Its body is still walked, so a nested `switch` inside an ignored `default` is analyzed as before. Identical `case` clauses are still reported, and `fallthrough` handling is untouched. Default behaviour is unchanged — the existing `identical_switch_branches` fixture passes as-is.

Follows the `enforce-switch-style` pattern: a `Configure` method, `isRuleOption` for the name, and an error naming the valid option for anything else.

## Testing

New fixture `testdata/identical_switch_branches_allow_identical_default.go` covering: a default identical to a case (not reported), two identical *cases* alongside a default (still reported), an empty default matching an empty case, a nested switch inside an ignored default (still analyzed), and a fallthrough case.

I also checked the option is actually load-bearing rather than the fixture just being permissive — running the same fixture with the rule at its defaults produces **four** additional failures that the option suppresses:

```
got:  "switch" with identical branches (lines 7 and 11)
got:  "switch" with identical branches (lines 19 and 23)
got:  "switch" with identical branches (lines 29 and 31)
got:  "switch" with identical branches (lines 53 and 55)
```

```
go test ./...                      # all packages ok
go vet ./...                       # clean
gofmt -l .                         # clean
go run . --config revive.toml ./...  # clean (revive on itself)
```

`RULES_DESCRIPTIONS.md` documents the option with a config example.

## Naming

I went with `allow-identical-default` to sit beside `enforce-switch-style`'s `allow-no-default` / `allow-default-not-last`. Happy to rename if you'd prefer something like `ignore-default`.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -69,6 +69,8 @@ The canonical example is [`rule/argument_limit.go`](rule/argument_limit.go). For
 
     If the rule takes arguments, also implement `lint.ConfigurableRule.Configure(lint.Arguments) error`.
     Validate arguments there and return errors rather than panicking.
+    Any newly added option must be passed as a `map[string]any` (a single map argument with named keys),
+    not as a positional/scalar argument — this keeps rule configuration extensible and self-documenting.
 2. **Naming** — `Name()` returns `kebab-case` (e.g. `argument-limit`).
    The Go type is `ArgumentsLimitRule`. Source file is `argument_limit.go`. Keep these three in lockstep.
 3. **Concurrency** — `Apply` may be called concurrently for different files.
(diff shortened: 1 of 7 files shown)


ccoVeille:
The code seems OK, but I'm insure what is the need

So I posted a comment on the issue

- https://github.com/mgechev/revive/issues/1546#issuecomment-5380668883

Please reply about it on the issue itself, and not here on the PR

vaibhav8a:
Answered the "do you have real-life examples" question on the issue: https://github.com/mgechev/revive/issues/1546#issuecomment-5461060810

Short version — I scanned the Go sources of several projects for a `default` clause whose body is identical to a listed case, and found the pattern in `perses`, `devguard`, `aibrix` and `kubescape`. It shows up in two shapes: an unknown input falling back to a listed value (`--logLevel` defaulting to `info`), and an exhaustive-ish mapping with a safe default (`SameSite.String()` returning Lax for anything unrecognised). In both, the duplication is deliberate and the alternatives are merging the default into the case list or a nolint.

alexandear:
@vaibhav8a thank you for your contribution. Merging.

PR #1781, opened 2026-08-25
feat: add `exit-after-defer` rule

### LINKED ISSUE #1300: defer: consider reporting exit like in deep-exit
We have `deep-exit` rule that reports issues when using

We also have `defer` rule that reports bad usage of defer.

Both of them are useful, some people are using only one, other use

I feel like the code used in deep-exit rule could be reused in defer rule to report the defer that are broken.

The issue I see now with revive is that a code like this is not reported.

```go
 func main() {
     defer func() {
         fmt.Println("whatever")
     }()
 
     err := errors.New("test")
     if err != nil {
         log.Fatal("anything", err)
     }
 }
```

I do use defer and deep-exit, deep-exit reports the issues with the usage of exiter out of `main`

It would be something comparable to go-critic rule

https://go-critic.com/overview.html#exitafterdefer

ChrisJr404:
Adds a new `exit-after-defer` rule that flags calls to `os.Exit`, `syscall.Exit` or `log.Fatal` that are reachable after a `defer` statement, since the deferred calls will not run in that case.

`panic` and `log.Panic` are not reported because deferred functions still run while a panic unwinds the stack.

The rule is opt-in like the others, so existing configurations are unaffected. Tests and docs are included.

Closes #1300

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,6 +546,7 @@ List of all [available rules](./RULES_DESCRIPTIONS.md).
 | [`error-return`](./RULES_DESCRIPTIONS.md#error-return)        |  n/a   | The error return parameter should be last.                       | 1.0 |   yes    |  no   |
 | [`error-strings`](./RULES_DESCRIPTIONS.md#error-strings)       |  []string   | Conventions around error strings.                                | 1.0 |   yes    |  no   |
 | [`errorf`](./RULES_DESCRIPTIONS.md#errorf)              |  n/a   | Should replace `errors.New(fmt.Sprintf())` with `fmt.Errorf()`   | 1.0 |   yes    |  yes  |
+| [`exit-after-defer`](./RULES_DESCRIPTIONS.md#exit-after-defer) |  n/a   | Spots `os.Exit`, `syscall.Exit` or `log.Fatal` calls reachable after a `defer` | 1.0 |    no    |  no   |
 | [`exported`](./RULES_DESCRIPTIONS.md#exported)            |  []string   | Naming and commenting conventions on exported symbols.           | 1.0 |   yes    |  no   |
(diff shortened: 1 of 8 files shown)


alexandear:
@ccoVeille, could you take a look?

ccoVeille:
@alexandear yes, but later this week.

I feel like the issue might have not be clear enough. I will have to check if there are no other linters that report usage of defer with "exiters".

Right now, I'm unsure whether the rule should be:
- separate from the deep-exit or not.
- be limited to `func main` and `func TestMain`

This PR might be OK, but I have to check again, as the issue I created is pretty old

alexandear:
I also don't understand why we need a separate rule.

ChrisJr404:
The rule targets a specific bug: a call that ends the program (`os.Exit`, `log.Fatal`, `log.Panic`) placed after a `defer` in the same function, where the deferred call silently never runs. That is a real correctness trap (unflushed writers, unreleased locks, un-run cleanup) that the existing `defer` and `deep-exit` rules do not catch: `deep-exit` flags the exit call itself regardless of defers, and `defer` covers loop/return/recover misuse, not ordering against a program-ending call.

If you would rather fold this into `deep-exit` as an option than ship a separate rule, I am fine with that; I split it out only because the two checks flag different code and users may want one without the other. Let me know which direction you prefer.

alexandear:
Thanks @ChrisJr404.

My hesitation is only about where the check lives: I'd rather it be a subcase of the `defer` rule than a new rule.

Two things I'd want changed while you're at it. First, drop `isUnrecoverableExitCall` and use `astutils.IsCallToExitFunction`.

Second, confidence `0.8` rather than `1`. The check is source-order only, so this false positives:

```go
func branch(a bool) {
      if a {
              defer log.Println("cleanup")
              return
      }
      os.Exit(1) // reported, but no defer is registered on this path
}
```

PR #1790, opened 2026-09-09
enforce-slice-style: suggest nil expressions

### LINKED ISSUE #1543: enforce-slice-style: false positive in an expression context
**Describe the bug**
`enforce-slice-style` dictates using `var x []int` rather than `x := []int{}` for slice definitions when configured with the `"nil"` argument, but it also reports the usage of `[]int{}` in an expression context where `var x []int` can't be used

**To Reproduce**

```shell
revive -config=revive.toml ./...
```

```toml
[rule.enforce-slice-style]
Arguments = ["nil"]
```

With the following code:
```go
package main

func main() {
    foo([]int{})
}

func foo(x []int) {}
```

And got this output:
```
main.go:4:6: use nil slice declaration (e.g. var args []type) instead of []type{}
```

**Expected behavior**
I didn't expect this to be a lint finding because this particular rule seems to be all about declarations, not expressions. The error message makes me think the linter is suggesting the following because it mentions `var args []type`

```go
func main() {
    var x []int
    foo(x)
}
```

So I wonder if either:
1. Only statements should be checked
2. Or, maybe the error message could say something about `use nil instead of []type` when findings are in an expression context; using `foo(nil)` here also suppresses the lint.

fzlzjerry:
Closes #1543.

The nil-style diagnostic always suggested a `var` declaration (`var args []type`), even when the empty slice appears as an expression (function argument, return value, struct/map value, ...) where a declaration can't be substituted.

Make the suggestion context-aware, without changing which expressions the rule reports:
- variable declarations keep the `var args []type` suggestion;
- every other context suggests a nil slice expression (`[]type(nil)`).

Update the diagnostic expectations per context, add return-value cases for literals, `make`, and typed nil, and document both suggestions (with examples) in the nil-style rule description.

### DIFF
--- a/RULES_DESCRIPTIONS.md
+++ b/RULES_DESCRIPTIONS.md
@@ -758,7 +758,7 @@ arguments = [{ func-arg-style = "full", func-ret-val-style = "short" }]
 
 _Go version_: 1.0.
 
-_Description_: This rule enforces consistent usage of `make([]type, 0)`, `[]type{}`, or `var []type` for slice initialization.
+_Description_: This rule enforces consistent usage of `make([]type, 0)`, `[]type{}`, or nil slices for slice initialization.
 It does not affect `make([]type, non_zero_len, or_non_zero_cap)` constructions as well as `[]type{v1}`.
 Nil slices are always permitted.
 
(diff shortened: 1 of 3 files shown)


alexandear (approved):

alexandear:
@fzlzjerry thank you! Merging.

PR #1802, opened 2026-09-22
deep-exit: improve report message

### LINKED ISSUE #1612: deep-exit: improve error message for flag.Parse
I would like to suggest the deep-exit to provide a better error message for `flag.Parse` detection

https://github.com/mgechev/revive/blob/0527051da32f0327a5e13b55a9572f9301ef4e01/rule/deep_exit.go#L68-L81

Right now, we report `deep-exit: calls to flag.Parse only in main() or init() functions`

We could complete this error message with a suggestion that would explain the user they can refactor the code to use [flag.NewFlagSet](https://pkg.go.dev/flag#NewFlagSet) with [flag.ContinueOnError](https://pkg.go.dev/flag#ContinueOnError)

alexandear:
Closes #1612


ccoVeille:
Thank you for working in this

